PR #1, opened 2019-02-01
add gcloud-wrapped container, and be more verbose

listx:
Add a GCB pipeline to get the promoter tested, and built as a Docker image, and finally published.


Makefile
```diff
@@ -1,8 +1,13 @@
+# This is not really the source of truth for the build system; refer to
+# cloudbuild.yaml for that.
+
+BAZEL_BUILD_OPTS:=--workspace_status_command=${PWD}/workspace_status.sh
```
BenTheElder: non-blocking suggestion:
I think we usually put this in a bazelrc, which allows invoking bazel w/o any wrappers.
eg:
https://github.com/kubernetes/test-infra/blob/9a497cc7808fbb19cd85eb8caa1453732272745a/.bazelrc#L2

lib/stream/types.go
```diff
@@ -23,7 +23,7 @@ import (
 // Producer is an interface for anything that can generate an io.Reader from
 // which we can read from (typically JSON output).
 type Producer interface {
-	Produce() (io.Reader, error)
+	Produce() (io.Reader, io.Reader, error)
```
BenTheElder: might be worth noting what these are supposed to be / updating the docs?

workspace_status.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+p_() {
```
BenTheElder: this is cute, I would put a short comment about p_

lib/stream/subprocess.go
```diff
@@ -30,20 +30,24 @@ type Subprocess struct {
 
 // Produce runs the external process and returns a handle to stdout (an
```
BenTheElder: this returns two readers now

lib/stream/fake.go
```diff
@@ -29,9 +29,9 @@ type Fake struct {
 
 // Produce a fake stream. Unlike a real stream, this does not call a subprocess
 // --- instead it just reads from a predefined stream (Bytes).
```
BenTheElder: s/reads from/provides/ -- the thing on the other end of the io.Reader is reading.

cloudbuild.yaml
```diff
@@ -0,0 +1,98 @@
...
+  - setup
+  entrypoint: "bash"
+  dir: ${_GOPATH}/src/github.com/GoogleCloudPlatform/${_REPO}
+  args:
+  - "-c"
+  - |
+    set -o errexit
+    set -o nounset
+    set -o pipefail
+    set -o xtrace
```
BenTheElder: why not just call bazel directly?

cloudbuild.yaml
```diff
@@ -0,0 +1,98 @@
...
+  _REPO: k8s-container-image-promoter
+  _GOLANGCI_LINT_VERSION: v1.13.2
+  _GOPATH: /workspace/go
+
+steps:
+- name: "gcr.io/cloud-builders/git"
+  id: setup
+  entrypoint: "bash"
+  args:
+  - "-c"
```
BenTheElder: is it usual to write these scripts entirely inline as opposed to writing them as files in the repo and calling them?

Gopkg.toml
```diff
@@ -0,0 +1,34 @@
...
+#   unused-packages = true
+
+
+[[constraint]]
+  name = "gopkg.in/yaml.v2"
+  version = "2.2.2"
+
+[prune]
+  go-tests = true
+  unused-packages = true
```
BenTheElder: you may also want:
```toml
  non-go = true
```

Gopkg.toml
```diff
@@ -0,0 +1,34 @@
+# Gopkg.toml example
```
BenTheElder: remove the example contents?

Makefile
listx: Opened a new issue for this.

listx: Errr the issue is #4

cloudbuild.yaml
listx: Mainly for uniformity with the other steps and also because I don't like breaking up all the arguments to bazel into a YAML list. I guess it looks a bit obtuse.

BTW I think cloudbuild.yaml might change in the future though to be yet another Prow job on some Prow cluster.

cloudbuild.yaml
listx: Depends. In this case this logic is only meant to be used by the CI (no one else), so keeping it all inside this single file isn't so bad.  /shrug

Gopkg.toml
listx: Hmm, I didn't know about this (as this file was just auto-generated by dep --- dep being used only to pull in the  so that we can get golangci-lint to work out of the box). But at the same time I actually do not track what is in /vendor, so maybe it would be a NOP even with `non-go = true`? And like, if it ain't broke don't fix it... right?

Makefile
ixdy: I agree with Ben's suggestion. Why not just make that change in this PR?

Gopkg.toml
BenTheElder: ok, FWIW the demo file includes this.

workspace_status.sh
```diff
@@ -0,0 +1,30 @@
...
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# p_() takes 2 inputs, the name of a variable and a value. It then prints the
+# variable and the value to be fed into bazel's --workspace_status flag. The
+# interesting thing is that it will *only* print the given value if there is
+# *not* an environment variable with the same name that is already set (with
+# some nonempty string). This way, bazel will either read in these ENV vars, but
```
ixdy: I don't follow why you're doing this; in general, bazel rules aren't dependent on the environment, so anything you want as an input to the rule should be set in the workspace status.

workspace_status.sh
```diff
@@ -0,0 +1,30 @@
...
+    else
+        return 1
+    fi
+}
+
+git_branch="$(git rev-parse --abbrev-ref HEAD)"
+git_desc="$(git describe --always)"
+img_name="cip"
+
+p_ IMG_REGISTRY gcr.io
```
ixdy: you probably need to prefix most/all of these with `STABLE_`, so that they trigger rebuilds of stamped targets.

otherwise, changes to the values of these won't trigger rebuilds, which can be confusing.

cloudbuild.yaml
BenTheElder: the usual pattern in prow at least is to _not_ encode the logic in the job, fwiw

lib/dockerregistry/inventory_test.go
```diff
@@ -104,7 +104,7 @@ func TestReadJSONStream(t *testing.T) {
 	for _, test := range tests {
 		var sr stream.Fake
 		sr.Bytes = []byte(test.input)
-		h, _ := sr.Produce()
+		h, _, _ := sr.Produce()
```
BenTheElder: hmm why aren't we handling these?

lib/stream/fake.go
```diff
@@ -28,10 +29,11 @@ type Fake struct {
 }
 
 // Produce a fake stream. Unlike a real stream, this does not call a subprocess
-// --- instead it just reads from a predefined stream (Bytes).
-func (producer *Fake) Produce() (io.Reader, error) {
+// --- instead it just provides a predefined stream (Bytes).
```
BenTheElder: tin nit again: on stdout only

workspace_status.sh
listx: The missing context here that I forgot to add is that this way, I can pass in substitution vars into the GCB (cloudbuild.yaml), which can then be passed in as env vars to the image that is running in a step (to be consumed by this p_() function into bazel). The idea is that I can now reuse the same cloudbuild.yaml for different targets.

Makefile
listx: @ixdy Because he said it was non-blocking. Also, I didn't want to hold back #2 because this PR has stuff that touches a lot of bazel stuff.

cloudbuild.yaml
listx: Noted.
Mind if I make this into another issue to create some "ci" folder with all this logic in it (instead of embedding inline like this into the CI logic directly)?

Gopkg.toml
listx: Oh odd --- I guess dep changed behavior recently. Will add.

workspace_status.sh
listx: Gotcha, will do.

lib/dockerregistry/inventory_test.go
listx: It's because ReadJSONStream only reads/parses JSON from stdout, not stderr. Also, the stream.Fake only produces output in stdout, so these are always empty/nil.

listx: Oops I meant json.Consume()

lib/stream/fake.go
listx: Thank you so much for catching stuff like this. Fixed in newest version.

lib/dockerregistry/inventory_test.go
BenTheElder: if the fake ever does produce errors we won't be caching them though.

cloudbuild.yaml
listx: Added a new issue #5 for this.

lib/dockerregistry/inventory_test.go
listx: True. Will add.

listx:
Hmm, the fact that you can't /lgtm bothers me. I really want to convert the cloudbuild.yaml to a Prow job...

BUILD.bazel
```diff
@@ -54,3 +63,22 @@ go_binary(
     ],
     visibility = ["//visibility:public"],
 )
+
+container_push(
+    name = "push-cip-latest",
+    image = ":cip-with-gcloud",
+    format = "Docker",
+    registry = "{STABLE_IMG_REGISTRY}",
+    # repository should end with the name of the image to use.
```
ixdy: is `{STABLE_IMG_NAME}` missing here?

lib/dockerregistry/inventory.go
```diff
@@ -833,14 +856,18 @@ func (op *TagOp) PrettyValue() string {
 // PrettyValue is a prettified string representation of a PromotionRequest.
 func (pr *PromotionRequest) PrettyValue() string {
 	var b strings.Builder
-	fmt.Fprintf(&b, "%v -> %v: Tag: '%v' <%v> %v@%v (from '%v')\n",
+	fmt.Fprintf(&b, "%v -> %v: Tag: '%v' <%v> %v@%v", //
```
ixdy: trailing `//` ?

workspace_status.sh
ixdy: I still don't really understand. If I'm understanding the comment and this code properly, if you set e.g. `STABLE_IMG_REGISTRY` in the environment, then it won't be available to bazel rules anymore. Why do you want to do that?

BUILD.bazel
listx: Doh, that looks odd. It's not missing because the image name is already embedded into STABLE_IMG_REPOSITORY. But it's better to un-embed it to make it all more explicit.

ixdy: ah, that's inconsistent with the `container_bundle` rule above?

lib/dockerregistry/inventory.go
listx: Oops, yep. Thank you.

workspace_status.sh
listx: Sorry, allow me to reword your example. If you set `STABLE_IMG_REGISTRY` in the environment, then bazel rules *will* use them (as per `p_()` in `workspace_status.sh`). Like you said, bazel rules aren't dependent on the environment; and so I wrote `p_()` to pass along env vars into bazel. The complete picture (accounting for how GCB will use these env vars) is like this:

1. User sets GCB substitution variables, and passes them into the environment for a `step` definition in the GCB YAML. E.g., `gcloud container builds submit --substitutions=STABLE_IMG_NAME=foo`
2. The STABLE_IMG_NAME env var is exposed into the bazel build step inside GCB.
3. The bazel build step's invocation of `workspace_status.sh` then uses `foo` for STABLE_IMG_NAME instead of `cip` in `workspace_status.sh`, because it is set in the environment.
4. Now bazel uses STABLE_IMG_NAME=foo in its build.

BUILD.bazel
listx: Ah yes, that too. I've changed this to un-embed it.

workspace_status.sh
ixdy: ohhh, now I understand. I was a bit confused by "only prints the value if the environment variable isn't set" - I thought that meant it wouldn't print anything at all.

It might be clearer to write something like

`p_` takes two arguments to define a bazel workspace status variable:
- the name of the variable
- a default value

If an environment variable with the corresponding name is set, its value is used. Otherwise, the provided default value is used.

listx: Done. Looking at this again, your comment for `p_()` is a lot easier to understand, so I used it as-is. I hope you don't mind. =)

workspace_status.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+#`p_` takes two arguments to define a bazel workspace status variable:
```
ixdy: backticks in shell scripts always make me nervous. :)

listx: Hmm, I should remove it... in a later patch, heh.

PR #3, opened 2019-02-15
Changed BUILD.bazel to ignore gc_linkopts on MacOS

bartsmykla:
Closes: https://github.com/GoogleCloudPlatform/k8s-container-image-promoter/issues/2

Because MacOS X doesn't support static binaries:
https://github.com/bazelbuild/rules_go/issues/161#issuecomment-304469169
I made `gc_linkopts` inside cip binary configuration
present on systems which are not MacOS only.

Signed-off-by: Bart Smykla <bsmykla@vmware.com>

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -52,14 +52,6 @@ go_library(
 go_binary(
     name = "cip",
     embed = [":go_default_library"],
-    # These gc_linkopts make it a static binary. See
-    # https://github.com/bazelbuild/rules_go/issues/161#issuecomment-304469169.
-    gc_linkopts = [
-        "-linkmode",
-        "external",
-        "-extldflags",
-        "-static",
-    ],
     visibility = ["//visibility:public"],
 )
 


BUILD.bazel
```diff
@@ -46,11 +46,16 @@ go_binary(
     embed = [":go_default_library"],
     # These gc_linkopts make it a static binary. See
     # https://github.com/bazelbuild/rules_go/issues/161#issuecomment-304469169.
-    gc_linkopts = [
-        "-linkmode",
-        "external",
-        "-extldflags",
-        "-static",
-    ],
+    gc_linkopts = select({
```
ixdy: I think we can just remove `gc_linkopts` entirely. `cip` appears to be a pure go binary, so it'll be statically linked by default.

ixdy: we can also force it to be a pure go binary by setting `pure = "true"`, but I don't think we need or want that here.

listx:
Oops, just noticed that you already opened this PR @bartsmykla. Can we wait for #1 first?

listx:
Meanwhile, I have no idea how to add you as reviewers to #1... Can you still have a look there as a sanity check @bartsmykla / @ixdy ?

bartsmykla:
I signed it!

bartsmykla:
@listx sure we can! :-)

listx:
@bartsmykla Not sure whether you saw ixdy's review comments. Are you going to address them?

bartsmykla:
Sorry, missed that! I'm on it. :-)

BUILD.bazel
bartsmykla: Ack

bartsmykla:
@listx done :-)

listx:
LGTM. I tested it locally by cherry-picking and building locally.

Issue #11, opened 2019-02-21
validate digests and tags


listx:
If today we have a typo in the sha256 digest or Docker tag field, giving an invalid string is not caught during parsing. We just expect them to be stringlike things.

During manifest parsing, add a check to make sure sha256 digests are actually hex characters only, and that Docker tags meet the requirements of Docker tags described in https://docs.docker.com/engine/reference/commandline/tag/.

listx:
This was fixed with #12.

PR #12, opened 2019-02-21
validate Docker digest and tag formatting

listx:
The unit tests are a little mundane, but they are better than nothing.

### DIFF
--- a/lib/dockerregistry/inventory.go
+++ b/lib/dockerregistry/inventory.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"regexp"
 	"sort"
 	"strings"
 	"sync"
@@ -118,8 +119,50 @@ func ParseManifestFromFile(filePath string) Manifest {
 // from ParseManifestFromFile() so that it can be tested independently.
 func ParseManifest(bytes []byte) (Manifest, error) {
 	var m Manifest
-	err := yaml.UnmarshalStrict(bytes, &m)
-	return m, err
+	if err := yaml.UnmarshalStrict(bytes, &m); err != nil {
+		return m, err
+	}
+
+	return m, m.Validate()
+}
+
+// Validate checks for semantic errors in the yaml fields (the structure of the
+// yaml is checked during unmarshaling).
+func (m Manifest) Validate() error {
+	return validateImages(m.Images)
+}
+
+func validateImages(images []Image) error {
+	for _, image := range images {
+		for digest, tagSlice := range image.Dmap {
+			if err := validateDigest(digest); err != nil {
+				return err
+			}
+
+			for _, tag := range tagSlice {
+				if err := validateTag(tag); err != nil {
+					return err
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func validateDigest(digest Digest) error {
+	validDigest := regexp.MustCompile(`^sha256:[0-9a-f]{64}$`)
+	if !validDigest.Match([]byte(digest)) {
+		return fmt.Errorf("invalid digest: %v", digest)
+	}
+	return nil
+}
+
+func validateTag(tag Tag) error {
+	var validTag = regexp.MustCompile(`^[\w][\w.-]{0,127}$`)
+	if !validTag.Match([]byte(tag)) {
+		return fmt.Errorf("invalid tag: %v", tag)
+	}
+	return nil
 }
 
 // PrettyValue creates a prettified string representation of MasterInventory.
--- a/lib/dockerregistry/inventory_test.go
+++ b/lib/dockerregistry/inventory_test.go
@@ -206,6 +206,100 @@ images:
 	}
 }
 
+func TestParseImageDigest(t *testing.T) {
+	// nolint[lll]
+	var shouldBeValid = []string{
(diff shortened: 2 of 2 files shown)


listx:
This fixes #11.

listx:
@bartsmykla can you take a look at this? Not sure how to add you as a reviewer because you are not in the GoogleCloudPlatform org... doh.

lib/dockerregistry/inventory.go
```diff
@@ -119,9 +120,62 @@ func ParseManifestFromFile(filePath string) Manifest {
...
 
+// Validate checks for semantic errors in the yaml fields (the structure of the
+// yaml is checked during unmarshaling).
+func (m Manifest) Validate() error {
+	err := validateImages(m.Images)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
bartsmykla: Maybe:
```go
func (m Manifest) Validate() error {
	return validateImages(m.Images)
}
```
would be sufficient?

lib/dockerregistry/inventory.go
```diff
@@ -119,9 +120,62 @@ func ParseManifestFromFile(filePath string) Manifest {
...
+
+func validateImages(images []Image) error {
+	var err error
+	for _, image := range images {
+		for digest, tagSlice := range image.Dmap {
+			err = validateDigest(digest)
+			if err != nil {
+				return err
+			}
```
bartsmykla: Maybe you could write:
```go
if err := validateDigest(digest); err != nil {
    return err
}
```

lib/dockerregistry/inventory.go
```diff
@@ -119,9 +120,62 @@ func ParseManifestFromFile(filePath string) Manifest {
...
+
+			for _, tag := range tagSlice {
+				err = validateTag(tag)
+				if err != nil {
+					return err
+				}
```
bartsmykla: The same thing here:
```go
if err = validateTag(tag); err != nil {
    return err
}
```

lib/dockerregistry/inventory.go
```diff
@@ -119,9 +120,62 @@ func ParseManifestFromFile(filePath string) Manifest {
...
+func (m Manifest) Validate() error {
+	err := validateImages(m.Images)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func validateImages(images []Image) error {
+	var err error
```
bartsmykla: If you'll refactor the error handling I commented below, you could get rid off this variable declaration.

lib/dockerregistry/inventory.go
```diff
@@ -119,9 +120,62 @@ func ParseManifestFromFile(filePath string) Manifest {
...
+					return err
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func validateDigest(digest Digest) error {
+	var validDigest = regexp.MustCompile(`^sha256:[0-9a-f]{64}$`)
```
bartsmykla: Why are you using here `var [...] =` and in other places `[...] :=`?

lib/dockerregistry/inventory.go
```diff
@@ -119,9 +120,62 @@ func ParseManifestFromFile(filePath string) Manifest {
...
+	ok := validDigest.Match([]byte(digest))
+	if !ok {
+		return fmt.Errorf("invalid digest: %v", digest)
+	}
+	return nil
```
bartsmykla: I think I would refactor it to:
```go
if ok := validDigest.Match([]byte(digest)); !ok {
    return fmt.Errorf("invalid digest: %v", digest)
}
```

lib/dockerregistry/inventory.go
```diff
@@ -119,9 +120,62 @@ func ParseManifestFromFile(filePath string) Manifest {
...
+}
+
+func validateTag(tag Tag) error {
+	var validTag = regexp.MustCompile(`^[\w][\w.-]{0,127}$`)
+	ok := validTag.Match([]byte(tag))
+	if !ok {
+		return fmt.Errorf("invalid tag: %v", tag)
+	}
```
bartsmykla: I think I would refactor it to:
```go
if ok := validTag.Match([]byte(tag)); !ok {
    return fmt.Errorf("invalid tag: %v", tag)
}
```

lib/dockerregistry/inventory.go
```diff
@@ -119,9 +120,62 @@ func ParseManifestFromFile(filePath string) Manifest {
 func ParseManifest(bytes []byte) (Manifest, error) {
 	var m Manifest
 	err := yaml.UnmarshalStrict(bytes, &m)
+	if err != nil {
+		return m, err
+	}
+
+	err = m.Validate()
```
bartsmykla: Maybe you could write:
```go
if err := yaml.UnmarshalStrict(bytes, &m); err != nil {
    return m, err
}

return m, m.Validate()
```

lib/dockerregistry/inventory.go
listx: Ack

listx:
@bartsmykla Thx for the review!

The `if err := ...; err != nil` idiom is something I haven't gotten used to yet.

PTAL

bartsmykla:
LGTM now :-)

PR #22, opened 2019-02-27
bazel: always build images with gcloud as a base image

listx:
Because we depend on `gcloud`, it doesn't really make sense to build a
docker image with solely the cip Go binary in it in isolation.

This change affects 2 rules:
  - "cip-docker-loadable": for debugging
  - "push-cip-tagged": for pushing up official images that are tagged as
     something other than just "latest"

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -28,10 +28,9 @@ container_image(
 container_bundle(
     name = "cip-docker-loadable",
     images = {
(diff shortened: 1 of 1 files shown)


dims:
/approve
/lgtm

(don't think these will do anything here ... will they? :)

listx:
@dims unfortunately no, nothing happens because it's not hooked up to Prow. That will change in the future, but only after this repo gets transferred to the github.com/kubernetes org. For now we will have to do things the old fashioned way.

Issue #21, opened 2019-02-27
add a script to invoke the promoter multiple times against different manifests


listx:
Each time it is invoked, it should set activate the right service account.

PR #23, opened 2019-02-27
add script to invoke the promoter against multiple manifests

listx:
This also rearranges the layout of the binary and scripts in the docker
image to be more intuitive to use (everything under "/cip" directory).

### DIFF
--- a/util/multirun.sh
+++ b/util/multirun.sh
@@ -1,0 +1,67 @@
+#!/bin/bash
+
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


listx:
This fixes #21

util/multirun.sh
```diff
@@ -0,0 +1,59 @@
...
+    # Authenticate as the service account. This allows the promoter to later
+    # call gcloud with the flag `--account=...`. We can allow the service
+    # account creds file to be empty, for testing cip locally (for the case
+    # where the service account creds are already activated).
+    if [ -f "${service_account_creds}" ]; then
+        gcloud auth activate-service-account --key-file="${service_account_creds}"
+    fi
+
+    # Run the promoter against the manifest.
+    "${cip}" -verbosity=3 -manifest="${manifest}" ${CIP_OPTS:+$CIP_OPTS}
```
dims: don't we need to revoke the service account after the promoter is run?

listx: The promoter explicitly passes "--account=XXX" for each interaction with GCR (using the service account defined in the manifest YAML), so there's no concern about it using the wrong service account (as it appears from this shell code's invocation of the promoter).

But even so, there's no reason why a service account should be active if it won't be used. I'll update this...

listx: I just updated it to add in the revocations after each run of the promoter. PTAL?

PR #24, opened 2019-03-05
update copyright from Google to Kubernetes

listx:


spiffxp:
This came about as a result of feedback from mailing list discussion (ref: https://github.com/kubernetes/org/issues/551#issuecomment-468880069)

Checked out locally to confirm you got all occurrences

LGTM

PR #29, opened 2019-03-09
add '-no-service-account' flag

listx:
This allows requests to gcloud to be executed without the
    "--account=..." flag. The use case for this flag is for dry runs against
    publicly readable GCRs (to guarantee that we will not be making any
    modifications; it also allows us to avoid mounting any service account
    secrets, which has been the norm so far for all cases, even for dry
    runs). See also kubernetes/test-infra#11414.

    We keep the existing functionality of '-dry-run' independent of this
    flag; this is for the use case where we want to run '-dry-run' against
    private registries that are _not_ publicly readable.

    To summarize:
            - use '-dry-run -no-service-account' for doing dry runs against publicly readable registries
            - use '-dry-run' for doing dry runs against private registries

### DIFF
--- a/cip.go
+++ b/cip.go
@@ -52,6 +52,9 @@ func main() {
 		false,
 		"print what would have happened by running this tool;"+
 			" do not actually modify any registry (default: false)")
+	noSvcAcc := false
+	flag.BoolVar(&noSvcAcc, "no-service-account", false,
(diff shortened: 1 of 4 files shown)


listx:
Hmmm maybe I should use "-no-service-account" instead. I think it feels more natural that way...

PR #32, opened 2019-03-18
make multirun.sh aware of changed manifest files (only run on changed files)

listx:
This way, we don't have to blindly run against multiple manifests.

### DIFF
--- a/cip.go
+++ b/cip.go
@@ -19,6 +19,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"os"
 
 	// nolint[lll]
 	reg "github.com/kubernetes-sigs/k8s-container-image-promoter/lib/dockerregistry"
@@ -47,6 +48,11 @@ func main() {
 		false,
 		"delete tags in the destination registry that are not declared"+
 			" in the Manifest (default: false)")
+	parseOnlyPtr := flag.Bool(
+		"parse-only",
+		false,
+		"only check that the given manifest file is parseable as a Manifest"+
+			" (default: false)")
 	dryRunPtr := flag.Bool(
 		"dry-run",
 		false,
@@ -56,6 +62,11 @@ func main() {
 	flag.BoolVar(&noSvcAcc, "no-service-account", false,
 		"do not pass '--account=...' to all gcloud calls (default: false)")
 	flag.Parse()
+
+	if *parseOnlyPtr {
+		reg.ParseManifestFromFile(*manifestPtr)
+		os.Exit(0)
+	}
 
 	if *dryRunPtr {
 		fmt.Println("---------- DRY RUN ----------")
--- a/cip.go
+++ b/cip.go
@@ -19,6 +19,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"log"
 	"os"
 
 	// nolint[lll]
@@ -63,16 +64,17 @@ func main() {
 		"do not pass '--account=...' to all gcloud calls (default: false)")
 	flag.Parse()
 
+	mfest, err := reg.ParseManifestFromFile(*manifestPtr)
+	if err != nil {
+		log.Fatal(err)
+	}
 	if *parseOnlyPtr {
-		reg.ParseManifestFromFile(*manifestPtr)
 		os.Exit(0)
 	}
 
 	if *dryRunPtr {
 		fmt.Println("---------- DRY RUN ----------")
 	}
-
-	mfest := reg.ParseManifestFromFile(*manifestPtr)
 
 	mi := map[reg.RegistryName]reg.RegInvImage{}
 	for _, registry := range mfest.Registries {
--- a/util/multirun.sh
+++ b/util/multirun.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Copyright 2019 The Kubernetes Authors.
 #
--- a/util/multirun.sh
+++ b/util/multirun.sh
@@ -25,6 +25,7 @@ usage()
 {
     echo >&2 "usage: $0 <path/to/cip/binary> [<path/to/manifest.yaml>,<path/to/service-account.json>, ...]"
     echo >&2 "The 2nd argument onwards are '<manifest>,<service-account>,<service-account>...' strings."
+    echo >&2 "If an execution does not require any service account, pass in '<manifest>,' (notice the trailing comma)."
     echo >&2
 }
 
@@ -32,6 +33,14 @@ if (( $# < 2 )); then
     usage
     exit 1
 fi
+
+for opts in "$@"; do
+    if ! [[ "$opts" =~ .*,.* ]]; then
+        echo >&2 "invalid argument: $opts"
+        usage
+        exit 1
+    fi
+done
 
 cip="$1"
 shift
--- a/util/multirun.sh
+++ b/util/multirun.sh
@@ -41,7 +41,9 @@ for opts in "$@"; do
     service_accounts=()
     service_account_keyfiles=()
     while IFS= read -r keyfile; do
-        service_account_keyfiles+=( "$keyfile" )
+        if [[ -n "$keyfile" ]]; then
+            service_account_keyfiles+=( "$keyfile" )
+        fi
     done < <( echo "$opts" | cut -d, -f2- | tr , '\n' )
     for keyfile in "${service_account_keyfiles[@]}"; do
         # Authenticate as the service account. This allows the promoter to later
--- a/util/multirun.sh
+++ b/util/multirun.sh
@@ -23,10 +23,41 @@ set -o pipefail
 
 usage()
 {
-    echo >&2 "usage: $0 <path/to/cip/binary> [<path/to/manifest.yaml>,<path/to/service-account.json>, ...]"
-    echo >&2 "The 2nd argument onwards are '<manifest>,<service-account>,<service-account>...' strings."
-    echo >&2 "If an execution does not require any service account, pass in '<manifest>,' (notice the trailing comma)."
-    echo >&2
+cat <<EOF >&2
+usage: $0 <path/to/cip/binary> [<MANIFEST>,<KEYFILE>[,<KEYFILE>,...] ...]
+
+MANIFEST: path/to/manifest.yaml
+KEYFILE: path/to/service-account.json
+
+The 2nd argument onwards are
+
+    '<manifest>,<service-account>,<service-account>...'
+
+strings.
+
+If an execution does not require any service account, pass in '<manifest>,'
+(notice the trailing comma).
+
+If you only want to run against manifests that changed in a Git commitish,
+specify the CIP_GIT_REV and CIP_GIT_DIR environment variables. E.g.,
+CIP_GIT_REV=HEAD
+CIP_GIT_DIR=/path/to/git/repo
+
+EOF
+}
+
+# Filter the given list of files by checking if it is a Manifest file. Files
+# must be absolute paths.
+collect_manifests()
+{
+    local cip="$1"
+    shift 1
+
+    for f in "$@"; do
+        if "$cip" -parse-only -manifest="$f"; then
+            echo "$f"
+        fi
+    done
 }
 
 if (( $# < 2 )); then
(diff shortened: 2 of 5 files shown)


listx:
@bartsmykla PTAL?

bartsmykla:
@listx unfortunately I'm not an expert in bash, so I didn't do proper code review of that I'm afraid, but go code lgtm :-)

cip.go
```diff
@@ -57,6 +63,11 @@ func main() {
 		"do not pass '--account=...' to all gcloud calls (default: false)")
 	flag.Parse()
 
+	if *parseOnlyPtr {
+		reg.ParseManifestFromFile(*manifestPtr)
+		os.Exit(0)
```
thockin: Does ParseManifestFromFile call Exit internally?  Otherwise this always succeeds...

util/multirun.sh
```diff
@@ -23,9 +23,41 @@ set -o pipefail
...
+
+# Filter the given list of files by checking if it is a Manifest file. Files
+# must be absolute paths.
+collect_manifests()
+{
+    local cip="$1"
+    shift 1
+
+    for f in "$@"; do
+        if "$cip" -parse-only -manifest="$f"; then
```
thockin: Is it really OK to just ignore args that don't parse?  Why is that not an error?

util/multirun.sh
```diff
@@ -35,27 +67,78 @@ fi
 
 cip="$1"
 shift
+args=("$@")
```
thockin: Why?  The simpler pattern is `for arg` which implies "$@".  E.g.

```
$ function foo() { for arg; do echo ARG: $arg; done }

$ $ foo 1 2 3 
ARG: 1
ARG: 2
ARG: 3

$ foo 1 "2 two" 3 
ARG: 1
ARG: 2 two
ARG: 3
```

util/multirun.sh
```diff
@@ -35,27 +67,78 @@ fi
...
+    if [[ ! -d "${CIP_GIT_DIR:-}" ]]; then
+        echo >&2 "CIP_GIT_DIR not set (must be set if CIP_GIT_REV is set)"
+        usage
+        exit 1
+    fi
+    pushd "${CIP_GIT_DIR}"
+    changed_files=()
+    while IFS= read -r f; do
+        changed_files+=( "$f" )
+    done < <( git diff-tree --no-commit-id --name-only -r "${CIP_GIT_REV}" )
```
thockin: I don't know how this is managing the git checkout so I can't say it's right or not - comments might help the casual reader understand

util/multirun.sh
```diff
@@ -35,27 +67,78 @@ fi
...
-        # call gcloud with the flag `--account=...`. We can allow the service
-        # account creds file to be empty, for testing cip locally (for the case
-        # where the service account creds are already activated).
-        gcloud auth activate-service-account --key-file="${keyfile}"
-        service_accounts+=("$(gcloud config get-value account)")
-    done
+        if [[ -n "$keyfile" ]]; then
+            service_account_keyfiles+=( "$keyfile" )
+        fi
+    done < <( echo "$arg" | cut -d, -f2- | tr , '\n' )
```
thockin: This seems baroque.  Is it simpler to do:

```
A="foo,bar,bat"
read manifest keys < <(echo "$arg" | sed 's/,/ /g')
keyfiles=($keys)
echo M: "$manifest", K: "${keyfiles[@]}"
```

Also doesn't require the trailing comma

cip.go
listx: Yes, it dies with an error if the parse fails. But I can change it to return an error and make the error handling more explicit.

util/multirun.sh
listx: It's OK to ignore because it just removes all files that are not parseable as promoter manifests. It's  "filter(...)" in Python.

util/multirun.sh
listx: I didn't know this idiom. Will change.

util/multirun.sh
listx: Will add some comments here.

util/multirun.sh
listx: Oh nice --- I didn't realize `read` could take on multiple variable name args. Changing.

listx: I just opted for using 'readarray' which avoids the process substitution altogether.

listx:
PTAL?

thockin:
Under what circumstances do we want users passing args that don't parse and
we then ignore?  That seems like a pretty fatal situation to me?

listx:
> Under what circumstances do we want users passing args that don't parse and we then ignore? That seems like a pretty fatal situation to me?

The more complete context is --- users (Prow job config) would be passing args that *do* parse as promoter manifests. We would then filter out those manifests that were not touched in a git commit.

If users pass manifests that don't parse, the promoter would just error out and die (NOP).

util/multirun.sh
```diff
@@ -36,26 +68,99 @@ fi
...
+        usage
+        exit 1
+    fi
+
+    # Get a lit of all files that changed in the given CIP_GIT_REV.
+    # See https://stackoverflow.com/a/424142/437583.
+    changed_files=()
+    while IFS= read -r f; do
+        changed_files+=( "$f" )
+    done < <( git -C "${CIP_GIT_DIR}" diff-tree --no-commit-id --name-only -r "${CIP_GIT_REV}" )
```
thockin: Can this be one-lined as:
```
changed_files=($(git -C "${CIP_GIT_DIR}" diff-tree --no-commit-id --name-only -r "${CIP_GIT_REV}"))

File: util/multirun.sh [hunk 7]
```diff
@@ -23,9 +23,41 @@ set -o pipefail
...
+    '<manifest>,<service-account>,<service-account>...'
+
+strings.
+
+If an execution does not require any service account, pass in '<manifest>,'
+(notice the trailing comma).
+
+If you only want to run against manifests that changed in a Git commitish,
+specify the CIP_GIT_REV and CIP_GIT_DIR environment variables. E.g.,
+CIP_GIT_REV=HEAD
```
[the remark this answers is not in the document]
listx: Hmm. That's a good point. I didn't think of this scenario for some reason --- I could easily just force the user to specify a range of commits. I need to cross-check the Pod Utilities helper in the Prow job to see if that information is even available for this script at runtime.

util/multirun.sh
```diff
@@ -36,26 +68,99 @@ fi
...
+
+    # Get a lit of all files that changed in the given CIP_GIT_REV.
+    # See https://stackoverflow.com/a/424142/437583.
+    changed_files=()
+    while IFS= read -r f; do
+        changed_files+=( "$f" )
+    done < <( git -C "${CIP_GIT_DIR}" diff-tree --no-commit-id --name-only -r "${CIP_GIT_REV}" )
+
+    # Filter those changed files to only have promoter manifests.
+    changed_manifests=$(collect_manifests "${cip}" "${changed_files[@]}")
```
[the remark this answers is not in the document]
listx: Hmm, I think I understand your concern. E.g., it may be the case that we have a valid manifest file, but it gets changed with an errant typo (making it unparseable) in the PR.

I think I can reorder the filtering logic to record an error if a given manifest file argument to multirun.sh fails to parse (this is a condition that should never happen, regardless of Git changed files). If they all parse correctly, then we should optimize by filtering out those files that have not changed.

util/multirun.sh
listx: I think that would not work for filenames with spaces in them.

util/multirun.sh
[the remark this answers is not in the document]
listx: According to the [Pod Utilities docs](https://github.com/kubernetes/test-infra/blob/master/prow/pod-utilities.md), it *merges* the PR into the base branch. So we lose information about "where is HEAD in upstream?" (in order to get a list of commits in the PR by diffing commit-wise against origin/master vs the PR).

The workaround is to just add the upstream github repo as a remote and do a fetch (then just diff against remote); then we can diff against origin/master and local HEAD to get a list of all commits in the PR. A bit finicky, but doable with a handful of shell. The one downside is that the promoter manifest repo must not be a private repo, but I don't think that will be a concern.

Let me add that bit of logic in here...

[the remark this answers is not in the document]
thockin: Or you could cache the md5sum of all the files you have synced in an emptyDir.  It doesn't even need to be super persistent, maybe - if you get moved to a new machine you do a full sync, and then start tracking changes again.  Restarts on the same machine would still be cached.  Or you could use a small volume to keep the cache.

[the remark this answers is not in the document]
listx: That sounds too complicated. I'd like to try out the idea of diffing commits first to see where that leads.

I'd like to add that the bit about no support for private promoter manifest repos is not entirely true because technically the overall Prow bot still has read access (by definition) to the manifest repo (it's just that I am not aware of the way to inject this secret into the Prow job itself). Alternatively, I can add a patch to Prow to make it define an env var that answers the question "what are the commits in the PR?", or "at what commit is upstream HEAD?".

listx:
It turns out that the Prow jobs specify a number of env vars already that point to the target branch SHA and the PR HEAD's SHA. The CIP_GIT_REV stuff has been replaced in favor of CIP_GIT_REV_{START,END}. This means that from the Prow job config, as per https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md, we can do CIP_GIT_REV_START=PULL_BASE_SHA and CIP_GIT_REV_END=PULL_PULL_SHA.

I've tested this script locally, like this:

```
CIP_GIT_REV_START=master CIP_GIT_REV_END=HEAD CIP_GIT_DIR=$HOME/tmp/x CIP_OPTS=-dry-run ./util/multirun.sh bazel-bin/darwin_amd64_stripped/cip manifest.yaml
```

and it seems to work. PTAL

thockin:
I don't know prow internals - the e2e test here is to push an update that includes 2 commits each of which makes one change, and verify that both promotions happen.

How will we be e2e testing this?

listx:
> I don't know prow internals - the e2e test here is to push an update that includes 2 commits each of which makes one change, and verify that both promotions happen.

I don't think we need to involve Github. We could have a manifest repo locally (much like how I tested with `CIP_GIT_DIR=$HOME/tmp/x` with my last comment). "Locally" in this context means, running inside a step in another Prow Job that watches this promoter repository.

> 
> How will we be e2e testing this?

We just need another Prow Job that watches changes to this (promoter source code) repo. That job would reset (test-only) GCRs to a clean state, build a new promoter image, then process promotions by calling multirun.sh in the promoter image against a set of promoter manifests (each manifest would be its own test case).

It's not testing everything (we're not mocking Prow, or Github PRs), but the invariant here is that the e2e test *must* involve running the promoter against real GCRs, so, that is what we should start with.

I would like to add that the complexity around CIP_GIT_REV_{START,END} is only there because Prow doesn't tell us what files changed in a PR. If we modify Prow's Pod Utilities to give us that info, then this script would be much simpler.

thockin:
I am happy to make a GCR just for e2e testing the promoter.  I think that
this component *neeeeeeeds* to be well tested, and any additional time we
invest here is worth it.

It may also be worthwhile to break the script up into functions and have
"integration" tests - e.g. given a git-directory and a start and end range,
list the files that changed.  We can script that with local git.

Before we turn this live for real, we should audit the testing to make sure
we're all happy with it.

listx:
> I am happy to make a GCR just for e2e testing the promoter. I think that this component *neeeeeeeds* to be well tested, and any additional time we invest here is worth it. It may also be worthwhile to break the script up into functions and have "integration" tests - e.g. given a git-directory and a start and end range, list the files that changed. We can script that with local git. Before we turn this live for real, we should audit the testing to make sure we're all happy with it.

FWIW the newest version of multirun.sh just uses a single `git diff --name-only START...END` invocation to find all changed files (the "1 PR = 1 commit" assumption I had really threw me off the right track initially). Given this, I don't think there's an urgency to test the git logic.

Even though we spent a lot of time on multirun.sh, if we're going to write more code (to test the logic in it), I would actually opt for porting the logic to the Go binary directly. It would at least make testing the individual components a lot easier.

listx:
I forgot to add: yes, we should create test GCRs just for testing promotions?

As for the GCR (GCP project) naming, how about:
- k8s-cip-test-staging
- k8s-cip-test-prod1
- k8s-cip-test-prod2
?

The reason there is more than 1 prod is because we already support pushing to multiple destinations in a single promoter manifest.

PR #35, opened 2019-04-03
Added --gen-images flag

bartsmykla:
It's WIP because I'm still working on writing tests and also I'm not sure about the second flag `--output`. I would appreciate any other opinions if we should allow to display generated manifest to stdout without second flag (path for the generated manifest file) or just use one flag (`--create-manifest`) which will expect to give a path for the file to be generated.

Signed-off-by: Bart Smykla <bsmykla@vmware.com>


cip.go
```diff
@@ -72,6 +72,16 @@ func main() {
 		"version",
 		false,
 		"print version")
+	createManifestPtr := flag.Bool(
+		"create-manifest",
+		false,
+		"generating manifest file from all images in destination"+
+			" repositories instead of promoting images")
+	createMfestOutputPtr := flag.String(
```
listx: Can we merge this option with the createManifestPtr option?

If I say `-create-manifest=STDOUT` it should print to STDOUT.
If I say `-create-manifest=<any-other-nonzero-path>` it should save to that path.
If the `-create-manifest` flag is not specified, it's the same as `false` (default).

I think `flag.StringVar` should work for this?

cip.go
```diff
@@ -84,7 +94,7 @@ func main() {
 		os.Exit(0)
 	}
 
-	mfest, err := reg.ParseManifestFromFile(*manifestPtr)
+	mfest, err := reg.ParseManifestFromFile(*manifestPtr, *createManifestPtr)
```
listx: I don't understand why ParseManifestFromFile needs to care about creating manifests, which is already handled by CreateManifest().

cip.go
```diff
@@ -131,6 +147,16 @@ func main() {
 	}
 	sc.ReadDigestsAndTags(mkRegistryListTagsCmd)
 
+	if *createManifestPtr {
+		if err := sc.CreateManifest(mfest, *createMfestOutputPtr); err != nil {
+			fmt.Fprintln(
+				os.Stderr,
+				fmt.Errorf("cannot create manifest: %s", err))
```
listx: We should exit with an error like `os.Exit(1)` here.

cip.go
```diff
@@ -131,6 +147,16 @@ func main() {
...
+		}
+
+		return
```
listx: I would prefer to `os.Exit(0)` here to make it more explicit that we are stopping execution entirely.

lib/dockerregistry/inventory.go
```diff
@@ -102,13 +102,13 @@ func (sc *SyncContext) Fatal(v ...interface{}) {
 }
 
 // ParseManifestFromFile parses a Manifest from a filepath.
-func ParseManifestFromFile(filePath string) (Manifest, error) {
+func ParseManifestFromFile(filePath string, allowEmptyImages bool) (Manifest, error) {
```
listx: The allowEmptyImages is no longer required as it is optionally allowed since 6cdac03ad6eff62f63ab3da91b9913ea95edda22

Gopkg.lock
```diff
@@ -2,16 +2,27 @@
 
 
 [[projects]]
-  digest = "1:4d2e5a73dc1500038e504a8d78b986630e3626dc027bc030ba5c75da257cdb96"
+  digest = "1:ffe9824d294da03b391f44e1ae8281281b4afc1bdaa9588c9097785e3af10cec"
+  name = "github.com/davecgh/go-spew"
```
listx: Looks like this was used for debugging during development only (I don't see it referenced anywhere); please remove.

cip.go
listx: Hmm maybe it would be more elegant if we could have this behavior:

Passing `-create-manifest` alone should print to STDOUT.
Passing `-create-manifest=<path>` will save to that path.

listx:
I just realized that the underlying command we use (`gcloud container images list --repository=gcr.io/project --format=json`) won't work to fully populate all images into the MasterInventory. E.g., consider these two images:

```
gcr.io/foo:latest
gcr.io/foo/bar:latest
```

They are both under the project `foo`, but are under *different* repositories (a registry = set of repositories). So running `gcloud container images list --repository=gcr.io/foo --format=json` will only list `latest` and *NOT* the `foo/bar` image!

This means that currently we can only snapshot toplevel images that are *not* nested under subdirectories.

We could instead use a direct HTTPS request like this (pasting a curl version below, but we could run a native HTTPS request in Go):
`curl -u "oauth2accesstoken:$(gcloud auth print-access-token)" https://gcr.io/v2/<PROJECT>/<PATH/TO/IMAGE>/tags/list` and use that to walk all repositories in the registry, because that endpoint gives us a list of all child repos, which we can use to make this call recursively.

The promoter already assumes that the given credentials in the manifest are *already* activated, so it's just a matter of getting the oauth2accesstoken and making the HTTPS request.

As a side benefit, fixing this part will allow us to support images with path prefixes in them in our manifest (we currently don't support them, as I learned today...).

cip.go
bartsmykla: removed

cip.go
bartsmykla: Done :-)

lib/dockerregistry/inventory.go
bartsmykla: Done :-)

Gopkg.lock
bartsmykla: Yes, my bad - removing.

bartsmykla: Done

cip.go
bartsmykla: Done

cip.go
bartsmykla: As discussed, I completely removed option to save to the path.

cip.go
bartsmykla: Done

bartsmykla:
Hi @listx as we discussed last week during the call, I have simplified things and removed all logic related to merging etc.
I have also changed the name of a flag to `gen-images`, but we can change it to something else, because I couldn't figure out any better one (being non-native English speaker can be tricky :-))
Currently if you have in your manifest only one destination repository and use flag `--gen-images`, all images from the repository will be displayed in your stdout, so you can copy or pipe it to the file (we can improve it in the future).

I'm also not sure about the error message here: https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/35/files#diff-78dfc203b762cc0ff3a086a0c34dfd95R621 and would appreciate any help with that.

lib/dockerregistry/inventory.go
```diff
@@ -587,6 +589,60 @@ func ToPQIN(registryName RegistryName, imageName ImageName, tag Tag) string {
...
+			ImageName: name,
+			Dmap:      dmap,
+		})
+	}
+
+	return result
+}
+
+// getImagesFromDestRegistry returns all images from not src repository
+func (sc *SyncContext) getImagesFromDestRegistry(
```
listx: Can we simplify this to only read out 1 registry, specified from the CLI as an argument for `--gen-images` ? There is no need to discern "src" or "dest" registries. You would just call ReadRepository() against some registry, which would populate sc.Inv for us for that 1 registry. Then you can use the getImageSliceFromMap() conversion function you wrote to generate the `images:` field.

A better flag name might be just `--snapshot`.

lib/dockerregistry/inventory.go
```diff
@@ -587,6 +589,60 @@ func ToPQIN(registryName RegistryName, imageName ImageName, tag Tag) string {
 	return string(registryName) + "/" + string(imageName) + ":" + string(tag)
 }
 
+func getImagesSliceFromMap(images map[ImageName]DigestTags) []Image {
```
listx: I think a better signature would be:

```
func (rii RegInvImage) toImagesSlice() ([]Image) { ... }
```

lib/dockerregistry/inventory.go
listx: To clarify, we could indeed read out multiple registries and take their intersection (`--gen-images` could take a list of registries instead of just 1), but I think in actual practice it would probably not be used. The use-case we are trying to support is snapshotting `gcr.io/google-containers`.

PR #40, opened 2019-04-15
allow image renames, use HTTP API for reading registry state, vendor deps

### LINKED ISSUE #20: allow remapping paths during promotion (do not force src registry to share same namespace as dest registry)
Currently we expect the source image to have the *exact* same path as the destination path. This limits flexibility. We want to be able to optionally specify the target path in the destination, to allow it to be different than the source path.

This will require a slight tweak to the promoter manifest YAML.

listx:
Basically this does 2 things, allow image renames and also changes the "read state" logic to use HTTP API calls instead of `gcloud`.

Fixes #20 

/hold


listx:
@bartsmykla PTAL

listx:
Oh one more thing, garbage collection logic will need tweaking now that we have renames. But GC is not super critical, so it's not a blocker.

cip.go
```diff
@@ -129,47 +153,83 @@ func main() {
...
+			// handle err... Maybe just print a warning message. Because this
+			// and other operations will take time, and the service is mutating
+			// live (even though this process should be the only one, others might push images that are live.)...
+
+			// TODO: If we detect that we cannot read a repo correctly, we may have to REMOVE that repo's contents from the Manifest, because, e.g., it may be that the manifest defines tags in...
+			//
+			// Basically the idea is that we have to perform safety
+			// modifications to our set data structure info before doing things
+			// like garbage collection, _if we have an incomplete view of the world_.
+		}
```
bartsmykla: If `err != nil` the function won't stop, are you sure that `httpReq` will be in expected state?

lib/dockerregistry/types.go
```diff
@@ -294,3 +352,21 @@ func (riid RegInvImageDigest) ToRegInvImageTag() RegInvImageTag {
...
 	return riit
 }
+
+// ToRegInvImageDigest converts a RegInvImageTag to a RegInvImageDigest.
+func (riit RegInvImageTag) ToRegInvImageDigest() RegInvImageDigest {
+	riid := make(RegInvImageDigest)
+	for imageTag, digest := range riit {
+		id := ImageDigest{}
+		id.ImageName = imageTag.ImageName
+		id.Digest = digest
```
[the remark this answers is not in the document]
listx: I agree, but I think we should postpone this until after this PR --- because we can do it in the other conversion functions in this file. Let's have a refactor/cleanup PR that does all of that in one go.

cip.go
listx: I have to handle this error --- I think just aborting entirely is OK. We could get fancy and perform a limited promotion (basically reducing the set data structure to a subset), but I don't think that's what we want. We don't want to print SUCCESS at the end when a http error prevented us from promoting the 1 image we added to our manifest.

listx:
Sorry about the various ugly print statements littered everywhere --- they will all have to be removed of course (esp. the part about printing tokens to stdout!!!).

I will add tests today, hopefully, for the new rename logic.

bartsmykla:
No worries, everything else looks good for me :-)

listx:
@bartsmykla I added a bunch of tests, also taking into account how renames affect garbage collection.

PR #44, opened 2019-04-24
Change imports to use sigs.k8s.io vanity URL

justinsb:
This means that it matches the normal checkout path for k8s repos.


listx:
/lgtm
/approve

I don't care either way, but if anything this makes the imports shorter, so it's a win in my book.
Thanks Justin!

PR #46, opened 2019-04-24
Explicitly tag all yaml fields that are serialized

justinsb:
Avoids the implication on reading that these are not serialized,
avoids future problems if we change the field name.


listx:
/lgtm
/approve

All else being even, I think more explicit = better, so thanks for this change.

PR #47, opened 2019-04-24
Switch to k8s-standard SCRIPT_ROOT approach

justinsb:
I suspect there was nothing wrong with the previous way, but this is
the k8s-way.


listx:
Normally for cosmetic changes like this I don't have strong opinions. I'm just a little sad that my carefully-written comments will be deleted. Oh well.

/lgtm
/approve

PR #48, opened 2019-04-24
Promotion for non-container artifacts

justinsb:


justinsb:
Working, but needs tests / more validation etc.  We probably should also refactor out the model classes from the syncing code.

cc @thockin @listx

listx:
Can you update the README to talk about non-container promotions?

Meanwhile I will start my code review in earnest.

listx:
Here is a version of this PR with some fixes around merge conflicts and also bazel deps (I could not get your PR to build without these changes). https://github.com/listx/k8s-container-image-promoter/commit/576e804d586138f742d5621da10299fc29255164

lib/dockerregistry/filestore.go
```diff
@@ -0,0 +1,408 @@
...
+		return fmt.Errorf("error creating temp file: %v", err)
+	}
+	tempFilename := f.Name()
+	defer func() {
+		if err := os.Remove(tempFilename); err != nil {
+			klog.Warningf("unable to remove temp file %q: %v", tempFilename, err)
+		}
+	}()
+
+	// Download to our temp file
```
listx: Ouch --- is there no way of promoting files server-side? That's how we promote images, surely there is a way to do the same with files...

From `gsutil help cp` I found:

```
COPYING IN THE CLOUD AND METADATA PRESERVATION
  If both the source and destination URL are cloud URLs from the same
  provider, gsutil copies data "in the cloud" (i.e., without downloading
  to and uploading from the machine where you run gsutil). In addition to
  the performance and cost advantages of doing this, copying in the cloud
  preserves metadata (like Content-Type and Cache-Control). In contrast,
  when you download data from the cloud it ends up in a file, which has
  no associated metadata. Thus, unless you have some way to hold on to
  or re-create that metadata, downloading to a file will not retain the
  metadata.
```

I *strongly* support the use of `gsutil` over downloading to local disk and re-uploading. We already have examples of shelling out to `gcloud` in this project so it's not a big stretch for shelling out to `gsutil` (or even better, just using the raw API calls, if they exist, for copying files server-side from one location to another).

It would also make the tests much easier to write, I think (no need to fake files on disk...).

lib/dockerregistry/filestore.go
```diff
@@ -0,0 +1,408 @@
...
+	return files, nil
+}
+
+// syncFileOp defines a synchronization operation
+type syncFileOp interface {
+	Run(ctx context.Context) error
+}
+
+// computeNeededOperations determines the list of files that need to be copied
+func (p *FilestorePromoter) computeNeededOperations(source, dest map[string]*syncFileInfo, destFilestore syncFilestore) ([]syncFileOp, error) {
```
listx: It will be much nicer if we can reuse (or at least emulate) the set-difference logic we use for figuring out what needs to be promoted (copied), as in https://github.com/kubernetes-sigs/k8s-container-image-promoter/blob/f2a8a6e4557beada0a6cebd45a9c8d2b8433d951/lib/dockerregistry/set.go. For example we could just serialize the metadata for the file into a single string key, which would be the value in a map (the keys would be the filenames). Then the logic to figure out what files need to be copied is just a set minus operation, as in https://github.com/kubernetes-sigs/k8s-container-image-promoter/blob/edca50185d79a18857b21e2ae8bdb4d346b0887e/lib/dockerregistry/inventory.go#L1067.

lib/dockerregistry/types.go
```diff
@@ -154,6 +154,14 @@ type Manifest struct {
 	// registries, in which case we would have more than just Src/Dest.
 	Registries []RegistryContext
 	Images     []Image `yaml:"images,omitempty"`
+
+	// Filestores contains the source and destination (Src/Dest) filestores.
+	// Filestores are (for example) GCS or S3 buckets.
+	// It is possible that in the future, we support promoting to multiple
+	// filestores, in which case we would have more than just Src/Dest.
+	Filestores []*Filestore `yaml:"filestores,omitempty"`
```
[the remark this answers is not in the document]
justinsb: Mostly just paranoid habit.  If we don't use pointers, it's all too easy to (shallow) copy the object on every loop iteration.  We should switch to something like this:

```
for i := range o.Filestores {
file := &o.Filestores[i]
}
```

That said, the tradeoff is less efficient memory packing, so I'll switch to the packed struct.  Also because I don't think this is performance critical code, so a few extra copies won't really matter!

justinsb:
(Finaly) got around to fixing the last few things:

* Fixed up the issues in your PR review (thanks!)
* Added an example to the docs
* Created a real-world example at https://github.com/kubernetes/k8s.io/pull/326/files
* I thought maybe we could compromise on 100 characters for max line-width?  At 80 characters I just found myself rewriting error messages to be concise, at the expense of clarity.

listx:
> (Finaly) got around to fixing the last few things:
> 
> * Fixed up the issues in your PR review (thanks!)
> * Added an example to the docs
> * Created a real-world example at https://github.com/kubernetes/k8s.io/pull/326/files
> * I thought maybe we could compromise on 100 characters for max line-width?  At 80 characters I just found myself rewriting error messages to be concise, at the expense of clarity.

I'll take a look at all of your recent changes, but for now regarding the 80 char thing, I think as a compromise we should just add an exclude pattern to the "lll" linter to exclude lines that have the pattern "fmt.Errorf". This should be doable according to https://github.com/walle/lll (golangci-lint, which we use, uses lll).

lib/dockerregistry/filestore.go
listx: Given the new upcoming changes for how the promoter works, this argument for reusing set-difference logic no longer holds true. I think the set-difference stuff will be used later in a different tool that performs audits/backups of entire registries/repositories.

lib/dockerregistry/filestore.go
```diff
@@ -0,0 +1,408 @@
...
+
+	// UseServiceAccount should be set to true to enable us to assume service accounts
+	UseServiceAccount bool
+}
+
+// syncFileInfo holds information about a file during the synchronization operation
+type syncFileInfo struct {
+	RelativePath string
+	AbsolutePath string
+	MD5          string
```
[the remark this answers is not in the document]
listx: Ah thanks for clarifying. I think just adding a comment that MD5 is the common hash algorithm used across S3/GCS is fine.

lib/dockerregistry/filestore.go
```diff
@@ -0,0 +1,408 @@
...
+func (o *copyFileOp) String() string {
+	return fmt.Sprintf("COPY %q to %q", o.Source.AbsolutePath, o.Dest.AbsolutePath)
+}
+
+func computeSHA256ForFile(filename string) (string, error) {
+	f, err := os.Open(filename)
+	if err != nil {
+		return "", fmt.Errorf("error re-opening temp file %q: %v", filename, err)
+	}
+	defer f.Close()
```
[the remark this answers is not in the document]
listx: Ok sounds good.

lib/dockerregistry/filestore.go
listx: I guess we have to download to perform sha256 validation (because S3/GCS do not store this info themselves)? If so, then I think what you are proposing here makes sense as we should use a stronger hash than MD5.

I agree that hashing the compressed content is not a great idea --- *unless* we also perform the compression ourselves in a deterministic manner (which I am not even sure is a thing).

justinsb:
Reformatted to 80 characters max-width :-)

lib/dockerregistry/types.go
```diff
@@ -237,6 +245,26 @@ type RegistryContext struct {
...
+	Src            bool   `yaml:"src,omitempty"`
+}
+
+// File holds information about a file artifact.
+// File artifacts are copied from a source Filestore to N destination Filestores
+type File struct {
+	// Name is the relative path of the file, relative to the Filestore base
+	Name string `yaml:"name"`
+	// SHA256 holds the SHA256 hash of the specified file (hex encoded)
+	SHA256 string `yaml:"sha256,omitempty"`
```
listx: I stated this earlier somewhere else but it would be nice to share the same "style" as we have for the image promoter manifest, where we use a string `sha256:xxxx..`, mainly for consistency; then the field name would instead be `Hash` or some other generic field to denote a checksum.

README.md
```diff
@@ -24,6 +24,45 @@ images:
...
+  sha256: 2d4f26491e0e470236f73a0b8d6828db017eab988cd102fc19afe31f1f56aff7
+- name: vegetables/beetroot.png
+  sha256: 160b98e27ec99f77efe01e2996fa386f2b2aec552599f8bd861be0a857e7f29f
+```
+
+`filestores` is the equivalent of `registries`, and lists the buckets from which
+the promoter should read or write files.  `files` is the quivalent of `images`,
+and lists the files that should be promoted.
+
+`filestores` supports `service-account`, and it also supports relative paths -
```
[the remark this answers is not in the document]
listx: SGTM

PR #50, opened 2019-05-28
Extract logic to get token for ServiceAccount to its own function

justinsb:
We want to use this more widely, so split it out for easier reuse.


listx:
Thank you for fixing up my laziness.

/lgtm
/approve

PR #51, opened 2019-05-28
Simplify gcloud execution

justinsb:
We don't need the streaming wrapper for a token, and this avoids complexity in closing the command.


listx:
I liked the uniformity of using the stream for all external (shell exec) calls, but I think this is a small enough area that breaking uniformity is OK. Besides, it does simplifies things as stated in your commit message.

/lgtm
/approve

PR #53, opened 2019-05-28
Replace log with klog

justinsb:
klog is a little more k than just log, but mostly splitting this out
into its own commit to make file promotion smaller.

### DIFF
--- a/cip.go
+++ b/cip.go
@@ -19,11 +19,11 @@ package main
 import (
 	"flag"
 	"fmt"
-	"log"
 	"net/http"
 	"os"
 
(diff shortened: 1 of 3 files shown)


listx:
This is nice, thanks.

I was waiting for someone to say "we should use X logging library instead of the homegrown stuff in https://github.com/kubernetes-sigs/k8s-container-image-promoter/blob/5fb3b400034911b811880310b2b45257874df6a6/lib/dockerregistry/inventory.go#L72" and it looks like that time has come.

We should use klog everywhere for simplicity...

/lgtm
/approve

Issue #5, opened 2019-02-16
move CI logic into separate scripts


listx:
I.e., strip out steps from cloudbuild.yaml and elsewhere into standalone scripts.

amy:
@listx I have time to tackle this. Can you assign this issue to me.
Is there anything special other than these instructions that I need to follow in order to sanity test the moved logic: https://cloud.google.com/cloud-build/docs/create-custom-build-steps

dims:
@amy can you please request membership to the kubernetes-sigs org?

amy:
@dims Ah nevermind. I can't be assigned issues yet.

dims:
@amy please consider yourself assigned :)

listx:
@amy  Whatever logic in cloudbuild.yaml should be stripped out into separate shell scripts, and then moved into a new Prow job as per #6.

listx:
This is now tied to #8.

listx:
This is fixed. In a twist of irony, there isn't much shell code in the codebase any more, because we delegate so much to bazel. So, there isn't much "CI logic to put into separate shell scripts" to begin with.

/close

Issue #8, opened 2019-02-16
Create e2e test for use in CI (to actually promote/demote images in Docker Registries)


listx:
The idea is, we can create a test (as a Prow Job) to populate 2 test GCRs --- a fake "staging" and a fake "production" GCR. Then we would rotate through 1 or more different promoter manifests, with various flags passed to the promoter, and expect the target registry to look a certain way after the promoter finishes.

We could do the above multiple times in the test. This would be great to test things like garbage collection and whatnot, and other more advanced features of the promoter that might be implemented in the future (e.g., multi-stage promotions, replication, etc.).

The basic ingredients are:
- Have some logic to create real Docker images deterministically (bazel already does this...)
- Populate a staging GCR with these images
- Run a just-compiled promoter (as a presubmit) and run it against a battery of various promoter manifest.yamls.

All of the above could run in a Prow step. It would simply fetch an image with gcloud + bazel, then compile the promoter and then run it.

The integration test would enable developers (who might not have a Docker Registry to play with or have GCP accounts set up) to make changes (such as refactorings) without fear of modifying the behavior of the promoter.

The original idea of the various components of how an integration test might work is documented [here](https://docs.google.com/document/d/1WGFt5ck_XGf71PO4c87UMPVU_4Q7AV-7tRV4Z6wmZL4/edit?usp=sharing).

listx:
For the actual images used during testing, another alternative is to use a set of known images and to just check these in as long as they are reasonably small. Ultimately, we need to test for:

- plain images
- manifest lists (aka "fat manifests") that reference multiple images
- windows images
- images with foreign layers

For the more exotic images like windows images and foreign layers (ones we might have trouble generating deterministically), the options are:
- find (very small!) versions of these and check them into our source tree directly
- find small images and copy them into a safe GCR location that we own

listx:
I will create a short design doc for this before starting work on this.

listx:
The design doc is here: https://docs.google.com/document/d/1PgLI7OCEd09qLLz9yGmY2KK0uxhROc9GGC-PkwgnKN4/edit?usp=sharing

listx:
I got some code written over KubeCon for this. I have a crude script that can delete all images from the  test-only repos, and also repopulate it with deterministic docker images created with bazel, plus a manifest list.

I will continue to work on this in my own spare time, but feel free to ping me to ask for updates either here or on Slack (#wg-k8s-infra).

[...]

PR #55, opened 2019-05-31
moved cloudbuild scripts to util/cloudbuild

amy:
Addresses: https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/5

Still figuring out how to sanity test this (as in, I'm assuming it'll create resources if I use a personal GCP account.)

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -9,28 +9,7 @@ steps:
   entrypoint: "bash"
   args:
   - "-c"
-  - |
-    set -o errexit
-    set -o nounset
-    set -o pipefail
-    set -o xtrace
-
-    shopt -s dotglob
-    # Remove the git archive that GCB auto-injects to /workspace, because it's
-    # not a real Git repository (it's just a tarball). This is required because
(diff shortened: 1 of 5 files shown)


listx:
Hi Amy, sorry but the cloudbuild.yaml isn't actually used by this project (indeed it can be deleted even, it's just that no one has bothered to (yet)).

The issue #5 that you referenced is missing an update --- we have since moved to using Prow for running workloads for the promoter. I am in favor of just using Prow for everything if possible, instead of splitting things between Prow and GCB. This is because there is nothing stopping us from running builds/tests of the promoter itself (the logic that you see in cloudbuild.yaml) from within Prow. We would just add another Prow job entry in https://github.com/kubernetes/test-infra/tree/master/config, and have Prow watch this github project to run CI logic on every presubmit (similarly for adding e2e tests, I would just add on more logic into that same (new) Prow job for https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/8.

Meanwhile, you can take the spirit of what you have now and try to create the Prow job itself. The issue for that is #6.

Closing this for now.

PR #56, opened 2019-06-01
Migrate logging to klog

### LINKED ISSUE #54: Use klog everywhere for logging
Delete the (basic) homegrown stuff in https://github.com/kubernetes-sigs/k8s-container-image-promoter/blob/5fb3b400034911b811880310b2b45257874df6a6/lib/dockerregistry/inventory.go#L72

Jukie:
Fixes #54 
Removes homegrown logging stuff and adds klog as a dependency

### DIFF
--- a/lib/dockerregistry/inventory.go
+++ b/lib/dockerregistry/inventory.go
@@ -31,6 +31,7 @@ import (
 	yaml "gopkg.in/yaml.v2"
 
 	"github.com/google/go-containerregistry/pkg/v1/google"
+	"k8s.io/klog"
 	cipJson "sigs.k8s.io/k8s-container-image-promoter/lib/json"
 	"sigs.k8s.io/k8s-container-image-promoter/lib/stream"
 )
@@ -67,60 +68,6 @@ func MakeSyncContext(
 		RenamesDenormalized: rd,
 		SrcRegistry:         srcRegistry,
 		RegistryContexts:    rcs}, nil
-}
-
-// Basic logging.
-
-// Infof logs at INFO level.
-func (sc *SyncContext) Infof(s string, v ...interface{}) {
-	if sc.Verbosity > 2 {
-		fmt.Printf(s, v...)
-	}
-}
-
-// Warnf logs at WARN level.
-func (sc *SyncContext) Warnf(s string, v ...interface{}) {
-	if sc.Verbosity > 1 {
-		fmt.Printf(s, v...)
-	}
-}
-
-// Errorf logs at ERROR level.
-func (sc *SyncContext) Errorf(s string, v ...interface{}) {
-	if sc.Verbosity > 0 {
-		fmt.Printf(s, v...)
-	}
-}
-
-// Fatalf logs at FATAL level.
-func (sc *SyncContext) Fatalf(s string, v ...interface{}) {
-	fmt.Printf(s, v...)
-}
-
-// Info logs at INFO level.
-func (sc *SyncContext) Info(v ...interface{}) {
-	if sc.Verbosity > 2 {
-		fmt.Println(v...)
-	}
-}
-
-// Warn logs at WARN level.
-func (sc *SyncContext) Warn(v ...interface{}) {
-	if sc.Verbosity > 1 {
-		fmt.Println(v...)
-	}
-}
-
-// Error logs at ERROR level.
-func (sc *SyncContext) Error(v ...interface{}) {
-	if sc.Verbosity > 0 {
-		fmt.Println(v...)
-	}
-}
-
-// Fatal logs at FATAL level.
-func (sc *SyncContext) Fatal(v ...interface{}) {
-	fmt.Println(v...)
 }
 
 // ParseManifestFromFile parses a Manifest from a filepath.
@@ -708,12 +655,12 @@ func (sc *SyncContext) ExecRequests(
 	go func() {
 		for reqRes := range requestResults {
 			if len(reqRes.Errors) > 0 {
-				sc.Errorf(
+				klog.Errorf(
 					"Request %v: error(s) encountered: %v\n",
 					reqRes.Context,
 					reqRes.Errors)
 			} else {
-				sc.Infof("Request %v: OK\n", reqRes.Context.RequestParams)
+				klog.Infof("Request %v: OK\n", reqRes.Context.RequestParams)
 			}
 		}
 	}()
@@ -863,7 +810,7 @@ func (sc *SyncContext) GetLostImages(mfest Manifest) RegInvImageDigest {
 	// lost = all images that cannot be found from src.
 	lost := mfest.ToRegInvImageDigest().Minus(src)
 	if len(lost) > 0 {
-		sc.Errorf(
+		klog.Errorf(
 			"ERROR: Lost images (all images in Manifest that cannot be found"+
 				" from src registry %v):\n",
 			sc.SrcRegistry.Name)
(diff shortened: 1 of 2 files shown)


Jukie:
cla signed

Jukie:
@listx Forgive the newbie question but where do I need to add the klog dependency to make bazel happy?

```
$ make test
bazel test --test_output=all //lib/...
INFO: Analyzed 5 targets (0 packages loaded, 0 targets configured).
INFO: Found 4 targets and 1 test target...
ERROR: /go/src/github.com/jukie/k8s/lib/dockerregistry/BUILD.bazel:3:1: GoCompile lib/dockerregistry/linux_amd64_stripped/go_default_library%/sigs.k8s.io/k8s-container-image-promoter/lib/dockerregistry.a failed (Exit 1) compile failed: error executing command bazel-out/host/bin/external/io_bazel_rules_go/go/tools/builders/linux_amd64_stripped/compile -sdk external/go_sdk -installsuffix linux_amd64 -src lib/dockerregistry/inventory.go -src ... (remaining 22 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
GoCompile: missing strict dependencies:
	/root/.cache/bazel/_bazel_root/da3a1ddc4d3ac9f166d69bc641f23111/sandbox/processwrapper-sandbox/79/execroot/__main__/lib/dockerregistry/inventory.go: import of "k8s.io/klog"
Known dependencies are:
	sigs.k8s.io/k8s-container-image-promoter/lib/container
	sigs.k8s.io/k8s-container-image-promoter/lib/json
	sigs.k8s.io/k8s-container-image-promoter/lib/stream
	github.com/google/go-containerregistry/pkg/v1/google
	gopkg.in/yaml.v2
Check that imports in Go sources match importpath attributes in deps.
INFO: Elapsed time: 0.486s, Critical Path: 0.07s
INFO: 1 process: 1 processwrapper-sandbox.
FAILED: Build did NOT complete successfully
//lib/dockerregistry:go_default_test                                  NO STATUS

FAILED: Build did NOT complete successfully
Makefile:15: recipe for target 'test' failed
make: *** [test] Error 1
```

listx:
Hi Jukie, I think it's worth squashing all of this into 1 commit (or, 2 -- business logic changes into 1, and all dependency/bazel updates into another).

The short answer is that we use `gazelle` to make Bazel understand new Go dependencies. This was first described in the commit message for https://github.com/kubernetes-sigs/k8s-container-image-promoter/commit/24dd1cfbd71883432406f28f56daf9219a146f67, when we first moved to go modules:

```
        # Remove vendoring logic.
        rm Gopkg.*
        rm -rf vendor
        rm lib/**/BUILD.bazel

        # Get vgo.
        go get -u golang.org/x/vgo

        # Populate go.mod, go.sum files.
        vgo get

        # Generate the BUILD.bazel files under lib/.
        bazel run //:gazelle

        # Update WORKSPACE to point to the vgo-ed dependencies.
        bazel run //:gazelle -- update-repos -from_file=go.mod
```

The related rules around updating go modules has been codified into our Makefile', "update" rule:

```
update:                                                                                                                                                                            
    # Update go modules (source of truth!).
    GO111MODULE=on go mod verify
    GO111MODULE=on go mod tidy

    # Update bazel rules to use these new dependencies.
    bazel run $(BAZEL_BUILD_OPTS) //:gazelle -- update-repos -from_file=go.mod

    # Fixup vendor/ folder to make golangci-lint happy.
    rm -rf vendor
    GO111MODULE=on go mod vendor
```

That being said, I just realized that the current master branch fails with a build failure (doh...). Mind if I fix up the bazel failure on master branch first to take care of the dependencies? Then you could rebase your PR.

Jukie:
Sure, yeah I noticed that too. I'll rebase whenever it's ready.

Jukie:
@listx any updates here?

listx:
Hmm lemme poke the reviewers for #57

Jukie:
Ready for review :)

lib/dockerregistry/inventory.go
```diff
@@ -69,60 +70,6 @@ func MakeSyncContext(
 		RegistryContexts:    rcs}, nil
 }
 
-// Basic logging.
-
-// Infof logs at INFO level.
-func (sc *SyncContext) Infof(s string, v ...interface{}) {
-	if sc.Verbosity > 2 {
```
listx: The old system assigned the value of verbosityPtr to sc.Verbosity, like this: https://github.com/kubernetes-sigs/k8s-container-image-promoter/blob/cca667f4e15627f654379a7a4ceacaa83d89d570/cip.go#L130. Can you somehow make our new use of klog observe the intent of verbosityPtr as well?

Jukie: Would something like the following be sufficient? I can't think of a better way to implement it

Instread of:
```
if *garbageCollectPtr {
    klog.Infof("---------- BEGIN GARBAGE COLLECTION: %s ----------\n",
        *manifestPtr)
...
```
Use:
```
if *garbageCollectPtr {
    if sc.Verbosity > 2 {
        klog.Infof("---------- BEGIN GARBAGE COLLECTION: %s ----------\n",
			      *manifestPtr)
    }
...
```

From https://github.com/kubernetes-sigs/k8s-container-image-promoter/blob/27b1130f3e64859b874d04662551c28a50d332b2/cip.go#L224-#L226

listx: I was wondering if there is some klog best practice that we are missing here. Some research turned up the fact that klog is a fork of glog, which has support for verbosity levels https://godoc.org/github.com/golang/glog#V, but apparently I suppose klog doesn't support verbosity levels according to https://github.com/kubernetes/klog/issues/35.

So maybe for now instead of deleting all of the lines 72-125, maybe you can just replace the calls to `fmt.Println()` with the appropriate calls to klog. Wdyt?

Jukie: Yeah that's probably the most efficient way, I couldn't think of a way to support verbosity levels without an if block for every call. I'll update my branch.

I wonder what the reasoning was for _removing_ verbosity levels support, hopefully that gets brought back.

Jukie: Correction, I do still see it here - https://github.com/kubernetes/klog/blob/master/klog.go#L1107-#L1134 - so I'll see if this is viable.

listx: SG. I would like to add that according to https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md the klog.Infof() has multiple logging levels (e.g., `klog.V(1)`). I think we can just use those, probably.

justinsb:
What's the status of this one @listx?  I think this is a good change, in that it achieves consistency of logging.  While the verbosity flag may be nicer, I'd say that consistency is more valuable, and we should just use the normal klog/glog behaviour.

I propose we merge this, reduce log levels where the logging is excessive, and remove the verbosity flag.

listx:
> What's the status of this one @listx? I think this is a good change, in that it achieves consistency of logging. While the verbosity flag may be nicer, I'd say that consistency is more valuable, and we should just use the normal klog/glog behaviour.
> 
> I propose we merge this, reduce log levels where the logging is excessive, and remove the verbosity flag.

SG. I was waiting for the conflict to be resolved. I'll get around to it later today if no one else does. The main thing to keep in mind is that the dependency updates should be minimal.

listx:
@Jukie closing this in favor of #67

PR #57, opened 2019-06-04
fix `make build` failure due to un-bazel-ified "klog" dependency

listx:


listx:
/cc @justinsb @Jukie 

Related to #56

PR #60, opened 2019-06-22
simplify snapshotting repositories

listx:
The changes here are in preparation for the e2e test. The new `-snapshot` flag allows us to snapshot repository contents (the same contents always output the same YAML, because we sort the hashmap items). This ability to snapshot repositories gives us a "fingerprint" of a GCR, so that we can use predefined fingerprints to check against a live repository (the test-only GCRs) during tests.

Also update dependencies.


cip.go
```diff
@@ -186,14 +209,11 @@ func main() {
...
-		if err := sc.GenImages(mfest); err != nil {
-			fmt.Fprintln(
-				os.Stderr,
-				fmt.Errorf("cannot create manifest: %s", err))
-			os.Exit(1)
-		}
-
+	if len(*snapshotPtr) > 0 {
+		var rii reg.RegInvImage
+		rii = sc.Inv[mfest.Registries[0].Name]
```
justinsb: Nit: `rii := ...` is easier than declaring the var and then immediately assigning it

lib/dockerregistry/inventory.go
```diff
@@ -784,68 +784,69 @@ func ToPQIN(registryName RegistryName, imageName ImageName, tag Tag) string {
...
+	sort.Slice(images, func(i, j int) bool {
+		return images[i].name < images[j].name
+	})
 
-func getSrcRegistryFromManifest(manifest Manifest) RegistryContext {
-	var srcRegistry RegistryContext
-	for _, registry := range manifest.Registries {
-		if registry.Src {
-			srcRegistry = registry
+	var b strings.Builder
```
justinsb: Why not just use `yaml.Marshal` (from `sigs.k8s.io/yaml`)?

justinsb: Or ... can we make this format match the format of the manifest?  This could also be super-handy for creating the manifest, I think!

cip.go
```diff
@@ -83,11 +83,14 @@ func main() {
...
-		"gen-images",
-		false,
-		"generating all images in destination"+
-			" repository instead of promoting images")
+	snapshotPtr := flag.String(
+		"snapshot",
+		"",
+		"read all images in a repository and print to stdout")
+	snapshotSvcAccPtr := flag.String(
+		"snapshot-service-account",
```
justinsb: Maybe it should just be a separate command?  (In future?)

justinsb:
Looks good - one question about the format, but I'm going to lgtm & hold so as not to slow this down any further.  (Feel free to unhold if you disagree / would rather address in a subsequent PR)

/lgtm
/hold

cip.go
listx: Heh, I keep forgetting about `:=`. Ack

lib/dockerregistry/inventory.go
listx: > Why not just use `yaml.Marshal` (from `sigs.k8s.io/yaml`)?

This is because we want to deterministically (alphabetically) sort the map entries (these are the items under `dmap`, as well as the images in the list. I did some trial runs with `yaml.Marshal` and the items were not listed in a sorted fashion --- hence this scheme.

I could use yaml.MapSlice{} or something but this was simpler.

That said... does `sigs.k8s.io/yaml` already do this kind of sorting automatically? But at the same time I do want `name:` to come before `dmap:`, so... I want *mostly* alphabetical sorting of the key elements.

listx: > Or ... can we make this format match the format of the manifest? This could also be super-handy for creating the manifest, I think!

It already matches the format of a subset of the promoter manifest (`images` key portion). We can use it later to build an complete manifest dynamically, I suppose.

cip.go
listx: As in like https://github.com/spf13/cobra (which supports multi-command option handling, a la kubectl)? YES. This is a no-brainer, but it is low on the priority list.

cip.go
```diff
@@ -147,17 +169,18 @@ func main() {
 	// the manifest desires a completely blank registry. In practice this would
 	// almost never be the case, so given a fully-parsed manifest with 0 images,
 	// treat it as if -parse-only was implied and exit gracefully.
-	if len(mfest.Images) == 0 {
-		fmt.Println("No images in manifest --- nothing to do.")
-		os.Exit(0)
-	}
+	if len(*snapshotPtr) == 0 {
+		if len(mfest.Images) == 0 {
+			fmt.Println("No images in manifest --- nothing to do.")
```
amy: Small nit that doesn't have to be addressed in this PR. I see a lot of `fmt.Print()` around. Do we want to use a logger?

amy: Small nit that doesn't need to be addressed in this PR. I see a lot of `fmt.Print` in this file. Do we want to use a logger?

listx: Hmm well we do want to always print that string regardless of the logging level. Not sure if that makes sense in a logging-level context.

Even if we should use a logger, I think that can be a separate PR (don't wanna change the hash of this commit to throw off #62 which depends on this).

cip.go
justinsb: Cool, SGTM

lib/dockerregistry/inventory.go
justinsb: Ah - ordering makes sense, and I love that it matches the input format.  We should put that in a tip somewhere as I don't want to tell you how I was building the list when I've been doing this...

justinsb:
Thanks for looking at the feedback!

/lgtm

cip.go
listx: I made an issue for this: #61

PR #62, opened 2019-06-26
basic e2e testing framework

listx:
Please approve/merge #60 first. Please see the commit message in b17370a for more details.

Before and/or soon after merging this, we should give `k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com` "Storage Admin" and "Storage Object Admin" perms for the test-only GCRs:
        - gcr.io/k8s-staging-cip-test
        - gcr.io/k8s-cip-test-prod

Then from the prow config side, we just need to run e2e.go from a container that has gcloud + Docker + Bazel, as it needs all these things to run. There might already be such an image out there in the wild. If not, we could probably add another Bazel rule that can build such an image, or just manually build/upload one using a Dockerfile. But that should be part of another PR.

Anyway, the idea is to have these tests run as presubmits for every change to this repo (kubernetes-sigs/k8s-container-image-promoter), to ensure that the core functionality of the promoter remains intact across refactors, etc.


listx:
/cc @amy @dims

Requesting some more eyes on this if possible.

thockin:
> Before and/or soon after merging this, we should give k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com "Storage Admin" and "Storage Object Admin" perms for the test-only GCRs:
> - gcr.io/k8s-staging-cip-test
> - gcr.io/k8s-cip-test-prod

Why does gcr-promoter need admin access to staging?  It doesn't have admin access to any of the real staging repos?

Why does gcr-promoter need more perms for prod?  It already has Legacy Bucket Owner and Storage Object Admin.

lib/dockerregistry/inventory.go
```diff
@@ -688,6 +701,34 @@ func (sc *SyncContext) ReadRepository(
...
+// over the network.
+func MkReadRepositoryCmdReal(
+	sc *SyncContext,
+	rc RegistryContext) stream.Producer {
+
+	var sh stream.HTTP
+
+	tokenKey, domain, repoPath := GetTokenKeyDomainRepoPath(rc.Name)
+
+	httpReq, err := http.NewRequest(
```
amy: Non-blocker nit. Do we want to add some sort of timeout / retry for this.

lib/dockerregistry/inventory.go
```diff
@@ -1531,6 +1572,152 @@ func (sc *SyncContext) GarbageCollect(
...
+// separately (deletion of manifest lists vs deletion of other media types).
+func (sc *SyncContext) ClearRepository(
+	regName RegistryName,
+	mfest Manifest,
+	mkProducer func(RegistryContext, ImageName, Digest) stream.Producer,
+	customProcessRequest *ProcessRequest) {
+
+	// deleteRequstsPopulator returns a PopulateRequests that
+	// varies by a predicate. Closure city!
+	var deleteRequstsPopulator func(func(cr.MediaType) bool) PopulateRequests = func(predicate func(cr.MediaType) bool) PopulateRequests {
```
amy: Nit. Spelling typo. Requst -> Request. Here and in comment above.

amy: nit. Let's pull out these anonymous functions here and throughout this PR into standalone private functions. Its hard to follow a higher level flow of what the code is generally doing.

lib/dockerregistry/inventory.go
```diff
@@ -1531,6 +1572,152 @@ func (sc *SyncContext) GarbageCollect(
...
+				requestResults <- reqRes
+				// Don't skip over this request, because stderr here is ignorable.
+				// continue
+			}
+			for _, json := range jsons {
+				sc.Info("DELETED image:", json)
+			}
+			reqRes.Errors = errors
+			requestResults <- reqRes
+			wg.Add(-1)
```
amy: use `wg.Done()` here

lib/dockerregistry/inventory.go
```diff
@@ -1531,6 +1572,152 @@ func (sc *SyncContext) GarbageCollect(
...
+		processRequest = processRequestDryRun
+	} else {
+		processRequest = processRequestReal
+	}
+
+	if customProcessRequest != nil {
+		processRequest = *customProcessRequest
+	}
+
+	var isEqualTo (func(cr.MediaType) func(cr.MediaType) bool) = func(want cr.MediaType) func(cr.MediaType) bool {
```
amy: isEquallTo is currently a function that takes a cr.MediaType that returns a function that takes a cr.MediaType that returns a bool. Please also split this out.

Same as below.

listx:
> > Before and/or soon after merging this, we should give [k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com](mailto:k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com) "Storage Admin" and "Storage Object Admin" perms for the test-only GCRs:
> > 
> > * gcr.io/k8s-staging-cip-test
> > * gcr.io/k8s-cip-test-prod
> 
> Why does gcr-promoter need admin access to staging? It doesn't have admin access to any of the real staging repos?

Hmm, good point. I was going to say "let's just reuse the same k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com SA to run the e2e tests", but mixing prod creds with test creds is bad practice. We should create a test-only SA that has the admin creds to the test-only GCRs. Then the Prow job that will run e2e.go will have the creds necessary to carry out the logic inside it.

> 
> Why does gcr-promoter need more perms for prod? It already has Legacy Bucket Owner and Storage Object Admin.

lib/dockerregistry/inventory.go
```diff
@@ -1531,6 +1572,152 @@ func (sc *SyncContext) GarbageCollect(
...
+	var isNotEqualTo (func(cr.MediaType) func(cr.MediaType) bool) = func(want cr.MediaType) func(cr.MediaType) bool {
+		return func(got cr.MediaType) bool {
+			return want != got
+		}
+	}
+
+	// Avoid the GCR error that complains if you try to delete an image which is
+	// referenced by a DockerManifestList, by first deleting all such manifest
+	// lists.
+	deleteManifestLists := deleteRequstsPopulator(isEqualTo(cr.DockerManifestList))
```
amy: It took me a bit to figure out what this line was doing in comparison to the `isEqualTo` anonymous function. Can you simplify `isEqualTo` to take in 2 parameters a & b, compare a& b, then return a bool instead of the nested functions.

lib/dockerregistry/inventory.go
listx: Yes, but if we're going to add it here we should add it everywhere. But even so, I don't think there's a real need to have this be robust over network failures. This is because we can simply re-run the promoter when the network recovers.

lib/dockerregistry/inventory.go
listx: ACK

listx: ACK. Unfortunately I am not sure how to do this gracefully. The main reason we are passing functions around like this (e.g. why we have the PopulateRequests type to begin with) is because for unit tests we do not want to actually go over the real network for some of these functions.

For the case here, we use deleteRequestsPopulator to create network requests for only those manifests that are/are-not Manifest Lists.

lib/dockerregistry/inventory.go
listx: I prefer it as-is because (1) it is more explicit, and (2) because we do some math around the number of these live goroutines (we do `wg.Add(1)` elsewhere).

test/e2e.go
```diff
@@ -0,0 +1,351 @@
...
+}
+
+// E2ETests is an array of E2ETest.
+type E2ETests []E2ETest
+
+// RegistrySnapshot is the snapshot of a registry. It is basically the key/value
+// pair out of the reg.MasterInventory type (RegistryName + []Image).
+type RegistrySnapshot struct {
+	Name   reg.RegistryName `yaml:"name,omitempty"`
+	Before []reg.Image      `yaml:"before,omitempty"`
```
amy: It doesn't seem like this field is used in this PR. Do we want to make sure that the populated registry matches the `Before` before we run the promotion tests. I'd also understand if you think that's overkill.

lib/dockerregistry/inventory.go
listx: While these anonymous functions are independent, their names shouldn't pollute the namespace of this module. This is because they are not used by any other function in this module.

Can we defer refactoring it to later?

lib/dockerregistry/inventory.go
listx: This is not possible AFAICS because we only know `a`, not `b`, at the time we make the call to `deleteRequestsPopulator`. That's why we're passing in a function (I guess you could say it's almost like a callback?) that can deal with `b` when we get it later down the line.

test/e2e.go
listx: Correct, we don't use `Before` in this PR. But thinking about this again, maybe it should. At the very least we would be verifying that the test-registries did get wiped clean (hermetic test environment). I'll work on this a bit.

listx: I just added a check to loop through all expected `Before` snapshots as well. It required a small change to how `getSnapshot` works (need to allocate 0-length array instead of nil array by default, because otherwise `reflect.DeepEqual` will fail), but is otherwise the same.

PTAL

lib/dockerregistry/inventory.go
amy: cool

PR #66, opened 2019-07-09
Add --snapshot-tag flag: only output items with the specified tag

justinsb:
This proved easier for (my) image promotion process, where I want to
extract out the latest images by tag for promotion.


listx:
I would prefer to have it more general, such as: `--snapshot-filter=PATTERN` and have the output be filtered by only those images that match the pattern (so it can be tag or image name or some combination of the two), but I think this is a good start for now.

/lgtm

PR #68, opened 2019-07-09
Simplify code around reading from registries

justinsb:
Much slower as no longer parallel execution, but more direct/simpler.

### DIFF
--- a/cip.go
+++ b/cip.go
@@ -188,7 +188,9 @@ func main() {
 		klog.Fatal(err)
 	}
 
-	sc.ReadRepository(reg.MkReadRepositoryCmdReal)
+	if err := sc.ReadRepository(reg.FetchTags); err != nil {
+		klog.Fatalf("error reading repository: %v", err)
+	}
 
 	if len(*snapshotPtr) > 0 {
 		rii := sc.Inv[mfest.Registries[0].Name]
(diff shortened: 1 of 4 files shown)


justinsb:
So this is an experiment to see what the code looks like if it we do it in the more straightforward way (without the threadpool helpers).  I personally think the code is much easier to follow, but I wanted to raise it as a topic for discussion @listx .

I think we could re-add concurrent requests within the function and retain most of the simplicity, I think, if we thought it critical.

listx:
This is a non-trivial undertaking --- but I'm not sure why you are trying to change code that is known to work. And as you noted yourself, this only worsens the performance. "If it ain't broken, don't fix it", right?

So from those 2 standpoints, I would give this PR 2 strikes. I would give it a 3rd strike because this breaks the established worker pool pattern we use in other functions that go over the network, such as `Promote`, `GarbageCollect`, and `ClearRepository`.

Are there other reasons for this effort other than what you've already stated? Because I would like to know.

listx:
Putting in notes here for the record per our discussion offline @justinsb:

The current design, while somewhat unusual compared to other Go code with the workerpool pattern, is solid. It was designed with 2 things in mind: (1) concurrency and (2) testability (with fakes).

Still, it's worth exploring options to make it easier to understand. Let's examine ways to simplify the codebase but still keep these 2 qualities intact.

PR #77, opened 2019-07-17
clean up (unused) vendored code

listx:
There was a misconception that we needed to vendor codepaths to keep the linter (`lint.sh`) happy. We did not even use this code for building the promoter because the Bazel rules did not reference them; this is why there are no changes to the import paths.


dims:
/lgtm

same as @vincepri please remove hold when you wish this to merge

justinsb:
yay!

/lgtm

Issue #75, opened 2019-07-11
make json decoding boundaries more robust


listx:
We hit a transient error earlier today when trying to promote this manifest:

https://github.com/kubernetes/k8s.io/blob/1d14e8ec918ec2c71c30b6b0a06ee3ea9f724d01/k8s.gcr.io/k8s-staging-cluster-api/manifest.yaml

The error was:

https://storage.googleapis.com/kubernetes-jenkins/logs/post-k8sio-cip/1149377755071123456/build-log.txt

```
MULTIRUN: container image promoter version:
Built:   2019-04-23 03:13:59+00:00
Version: v2.0.1-0-g4db722e
Commit:  4db722e801f62b206f983d0a7adfad0fee99d463
MULTIRUN: gcloud version:
Google Cloud SDK 241.0.0
alpha 2019.04.02
app-engine-go 
app-engine-java 1.9.73
app-engine-python 1.9.85
app-engine-python-extras 1.9.85
beta 2019.04.02
bigtable 
bq 2.0.43
cbt 
cloud-datastore-emulator 2.1.0
core 2019.04.02
datalab 20190116
gsutil 4.38
kubectl 2019.04.02
pubsub-emulator 2019.04.02

MULTIRUN: running against k8s.gcr.io/k8s-staging-cluster-api/manifest.yaml
MULTIRUN: activating service account /etc/k8s-artifacts-prod-service-account/service-account.json
Activated service account credentials for: [k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com]
********** START: k8s.gcr.io/k8s-staging-cluster-api/manifest.yaml **********
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x61750d]

goroutine 10 [running]:
encoding/json.(*Decoder).refill(0xc0004242c0, 0xc000056b01, 0x100c0001aa060)
	GOROOT/src/encoding/json/stream.go:159 +0xcd
encoding/json.(*Decoder).readValue(0xc0004242c0, 0x0, 0x0, 0x6f14e0)
	GOROOT/src/encoding/json/stream.go:134 +0x222
encoding/json.(*Decoder).Decode(0xc0004242c0, 0x68ff80, 0xc000574000, 0xc0005c0020, 0x950268)
	GOROOT/src/encoding/json/stream.go:63 +0x78
[...]

PR #78, opened 2019-07-18
make ReadRepository more robust

listx:
This addresses the concerns in #75. There are 2 key changes:

(1) we do *not* proceed to call `extractRegistryTags` if the network call errored out for some reason. This alone should avoid the panics we have seen as described in #75.

(2) we use retry logic with backoff from k8s.io/apimachinery. This should increase the robustness of ReadRepository when Prow (where CIP runs) gets throttled by GCR.

This isn't a complete fix, because we should invalidate those parts of the registry inventory state that we have trouble populating; these sloppy areas should be removed from the promotion set difference logic entirely. That item is captured as a TODO item, and is the next thing to tackle. I wanted to put this up for review though, to get some feedback before I write too much code.


dims:
/assign @amy 

i took a quick look, the changes looks reasonable to me :)

lib/dockerregistry/inventory.go
```diff
@@ -373,29 +374,57 @@ func (riid *RegInvImageDigest) PrettyValue() string {
...
+		googleTags, err = getRegistryTagsFrom(req)
+
+		// We never return an error (err) in the second part of our return
+		// argument, because we don't want to prematurely stop the
+		// ExponentialBackoff() loop; we want it to continue looping until
+		// either we get a well-formed tags value, or until it hits
+		// ErrWaitTimeout. This is how ExponentialBackoff() uses the
+		// ConditionFunc type.
+		if err == nil && googleTags != nil {
+			return true, nil
```
justinsb: Maybe log the error though?

lib/dockerregistry/inventory.go
```diff
@@ -373,29 +374,57 @@ func (riid *RegInvImageDigest) PrettyValue() string {
...
-		return nil, errors
+		return nil, err
 	}
 
 	tags, err := extractRegistryTags(reader)
 	if err != nil {
-		errors = append(errors, Error{
-			Context: "parsing JSON",
-			Error:   err})
+		klog.Warning("error parsing *google.Tags from io.Reader handle:", err)
```
justinsb: Probably `klog.Warningf("error parsing tags: %v", err)` ; I don't think the types are important enough to be in the error message.

lib/dockerregistry/inventory.go
```diff
@@ -373,29 +374,57 @@ func (riid *RegInvImageDigest) PrettyValue() string {
...
+		return nil, err
+	}
+
+	return googleTags, nil
+}
+
+func getRegistryTagsFrom(req stream.ExternalRequest) (*google.Tags, error) {
 	defer req.StreamProducer.Close()
 
-	errors := make(Errors, 0)
```
justinsb: Nice change - makes the error handling much clearer :+1:

justinsb:
Looks good to me!

/lgtm

listx:
Oops, should probably have put a `/hold` to let Amy chime in as well. I will create a followup PR to address all comments from this PR, so please still take a look if you haven't already.

PR #79, opened 2019-07-21
bazel: allow building from NixOS

listx:
I use NixOS and this patch lets me hack on this project from there.

    bazel: allow building from NixOS

    Literally the only difference that this patch brings is that it will make Bazel call

        go_register_toolchains(go_version = "host")

    instead of

        go_register_toolchains()

    if the platform we are building from is running NixOS (detected with the
    "NIX_PATH" environment variable). We follow the pattern established in
    https://github.com/tweag/rules_purescript/commit/730d38ddb9db63f1e674715b54f813a9a9765b36.
    The original workaround of invoking go_register_toolchains() differently
    was first described in
    https://github.com/bazelbuild/rules_go/issues/1376#issuecomment-375681227.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,4 +1,11 @@
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+git_repository(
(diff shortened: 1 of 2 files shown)


listx:
cf. https://github.com/bazelbuild/rules_go/issues/1376

Issue #64, opened 2019-07-03
make the promoter act recursively for all manifests found under a directory


listx:
Instead of telling the promoter "here is the manifest, here is the SA, please do the promotion" (as we do with util/multirun.sh today), we should be able to say "here is a directory full of manifests and SA configuration info --- please perform promotions for all manifests found there". This is purely a convenience function, but very real and necessary.

Practically speaking, once this feature is implemented we could just add new staging repositories in github.com/kubernetes/k8s.io/k8s.gcr.io with a PR and that would be it --- the existing Prow job would just pick it up automatically (no need to update the Prow job to be invoked with the new staging repo manifest). It's a big win for convenience.

listx:
I think this would be simpler to implement if we made it so that all manifests within a directory all share the same SA creds. This is how it would look like in practice anyway, where we would give `k8s-infra-gcr-promoter@k8s-cip-test-prod.iam.gserviceaccount.com` the rights to read from all staging GCRs (that is, we already want to limit the number of write-access SAs to the prod GCR, so we would naturally just use 1 for it).

There are many possibilities, but one springs to mind: we will only read child manifests' src registry information. Then for each image that has to get promoted, we do a set difference with the dest registries and the src registry that is supposed to have that image. This means that SyncContext.SrcRegistry will no longer be a single registry, but a list (or map --- map would make it easier to do a check for membership without explicit looping). The rest of the business logic should more or less be the same.

There are basically 2 parts to this: (1) implement recursive filesystem tree-walk logic to first read in the set of manifests; (2) implement multi-src repository set difference/promotion logic.

listx:
Putting down some more thoughts here to record the discussion in the wg-k8s-infra meeting today: we can greatly simplify the problem if we ignore multi-manifest consistency and frame this task around just making the logic in `util/multirun.sh` smarter. That is, we can make that logic make the promoter run for *any* manifest (perhaps matching a certain path prefix) that changed across git commits. Today we just limit ourselves to those listed explicitly as an argument to `multirun.sh`, but we could extend that script to make it run dynamically for those manifests found under a directory.

Ideally we would port the logic in multirun.sh into Go, so that we have less shell code.

listx:
Another note: we still have to make the promoter read in *all* manifests recursively for those sanity runs that ensure that the intent is realized for *all* images. The git delta stuff doesn't make sense for this use case.

listx:
[...]

PR #81, opened 2019-07-30
remove -garbage-collect and -delete-extra-tags flags

### LINKED ISSUE #63: guarantee that deletions never happen by splitting off tag deletion logic (*removal* of existing tags) to a separate tool
We can still keep around the tag deletion code, but we shouldn't do it as part of the same `cip` tool. This means retiring the `-delete-extra-tags` and `-garbage-collect` flags from the tool.

listx:
Also make the Manifest type hold more information (this is in preparation for solving #64).

Fixes #63.

### DIFF
--- a/lib/dockerregistry/inventory.go
+++ b/lib/dockerregistry/inventory.go
@@ -575,8 +575,8 @@ func GetTokenKeyDomainRepoPath(
 	return key, s[:i], s[i+1:]
 }
 
-// ReadRepository takes a Repository endpoint, and lists all images at that
-// path. A repository is recursive.
+// ReadAllRegistries reads all images in all registries in the SyncContext Each
+// registry is composed of a image repositories, which can be recursive.
 //
 // To summarize: a docker *registry* is a set of *repositories*. It just so
 // happens that to end-users, repositores resemble a tree structure because they
@@ -603,7 +603,7 @@ func GetTokenKeyDomainRepoPath(
 // NOTE: Repository names may overlap with image names. E.g., it may be in the
 // example above that there are images named gcr.io/google-containers/foo:2.0
 // and gcr.io/google-containers/foo/baz:2.0.
-func (sc *SyncContext) ReadRepository(
+func (sc *SyncContext) ReadAllRegistries(
 	mkProducer func(*SyncContext, RegistryContext) stream.Producer) {
 	// Collect all images in sc.Inv (the src and dest registry names found in
 	// the manifest).
--- a/lib/dockerregistry/inventory.go
+++ b/lib/dockerregistry/inventory.go
@@ -86,7 +86,7 @@ func ParseManifestFromFile(
 	}
 	mfest.filepath = filePath
 
-	mfest, err = ParseManifest(b)
+	mfest, err = ParseManifestYAML(b)
 	if err != nil {
 		return mfest, err
 	}
@@ -107,9 +107,9 @@ func ParseManifestFromFile(
 	return mfest, nil
 }
 
-// ParseManifest parses a Manifest from a byteslice. This function is separate
-// from ParseManifestFromFile() so that it can be tested independently.
-func ParseManifest(b []byte) (Manifest, error) {
+// ParseManifestYAML parses a Manifest from a byteslice. This function is
+// separate from ParseManifestFromFile() so that it can be tested independently.
+func ParseManifestYAML(b []byte) (Manifest, error) {
 	var m Manifest
 	if err := yaml.UnmarshalStrict(b, &m); err != nil {
 		return m, err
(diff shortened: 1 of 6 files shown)


lib/dockerregistry/types.go
```diff
@@ -185,6 +185,13 @@ type Manifest struct {
 	// the promotion itself is performed, do we "denormalize" at the very last
 	// moment by modifying the argument to each destination path.
 	Renames []Rename `yaml:"renames,omitempty"`
+
+	// Hidden fields; these are data structure optimizations that are populated
+	// from the fields above. As they are redundant, there is no point in
+	// storing this information in YAML.
+	RenamesDenormalized RenamesDenormalized `yaml:"-"`
+	SrcRegistry         *RegistryContext    `yaml:"-"`
+	Filepath            string              `yaml:"-"`
```
justinsb: Cool.  You could also make these private & have methods, which would then let you hide the fact that they are cached.  But this is a nice change.

lib/dockerregistry/inventory.go
```diff
@@ -475,7 +475,7 @@ func (sc *SyncContext) IgnoreFromPromotion(regName RegistryName) {
 		klog.Errorf("unable to ignore from promotion: %s\n", err)
 	}
 
-	klog.Errorf("ignoring from promotion: %s\n", err)
+	klog.Infof("ignoring from promotion: %s\n", imgName)
```
justinsb: Do you want to log the error here?  (If the answer is "the errors are going to be obvious by looking at the registry name", that is a fine answer and I agree 100% there is no point logging the error)

lib/dockerregistry/inventory.go
```diff
@@ -475,7 +475,7 @@ func (sc *SyncContext) IgnoreFromPromotion(regName RegistryName) {
 		klog.Errorf("unable to ignore from promotion: %s\n", err)
```
justinsb: Should we return here?

lib/dockerregistry/inventory.go
```diff
@@ -1316,32 +1315,18 @@ func mkPopulateRequestsForPromotion(
...
-						"%s/%s:%s",
-						registry.Name,
-						imageTag.ImageName,
-						imageTag.Tag))
-				}
-				sort.Strings(xtras)
-				for _, img := range xtras {
-					klog.Warningf("Warning: extra tag found in dest: %s\n", img)
-				}
+			// Warn the user about extra tags:
```
justinsb: We probably want to make this optional also ... when I'm copying images around using cip for dev purposes, these are noise to me.

justinsb:
Awesome stuff.  Take a gander at the comments and then feel free to remove hold...

/lgtm
/hold

lib/dockerregistry/inventory.go
```diff
@@ -52,9 +52,7 @@ func getSrcRegistry(rcs []RegistryContext) (*RegistryContext, error) {
 // MakeSyncContext creates a SyncContext.
 func MakeSyncContext(
 	manifestPath string,
```
tpepper: Can this also be removed?

lib/dockerregistry/types.go
listx: Ah. Yes that would be a lot cleaner. I can't think of a good reason why we would want these to be public vs just private as you suggested. Plus the yaml-hiding thing felt ugly and hacky on its own anyway. Ack.

lib/dockerregistry/inventory.go
listx: Actually the error is already logged (in line 475) with Errorf; I just wanted to add an another informational message, which is why I used Infof instead of Errorf (although maybe I should have used Warningf instead).

lib/dockerregistry/inventory.go
listx: Ah! Good eye. Ack.

lib/dockerregistry/inventory.go
listx: Ah ok. I will create an issue for this (make extra tag warning optional) instead of addressing it here because I have some changes queued up which touch this same area.

lib/dockerregistry/inventory.go
listx: Yes, but it will be removed in a forthcoming change (when #64 is implemented and it no longer makes sense to include this in the logs).

lib/dockerregistry/inventory.go
listx: Created https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/82

listx:
Gonna wait this out while waiting for @amy to chime in as well.

PR #85, opened 2019-08-03
Misc fixes

listx:
These are some misc fixes discovered while doing some additional testing while implementing #64.

/hold
/assign @justinsb 
/cc @amy

### DIFF
--- a/lib/stream/http.go
+++ b/lib/stream/http.go
@@ -17,10 +17,13 @@ limitations under the License.
 package stream
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"net/http"
 	"time"
+
+	"k8s.io/klog"
 )
 
 // Checks is from https://stackoverflow.com/a/40398093/437583.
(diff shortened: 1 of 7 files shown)


lib/stream/http.go
```diff
@@ -53,7 +56,25 @@ func (h *HTTP) Produce() (io.Reader, io.Reader, error) {
...
 
-	return h.Res.Body, nil, nil
+	if h.Res.StatusCode == 200 {
+		return h.Res.Body, nil, nil
+	}
+
+	// Try to glean some additional information by reading from the response
+	// body.
+	buf := new(bytes.Buffer)
+	_, err = buf.ReadFrom(h.Res.Body)
```
justinsb: I think you still have to close the Body

justinsb:
/lgtm

One missed close I believe, but can be fixed separately (it's an error case)

lib/stream/http.go
listx: This is done in line 80, but I guess I should just do it here?

listx:
I will address closing the Body in a later PR.

/hold cancel

Issue #71, opened 2019-07-10
add --(re)strict-namespace flag to force images to be promoted within a particular namespace


listx:
E.g., images coming from `gcr.io/foo-project/...` should only be able to be promoted as `k8s.gcr.io/foo-project/...`. That is, the top-level folder name should match `1:1`.

This check would kick in after we do the rename logic (it's still possible to rename images for production); it's just that with the flag turned on we would perform an extra check prior to promotion.

This feature could also be enforced as a field in the manifest YAML, but it's better to keep it as a flag-only thing because we don't want individual promoter manifest owners to change this setting (the binary invocation of the promoter will sit in a tightly-controlled Prow job config).

thockin:
This is a bt more nuanced.  Specifically, images coming from `gcr.io/k8s-staging-foo-project/*` should only promote to `k8s.gcr.io/foo-project/*`.  The difference being that the "project name" in the source is embedded in the name.

Do we need to put that token into the YAML schema for promoter?  E.g.

```
registries:
- name: gcr.io/k8s-staging-foobar
  src: true
- name: us.gcr.io/k8s-artifacts-prod
  service-account: k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com
  subdir: foobar
- name: eu.gcr.io/k8s-artifacts-prod
  service-account: k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com
  subdir: foobar
- name: asia.gcr.io/k8s-artifacts-prod
  service-account: k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com
  subdir: foobar
images: []
renames: []
```

or somethng?

thockin:
Also, we need a special-case for legacy images to have the same name.

listx:
> This is a bt more nuanced. Specifically, images coming from `gcr.io/k8s-staging-foo-project/*` should only promote to `k8s.gcr.io/foo-project/*`. The difference being that the "project name" in the source is embedded in the name.
> 
> Do we need to put that token into the YAML schema for promoter? E.g.
> 
> ```
> registries:
> - name: gcr.io/k8s-staging-foobar
>   src: true
> - name: us.gcr.io/k8s-artifacts-prod
>   service-account: k8s-infra-gcr-promoter@k8s-artifacts-prod.iam.gserviceaccount.com
>   subdir: foobar
[...]

PR #86, opened 2019-08-05
Recursive manifests

listx:
Lots of changes, but the key ones are called "add -manifest-dir flag ...".

The last commit "optimize registry reads" helps greatly reduce the number of API calls against GCS.

/hold
/assign @justinsb


lib/stream/http.go
```diff
@@ -50,6 +50,8 @@ func (h *HTTP) Produce() (io.Reader, io.Reader, error) {
 
 	var err error
 
+	// We close the response body in Close().
+	// nolint[bodyclose]
```
justinsb: Oh!  Disregard my comment on the other PR, and +1 for the explanatory comment :-)

justinsb:
BTW, you don't seem to have a commit called "optimize registry reads"... looking at the remaining changes now, thought it might be easier to rebase ... I'm guessing there were some changes to #85!

lib/dockerregistry/inventory_test.go
```diff
@@ -334,6 +336,342 @@ could not find source registry in '[gcr.io/bar/banana gcr.io/cat/subdir/A/banana
 	}
 }
 
+func TestParseManifestsFromDir(t *testing.T) {
+	// TEST_SRCDIR is the test directory that Bazel creates during test runs.
+	pwd := filepath.Join(
+		os.Getenv("TEST_SRCDIR"),
+		"__main__/lib/dockerregistry/inventory_test/TestParseManifestsFromDir")
```
justinsb: I believe the reliable way to do this is to create a directory `lib/dockerregistry/tests`; gazelle should automatically include it, and then `cwd` during test execution will be `lib/dockerregistry`, so you should be able to use `tests/TestParseManifestsFromDir`...  I think!

listx:
> BTW, you don't seem to have a commit called "optimize registry reads"... looking at the remaining changes now, thought it might be easier to rebase ... I'm guessing there were some changes to #85!

Yes I removed that last commit because I thought you wanted it to be reviewed in its own separate PR. Should I add it back here?

lib/dockerregistry/inventory_test/TestParseManifestsFromDir/basic/a/manifest.yaml
```diff
@@ -0,0 +1,19 @@
...
+  service-account: sa@robot.com
+- name: eu.gcr.io/some-prod
+  service-account: sa@robot.com
+- name: asia.gcr.io/some-prod
+  service-account: sa@robot.com
+images:
+- name: foo-controller
+  dmap:
+    "sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa": ["1.0"]
+renames:
```
justinsb: Renames feels like it isn't worth the complexity?  Does anyone want to use it for anything more than a prefix?

lib/dockerregistry/inventory.go
```diff
@@ -1762,33 +1441,6 @@ func (sc *SyncContext) getPromotionEdgesFromManifests(
...
-	promotionCandidates := mfest.ToRegInvImageDigest().Intersection(src.ToRegInvImageDigest())
-
-	klog.Infof(
-		"To promote (intersection of Manifest and src registry %v):\n%v",
-		sc.SrcRegistry.Name,
-		promotionCandidates.PrettyValue())
-
-	return promotionCandidates.ToRegInvImageTag()
-}
-
```
justinsb: Yay for less code!

cip.go
```diff
@@ -108,38 +112,68 @@ func main() {
...
-	sc, err := reg.MakeSyncContext(
-		mfest.Filepath(),
-		mfest,
-		*verbosityPtr,
-		*threadsPtr,
-		*dryRunPtr,
-		!noSvcAcc)
-	if err != nil {
-		klog.Fatal(err)
+		err = reg.ValidateManifestsFromDir(mfests)
```
justinsb: Would it be easier if ParseManifestsFromDir validates, given that ParseManifestsFromFile does?  (or vice-versa)

cip.go
```diff
@@ -108,38 +112,68 @@ func main() {
...
+	}
+
+	if *manifestPtr != "" {
 		mfest, err = reg.ParseManifestFromFile(*manifestPtr)
 		if err != nil {
 			klog.Fatal(err)
 		}
-	}
+		mfests = append(mfests, mfest)
+		for _, registry := range mfest.Registries {
```
justinsb: Not clear why build sc from mfests is different in the file vs directory case, and also why we don't need to pre-allocate mi in the directory case.

lib/dockerregistry/inventory.go
```diff
@@ -51,59 +54,373 @@ func getSrcRegistry(rcs []RegistryContext) (*RegistryContext, error) {
...
+// ValidateManifestsFromDir parses checks for errors found in the manifests,
+// taken as a whole.
+//
+// nolint[gocyclo]
+func ValidateManifestsFromDir(mfests []Manifest) error {
+	if len(mfests) == 0 {
+		return fmt.Errorf("no manifests to validate")
+	}
+
+	if len(mfests) == 1 {
```
justinsb: I don't think it is worth skipping this validation here - even if it shouldn't catch anything today.

lib/dockerregistry/inventory.go
```diff
@@ -51,59 +54,373 @@ func getSrcRegistry(rcs []RegistryContext) (*RegistryContext, error) {
...
+						}
+
+						// Renames change how edges are created.
+						regImgPath := RegistryImagePath(mfest.srcRegistry.Name) + "/" + RegistryImagePath(image.ImageName)
+						if renames, ok := mfest.renamesDenormalized[regImgPath]; ok {
+							if imgName, ok := renames[rc.Name]; ok {
+								edge.DstImageTag = ImageTag{
+									ImageName: imgName,
+									Tag:       tag}
+								edges[edge] = nil
```
justinsb: Use of `nil` as the key is pretty confusing.  I usually use `map[foo]bool` or if I'm feeling particularly efficient, `map[foo]struct{}`

justinsb:
A few comments, but then lgtm.  Feel free to remove hold when you're happy.

/lgtm
/hold

lib/dockerregistry/inventory_test.go
listx: Noted... I will come back to this when bazel updates and this test breaks. :upside_down_face:

lib/dockerregistry/inventory_test/TestParseManifestsFromDir/basic/a/manifest.yaml
listx: This will probably be simplified with #71 which addresses the common use case around renames.

I agree that renames add a lot of complexity, but I think with the new logic around PromotionEdges the renames are simpler than before.

cip.go
listx: Main reason I split it off like this was for trying to keep the function sizes reasonably small. ParseManifestsFromDir is already quite large. On the other hand I don't want to just rely on ParseManifestFromFile (I think think is what you meant? there is no ParseManifestsFromFile) because 2 independently-valid manifests != 1 valid manifest folder of those same 2 manifests.

cip.go
listx: Oh I suppose they could share the same sc (I think you are asking why we are calling MakeSyncContext twice?)

lib/dockerregistry/inventory.go
listx: Can I fix this in a new PR?

lib/dockerregistry/inventory.go
listx: Wait, `edge` is the key. The value is `nil`. This is using `map[foo]interface{}`, which I suppose is close enough to `map[foo]struct{}`.

listx:
I would like to address the comments in a followup PR.

listx:
Because of the conflicts, I just went ahead and added some changes to simplify. PTAL

lib/dockerregistry/inventory.go
justinsb: Right, seeing you setting the values to nil, I was assuming you were zeroing them out to remove them.  Bad assumption, but it's not the common pattern...

PR #87, opened 2019-08-06
Optimized reads

listx:


lib/dockerregistry/inventory.go
```diff
@@ -1434,19 +1439,71 @@ func mkPopulateRequestsForPromotionEdges(
...
+	mfests []Manifest,
+	readRepos bool,
+) map[PromotionEdge]interface{} {
 
 	edges := ToPromotionEdges(mfests)
+
+	if readRepos {
+		regs := getRegistriesToRead(edges)
+		for _, reg := range regs {
+			fmt.Println("reading this reg:", reg)
```
justinsb: Is this leftover debug code?  klog maybe?

lib/dockerregistry/inventory.go
```diff
@@ -1434,19 +1439,71 @@ func mkPopulateRequestsForPromotionEdges(
...
+			// Do not read these registries recursively, because we already know
+			// exactly which repositories to read (getRegistriesToRead()).
+			false,
+			MkReadRepositoryCmdReal)
+	}
+
 	return sc.getPromotionCandidates(edges)
 }
 
+// getRegistriesToRead collects all unique Docker repositories we want to read
```
justinsb: Should this be called getRepositoriesToRead ?  Admittedly the difference gets me every time.

lib/dockerregistry/inventory.go
```diff
@@ -1434,19 +1439,71 @@ func mkPopulateRequestsForPromotionEdges(
...
+// from. This way, we don't have to read the entire Docker registry, but only
+// those paths that we are thinking of modifying.
+func getRegistriesToRead(
+	edges map[PromotionEdge]interface{}) []RegistryContext {
+
+	rcs := make(map[RegistryContext]interface{})
+
+	// Save the src and dst endpoints as registries. We only care about the
+	// registry and image name, not the tag or digest; this is to collect all
+	// unique Docker repositories that we want to care about.
```
justinsb: Drop "want to" I think

justinsb:
A few comments on the new commit, but it LGTM and I'll re-review after the rebase!

lib/dockerregistry/inventory.go
listx: I feel the same --- but I will do a global s/registry/repository in the future (don't want to tackle that here...).

lib/dockerregistry/inventory.go
listx: Oops, yes, I'll use klog.

PR #88, opened 2019-08-06
file promoter: use json tags

justinsb:
I switched to use sigs.k8s.io/yaml, but that uses json as the tag, not
yaml.  Most fields were compatible, except for the service-account.


listx:
/lgtm
/hold

Remove hold when you're happy.

PR #89, opened 2019-08-06
Set crc32c when uploading to GCS

justinsb:
This acts as an integrity check on the upload.

### DIFF
--- a/pkg/filepromoter/gcs.go
+++ b/pkg/filepromoter/gcs.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
+	"hash/crc32"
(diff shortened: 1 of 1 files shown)


listx:
I guess we are using crc32 because of https://cloud.google.com/storage/docs/hashes-etags which says that it is recommended to use crc32?

/lgtm
/hold

Remove hold when ready.

listx:
/approve

@justinsb I forgot that your PRs don't get auto-approved, because you are not in OWNERS

justinsb:
Right - crc32c is what is used by GCS for the integrity check.  Not sure why they didn't use one of the other ones they are computing as well :-)

/hold cancel

PR #91, opened 2019-08-07
some tweaks to the e2e tests and also bazel updates

listx:
/assign @justinsb


test-e2e/e2e.go
```diff
@@ -87,8 +88,9 @@ func main() {
 			if err != nil {
 				klog.Fatal("could not get pre-promotion snapshot of repo", snapshot.Name)
 			}
-			if !reflect.DeepEqual(snapshot.Before, images) {
-				klog.Fatalf("expected: %s\ngot: %s", snapshot.Before, images)
+			if err := checkEqual(snapshot.Before, images); err != nil {
+				fmt.Println(err)
```
justinsb: It's a test, so it doesn't much matter, but we should probably print to stderr

PR #92, opened 2019-08-12
misc fixes/changes

listx:
I tried to micro-commit as much as possible. PTAL

/assign @justinsb
/hold


lib/dockerregistry/inventory_test.go
```diff
@@ -1315,6 +1315,34 @@ func TestReadRegistries(t *testing.T) {
...
+	}{
+		{
+			"basic",
+			"gcr.io/foo/bar",
+			[3]string{"gcr.io/foo", "gcr.io", "foo/bar"},
+		},
+	}
+
+	for _, test := range tests {
+		tokenKey, domain, repoPath := GetTokenKeyDomainRepoPath(test.input)
```
justinsb: One thing I've found fairly recently is to call `t.Run(test.Name, func (t *testing.T) { ...` and then we don't need to log the name every time.

lib/dockerregistry/inventory.go
```diff
@@ -1033,6 +1022,22 @@ func (sc *SyncContext) ReadRegistries(
 	sc.ExecRequests(populateRequests, processRequest)
 }
 
+func SplitByKnownRegistries(
+	r RegistryName,
+	rcs []RegistryContext) (RegistryName, ImageName, error) {
+
+	for _, rc := range rcs {
+		//		if inameIdx := strings.LastIndex(string(r), string(rc.Name)); inameIdx != -1 {
```
justinsb: Probably should remove this dead code (not necessarily in this PR)

lib/dockerregistry/inventory.go
```diff
@@ -1033,6 +1022,22 @@ func (sc *SyncContext) ReadRegistries(
...
+		//			return rc.Name, ImageName(r[inameIdx+1:]), nil
+		//		}
+		if trimmed := strings.TrimPrefix(string(r), string(rc.Name)); len(trimmed) < len(r) {
+			return rc.Name, ImageName(trimmed[1:]), nil
+		}
+	}
+
+	return "", "", fmt.Errorf("oops #2")
```
justinsb: Maybe a better error?  ("unknown registry %q", r) maybe?

lib/dockerregistry/inventory.go
```diff
@@ -1033,6 +1022,22 @@ func (sc *SyncContext) ReadRegistries(
...
 
+func SplitByKnownRegistries(
+	r RegistryName,
+	rcs []RegistryContext) (RegistryName, ImageName, error) {
+
+	for _, rc := range rcs {
+		//		if inameIdx := strings.LastIndex(string(r), string(rc.Name)); inameIdx != -1 {
+		//			return rc.Name, ImageName(r[inameIdx+1:]), nil
+		//		}
+		if trimmed := strings.TrimPrefix(string(r), string(rc.Name)); len(trimmed) < len(r) {
```
justinsb: I'd probably just do the explicit `strings.HasPrefix` and then a `strings.TrimPrefix`

lib/dockerregistry/inventory.go
```diff
@@ -1033,6 +1022,22 @@ func (sc *SyncContext) ReadRegistries(
...
+			return rc.Name, ImageName(trimmed[1:]), nil
```
justinsb: Is this `[1:]` because we expected it not to have a slash?  If so I think that needs a comment and probably it's safer not to assume that rc.Name is formatted without a slash, and to check that the suffix does start with a slash, etc...

justinsb:
Some comments on the last commit, but otherwise lgtm.

Will apply lgtm & you can remove hold

/lgtm
/hold

lib/dockerregistry/inventory.go
listx: ACK

lib/dockerregistry/inventory.go
listx: So sorry about this; I am ashamed to submit such negligent code. ACK

lib/dockerregistry/inventory.go
listx: ACK

lib/dockerregistry/inventory.go
listx: Yes, exactly. ACK

lib/dockerregistry/inventory_test.go
listx: ACK

PR #95, opened 2019-08-14
Support transparent compression on GCS

justinsb:
We attempt to compress files as part of the upload, and upload a
compressed form if the compression was worthwhile.  We store metadata
in the file attributes, particularly because the md5 is now the md5 of
the compressed file, so need no longer match the source md5.

### DIFF
--- a/pkg/filepromoter/gcs.go
+++ b/pkg/filepromoter/gcs.go
@@ -18,17 +18,23 @@ package filepromoter
 
 import (
 	"context"
-	"encoding/hex"
 	"fmt"
 	"hash/crc32"
 	"io"
 	"os"
+	"strconv"
 	"strings"
 
 	"cloud.google.com/go/storage"
 	"google.golang.org/api/iterator"
 	"k8s.io/klog"
 	api "sigs.k8s.io/k8s-container-image-promoter/pkg/api/files"
+)
+
+const (
+	MetadataKeyUncompressedSize   = "uncompressed-size"
+	MetadataKeyUncompressedSHA256 = "uncompressed-sha256"
+	MetadataKeySHA256             = "sha256"
 )
 
 type gcsSyncFilestore struct {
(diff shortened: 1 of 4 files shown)


pkg/filepromoter/gcs.go
```diff
@@ -46,11 +52,57 @@ func (s *gcsSyncFilestore) OpenReader(
...
+
+		metadata[MetadataKeySHA256] = sha256
+
+		return s.uploadFileRaw(ctx, dest, localFile, "", "", metadata)
+	}
+
+	defer loggedRemove(tmpfile)
+
+	contentEncoding := "gzip"
+	contentType := "application/octet-stream"
```
listx: Ah OK, so we just compress to gzip when we upload it, but on the GCS server it will (of course) de-compress it back. Hence the phrase "transparent compression".

I like to state the obvious to make sure that I understand what's going on... :)

pkg/filepromoter/gzip.go
```diff
@@ -0,0 +1,126 @@
...
+	"os"
+
+	"k8s.io/klog"
+)
+
+// minCompressSize is the minimum size for even trying to compress files
+const minCompressSize = 8 * 1024
+
+// minCompressionRatio is the ratio we must obtain to consider a file compressible
+const minCompressionRatio = 0.9
```
listx: Can you explain the ration in plain English (maybe give an example)? It's not obvious if the 0.9 here means 90% reduction in size or a 10% reduction in size (or something else).

pkg/filepromoter/gzip.go
```diff
@@ -0,0 +1,126 @@
...
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+
+	"k8s.io/klog"
+)
+
+// minCompressSize is the minimum size for even trying to compress files
+const minCompressSize = 8 * 1024
```
listx: Is this in plain bytes, or KB (KiB?), or MB or...?

pkg/filepromoter/gzip.go
```diff
@@ -0,0 +1,126 @@
...
+		klog.V(2).Infof("file %s is too small to compress", localFile)
+		return "", nil
+	}
+
+	tmpfile, err := gzipToTempFile(localFile)
+	if err != nil {
+		return "", fmt.Errorf("error compressing file: %v", err)
+	}
+
+	removeTempfile := true
```
listx: Would it be cleaner to just split this function here (make maybeGzip return `tmpfile` here), and move the tempfile deletion elsewhere?

For example, you could move lines 53-70 to a function that validates the gzipped file (if it fails validation, delete and return empty string; otherwise return tmpfile).

I was just wondering if it was worthwhile to break up this function into smaller pieces.

pkg/filepromoter/gzip.go
```diff
@@ -0,0 +1,126 @@
...
+	defer func() {
+		if removeTempfile {
+			loggedRemove(tmpfile.Name())
+		}
+	}()
+
+	defer loggedClose(tmpfile)
+
+	// TODO: Allow more compression?  Default to more compression?
```
listx: Can you elaborate on this TODO comment? Is it the case that gzip.NewWriter() can be configured with a higher-than-default compression setting, or some such?

PR #97, opened 2019-08-14
Python tool to help building file manifests

justinsb:
And documentation on how to create a manifest.


dev/build-file-manifest
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python
```
listx: Why not just add this logic into the promobot directly under a flag?

Also, what about tests?

listx:
@justinsb ping?

PR #98, opened 2019-08-14
add e2e testing entrypoint (to be run from within a container in Prow)

listx:
/hold
/assign @justinsb


listx:
The sister change from the Prow job side (turning "on" the e2e tests) will be: https://github.com/listx/test-infra/commit/f41d7e2da0ff6dabf9c0fabd33a6bf9e67b37209

PR #100, opened 2019-08-15
cip: add service account activation logic

listx:
This is in preparation for the deprecation of util/multirun.sh (which currently performs the account activation for us).

/assign @justinsb


justinsb:
:100: :tada: 

/lgtm

PR #106, opened 2019-08-20
SplitByKnownRegistries: sort names for determinism

listx:
Otherwise, the order of looping in SplitByKnownRegistries is
    non-deterministic and in the case where there are 2 known registries
    with a shared leading prefix string, we might select the wrong one.

    E.g., if we have "gcr.io/foo" and "gcr.io/foo-bar", we might end up
    splitting the string "gcr.io/foo-bar/qux" with the first one, giving us
    "gcr.io/foo" and "-bar/qx" (incorrect). This patch prevents this
    scenario.

/assign @justinsb @ps882

### DIFF
--- a/lib/dockerregistry/inventory.go
+++ b/lib/dockerregistry/inventory.go
@@ -85,6 +85,11 @@ func MakeSyncContext(
 		sc.RegistryContexts = append(sc.RegistryContexts, r)
 	}
 
+	// Sort the list for determinism.
+	sort.Slice(sc.RegistryContexts, func(i, j int) bool {
+		return sc.RegistryContexts[i].Name < sc.RegistryContexts[j].Name
+	})
+
 	// Populate access tokens for all registries listed in the manifest.
 	err := sc.PopulateTokens()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


lib/dockerregistry/inventory.go
```diff
@@ -1028,6 +1046,10 @@ func SplitByKnownRegistries(
 	r RegistryName,
 	rcs []RegistryContext) (RegistryName, ImageName, error) {
 
+	sort.Slice(rcs, func(i, j int) bool {
+		return len(rcs[i].Name) > len(rcs[j].Name)
```
justinsb: Deliberately sorting in the other order?

justinsb:
/approve
/lgtm
/hold

Hold because of query around the other order.  Maybe you're just checking if I'm actually reading these :-)

Feel free to unhold if deliberate.

lib/dockerregistry/inventory.go
listx: Yes. I should have added a comment to explain (to echo my commit message). Will add a comment

listx: Done

listx:
Fixed typo in commit message. Please re-lgtm!

PR #108, opened 2019-08-21
lint: download go dependencies

listx:
In a CI setting, none of the go dependencies will be there on disk, so
we have to populate it first.


listx:
/hold

Testing this out to fix https://prow.k8s.io/?job=pull-cip-lint

listx:
The linter job passes now: https://prow.k8s.io/log?job=pull-cip-lint&id=1163982920172441600

PR #115, opened 2019-08-27
WIP: use go-containerregistry for promoting new images

listx:
This way we open up the path to promoting images without needing to tag them.

/hold

This should work in theory; let's see what the e2e test says...


listx:
Closing this in favor of #118 because the e2e test for that just passed.

/close

Issue #121, opened 2019-09-05
Build fails on macOS : date: illegal option -- -


glyn:
With bazel 0.29, `make build` fails on master (93aef5cdde14c6c68ff6456a7117d4b4c6e018f7) on macOS with:
```
[...]
go: finding github.com/cpuguy83/go-md2man v1.0.4
go: error loading module requirements
DEBUG: /private/var/tmp/_bazel_gnormington/8ca8190ddc76e058023fc72217a03eec/external/bazel_gazelle/internal/go_repository.bzl:147:13: fetch_repo: go: finding golang.org/x/net v0.0.0-20190620200207-3b0461eec859
DEBUG: /private/var/tmp/_bazel_gnormington/8ca8190ddc76e058023fc72217a03eec/external/bazel_gazelle/internal/go_repository.bzl:147:13: fetch_repo: go: finding cloud.google.com/go v0.44.0
INFO: Analyzed target //:cip (51 packages loaded, 6762 targets configured).
INFO: Found 1 target...
ERROR: Process exited with status 1
date: illegal option -- -
usage: date [-jnRu] [-d dst] [-r seconds] [-t west] [-v[+|-]val[ymwdHMS]] ...
            [-f fmt date | [[[mm]dd]HH]MM[[cc]yy][.ss]] [+format]
Target //:cip failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 260.607s, Critical Path: 0.11s
INFO: 0 processes.
FAILED: Build did NOT complete successfully
make: *** [build] Error 1
```

PR #122, opened 2019-09-05
Improve README

glyn:
The README doesn't look quite right. I propose to use this PR to discuss and tune it.

/cc @listx

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,11 +105,11 @@ If there are multiple destination registries, the above calculation is repeated
 for each destination registry. The promoter also prints warnings about images in
 `M` that cannot be found in the source registry (call it `S`):
 
-- `D \ S` = images that are lost (no way to promote it because it cannot be found!)
+- `M \ S` = images that are lost (no way to promote it because it cannot be found!)
 
 ## Server-side operations
 
-During the promtion process, all data resides on the server (currently, Google
+During the promotion process, all data resides on the server (currently, Google
 Container Registry for images). That is, no images get pulled and pushed back
 up. There are two reasons why it does things entirely server-side:
 


README.md
```diff
@@ -105,11 +105,11 @@ If there are multiple destination registries, the above calculation is repeated
 for each destination registry. The promoter also prints warnings about images in
 `M` that cannot be found in the source registry (call it `S`):
```
glyn: Does "it" refer to the images in the source registry?

glyn:
I was surprised by the definition of `D` as all images in the destination _registry_. There are typically lots of images that shouldn't be relevant.

Also, I couldn't find the flag `delete-extra-tags` in the code (and I couldn't build the code to try the online help: https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/121).

README.md
listx: > Does "it" refer to the images in the source registry?

Yes.

At the same time, the logic around using Set difference for calculating what needs to be copied over was dropped in https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/86/commits/b062e7a33898604917394986036d26c15c476562, so the language in the README using these concepts should be removed.

listx:
The README as it is today needs more work. You caught something that has been in the back burner of my mind for a while now.

> I was surprised by the definition of `D` as all images in the destination _registry_. There are typically lots of images that shouldn't be relevant.

Correct. There are 2 things to note here: (1) technically, gcr.io is a registry, and gcr.io/foo is a repository and we only really care about repositories, not registries; (2) the set-difference logic to calculate what needs to be promoted/deleted was removed, and so we no longer care about *all* images in a registry (or repository). I vote for simply removing the explanation via set theory.

> 
> Also, I couldn't find the flag `delete-extra-tags` in the code (and I couldn't build the code to try the online help: #121).

This flag was removed on purpose in https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/81

glyn:
@listx Thanks for the feedback! The set-based _implementation_ may have gone, but wouldn't a set-theoretic description of the _what_ the implementation does (rather than _how_ it does it) still be worth keeping? I would hope so...

listx:
> @listx Thanks for the feedback! The set-based _implementation_ may have gone, but wouldn't a set-theoretic description of the _what_ the implementation does (rather than _how_ it does it) still be worth keeping? I would hope so...

Hmm, I suppose so.

Meanwhile can you sign the CLA agreement? That's a blocker for this to get merged.

glyn:
> > @listx Thanks for the feedback! The set-based _implementation_ may have gone, but wouldn't a set-theoretic description of the _what_ the implementation does (rather than _how_ it does it) still be worth keeping? I would hope so...
> 
> Hmm, I suppose so.

Let's work that out together in this PR.

> 
> Meanwhile can you sign the CLA agreement? That's a blocker for this to get merged.

I was told that my name had been added to Pivotal's corporate CLA, so I'll follow up to see what else needs doing.

glyn:
Not sure why the CLA check is failing. When I click "details" and try to sign as an employee, it looks ok:
Screenshot 2019-09-16 at 08 49 03

glyn:
@listx The CLA is sorted now. Please could you merge or let me have further comments.

dims:
/approve

leaving `/lgtm` for @listx

listx:
My Github-foo is weak and I don't understand why this PR is called a "draft" PR and cannot be merged. Help?

Katharine:
This is causing tide issues.

/hold


@listx a draft PR is a type of PR that is natively flagged as a draft by github and cannot be merged. The original creator has to remove the flag via the UI.

glyn:
I made this a draft PR because I was expecting more iteration, but if it's good to go, let's merge it.

/lgtm

PR #128, opened 2019-09-13
bazel: update rules_docker

listx:


listx:
Hmm, I cannot reproduce this error. I'm going to assume that it has to do with a different bazel version (0.29.1) vs the one in Prow.

As this isn't really a blocker for anything, I'll close this PR for now.

PR #129, opened 2019-09-17
Tighten up set description

glyn:


README.md
```diff
@@ -87,25 +87,21 @@ from running `make build`, and then invoking it with the correct path under
...
-the destination registry (let's call it `D`). It also creates the set of all
-images defined in the promoter manifest as "promotion candidates in the
-manifest" (call it `M`).
+The promoter's behaviour can be described in terms of mathematical sets (as in Venn diagrams).
+Suppose `D` is the set of all images in the destination registry and `M` is the set of 
+images to be promoted (these are defined in the promoter manifest). Then:
 
-Given the above, we get the following results:
+- `M ∩ D` = images which do not need promoting since they are already present in the destination registry
+- `M \ D` = images that are copied
```
glyn: Or, if we take into account those which cannot be found (in which case the definition of `S` needs moving too):
```suggestion
- `(M ∩ S) \ D` = images that are copied
```

README.md
```diff
@@ -87,25 +87,21 @@ from running `make build`, and then invoking it with the correct path under
...
+The above statements are true for each destination registry.
 
-If there are multiple destination registries, the above calculation is repeated
-for each destination registry. The promoter also prints warnings about images in
-`M` that cannot be found in the source registry (call it `S`):
+The promoter also prints warnings about images that cannot be promoted.
+Suppose `S` is the set of images in the source registry. Then:
 
-- `M \ S` = images that are lost (no way to promote it because it cannot be found!)
+- `M \ S` = images that cannot be found
```
glyn: @listx are warning issues for images that are already present in the destination registry? If not then:
```suggestion
- `M \ (S ∪ D)` = images that cannot be found
```

listx: We do print [Infof messages](https://github.com/kubernetes-sigs/k8s-container-image-promoter/blob/master/lib/dockerregistry/inventory.go#L360) for images that are already present in the destination registry (it's not a warning because it is expected behavior after the image is promoted (on the 2nd and all subsequent runs)).

Relevant: https://github.com/kubernetes-sigs/k8s-container-image-promoter/blob/master/lib/dockerregistry/inventory.go#L358 we ignore images that are already found in the destination, regardless of whether it exists in the source or not.

glyn:
@listx Thanks for your comments. The PR is now ready to be reviewed/merged.

PR #118, opened 2019-08-28
WIP: Tagless promotions

listx:
If the e2e test passes then this PR will supersede #115.

/hold

Issue #130, opened 2019-09-18
add support for promoting V2 Schema 1 (2.1) images


listx:
This is because the way in which we use the new go-containerregistry library for doing the image copies only supports Version 2 Schema 2 images.

There are several choices:

1. wait for go-containerregistry to expose its `copySchema1Image` function that `gcrane` internally uses
2. shell out to `gcrane` (which is the frontend CLI for go-containerregistry)
3. vendor all Go dependencies (including go-containerregiistry) and manually lift/import the code in `copySchema1Image` which references a private function called `legacy.CopySchema1`

I would strongly prefer the first option.

jonjohnsonjr:
Use https://godoc.org/github.com/google/go-containerregistry/pkg/crane#Copy

PR #131, opened 2019-09-20
use crane.Copy for copying images

listx:
This function should have been used in #118 in
522e885080643b991cde38bfe65abfa6efc4b5e5.

/assign @justinsb


listx:
I should say that this addresses #130 , because go-containerregistry already supports Version 2, Schema 1 image copies.

tpepper:
nice simplification especially with github.com/google/go-containerregistry already an active dep
/lgtm

PR #132, opened 2019-09-20
Manifest validation

listx:
This makes the -manifest-dir flag more flexible and robust. There are 2 things in this PR:

(1) allow multiple manifest files to reference the *same* source registry (e.g., legacy manifests to backfill subprojects from gcr.io/google-containers)
(2) make sure that as a whole, the manifests do not step on each other's toes; this means that we have to check whether the image destinations do *not* overlap if the digests are different. IOW, if different images are being promoted (this is essentially always the case for -manifest-dir flag), then make sure that they do not try to promote to the same destination name + tag combination. If the digests are the same, then it's actually OK because it's a form of "redundant" promotion (different source registries with the *same* image trying to promote to the *same* destination).

/assign @justinsb

### DIFF
--- a/lib/dockerregistry/inventory.go
+++ b/lib/dockerregistry/inventory.go
@@ -218,17 +218,6 @@ func ValidateManifestsFromDir(mfests []Manifest) error {
 		return fmt.Errorf("no manifests to validate")
 	}
 
-	// Check that there are no overlapping manifests. Each manifest must be
-	// responsible for 1 unique source registry.
-	srcRegsSeen := make(map[RegistryName]string)
-	for _, mfest := range mfests {
-		if mfestFilepath, seen := srcRegsSeen[mfest.srcRegistry.Name]; seen {
-			// nolint[lll]
-			return fmt.Errorf("source registry '%s' defined in multiple manifests:\n- '%s'\n- '%s'\n", mfest.srcRegistry.Name, mfestFilepath, mfest.filepath)
-		}
-		srcRegsSeen[mfest.srcRegistry.Name] = mfest.filepath
-	}
-
 	// If two manifests are renaming images, then they should not share any
 	// rename information (should be mutually exclusive). We use a separate loop
 	// for clarity.
--- a/lib/dockerregistry/inventory.go
+++ b/lib/dockerregistry/inventory.go
@@ -239,7 +239,8 @@ func ValidateManifestsFromDir(mfests []Manifest) error {
 
 // ToPromotionEdges converts a list of manifests to a set of edges we want to
 // try promoting.
-func ToPromotionEdges(mfests []Manifest) map[PromotionEdge]interface{} {
+func ToPromotionEdges(
+	mfests []Manifest) (map[PromotionEdge]interface{}, error) {
 	edges := make(map[PromotionEdge]interface{})
 	// nolint[lll]
 	for _, mfest := range mfests {
@@ -278,7 +279,8 @@ func ToPromotionEdges(mfests []Manifest) map[PromotionEdge]interface{} {
 			}
 		}
 	}
-	return edges
+
+	return checkOverlappingEdges(edges)
 }
 
 func mkPromotionEdge(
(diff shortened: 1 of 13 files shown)


listx:
Oops, let me fix the failing presubmits first.

lib/dockerregistry/inventory.go
```diff
@@ -404,6 +395,92 @@ func (sc *SyncContext) getPromotionCandidates(
...
+	promotionIntent := make(map[string]map[Digest][]PromotionEdge)
+	for edge := range edges {
+		dstPQIN := ToPQIN(edge.DstRegistry.Name,
+			edge.DstImageTag.ImageName,
+			edge.DstImageTag.Tag)
+
+		digestToEdges, ok := promotionIntent[dstPQIN]
+		if ok {
+			// Store the edge in the appropriate bucket.
+			edgeList, ok := digestToEdges[edge.Digest]
```
justinsb: So I think you can do

`digestToEdges[edge.Digest] = append(digestToEdges[edge.Digest], edge)`

https://play.golang.org/p/IZRWwWY7FDG

lib/dockerregistry/inventory.go
```diff
@@ -404,6 +395,92 @@ func (sc *SyncContext) getPromotionCandidates(
...
+
+	// Review the promotionIntent to ensure that there are no issues.
+	overlapError := false
+	checked := make(map[PromotionEdge]interface{})
+	for pqin, digestToEdges := range promotionIntent {
+		if len(digestToEdges) < 2 {
+			for _, edgeList := range digestToEdges {
+				switch len(edgeList) {
+				case 0:
+					klog.Errorf("no edges for %v\n", pqin)
```
justinsb: Fatalf?  Something's gone seriously wrong here, right?

(Also you don't need the trailing newline)

justinsb:
Suggestion to simplify code, but LGTM

/lgtm
/hold

lib/dockerregistry/inventory.go
listx: So if I do

```
		digestToEdges, ok := promotionIntent[dstPQIN]
		if ok {
			// Store the edge.
			digestToEdges[edge.Digest] = append(digestToEdges[edge.Digest], edge)
			promotionIntent[dstPQIN] = digestToEdges
		} else {
			// Make this edge lay claim to this destination vertex.
			edgeList := make([]PromotionEdge, 0)
			edgeList = append(edgeList, edge)
			digestToEdges := make(map[Digest][]PromotionEdge)
			digestToEdges[edge.Digest] = edgeList
			promotionIntent[dstPQIN] = digestToEdges
		}
```

it compiles and the unit tests pass. But if I make it slightly more DRY with

```
		digestToEdges, ok := promotionIntent[dstPQIN]
		if ok {
			// Store the edge.
			digestToEdges[edge.Digest] = append(digestToEdges[edge.Digest], edge)
		} else {
			// Make this edge lay claim to this destination vertex.
			edgeList := make([]PromotionEdge, 0)
			edgeList = append(edgeList, edge)
			digestToEdges := make(map[Digest][]PromotionEdge)
			digestToEdges[edge.Digest] = edgeList
		}
		promotionIntent[dstPQIN] = digestToEdges
```

I get this unit test failure:

```
--- FAIL: TestValidateManifestsFromDir (0.01s)
panic: assignment to entry in nil map [recovered]
	panic: assignment to entry in nil map

goroutine 10 [running]:
testing.tRunner.func1(0xc000101200)
	GOROOT/src/testing/testing.go:874 +0x3a3
panic(0x8a0ac0, 0x9975f0)
	GOROOT/src/runtime/panic.go:679 +0x1b2
sigs.k8s.io/k8s-container-image-promoter/lib/dockerregistry.checkOverlappingEdges(0xc0001ef390, 0xc0001ef390, 0xc0001ef978, 0xc0001efad0)
	lib/dockerregistry/inventory.go:425 +0x226
sigs.k8s.io/k8s-container-image-promoter/lib/dockerregistry.ToPromotionEdges(0xc0003eb930, 0x2, 0x2, 0xc0000285a0, 0x60, 0xc0000285a0)
	lib/dockerregistry/inventory.go:283 +0x137
sigs.k8s.io/k8s-container-image-promoter/lib/dockerregistry.TestValidateManifestsFromDir(0xc000101200)
	lib/dockerregistry/inventory_test.go:677 +0x3d5
testing.tRunner(0xc000101200, 0x934600)
	GOROOT/src/testing/testing.go:909 +0xc9
created by testing.(*T).Run
	GOROOT/src/testing/testing.go:960 +0x350
```

I find this surprising. I guess I can't use the 2nd (DRY-er) form...?

lib/dockerregistry/inventory.go
listx: Oops, I should set `overlapError = true` here. Yes something's gone seriously wrong, but I want to collect as much error as possible so I would like to avoid Fatalf here.

Hmm I don't get why I added the trailing newlines (looks like I forgot this is not C again). Ack.

listx: Hmm actually it should be a different type of error, because it's not like an overlap was actually detected. Will change.

PR #134, opened 2019-09-24
 allow using the promoter for diffing GCR contents

listx:
The new flags are `-manifest-based-snapshot-of` and `-minimal-snapshot`.

`-manifest-based-snapshot-of=<registry>`: This is only usable with either a `-manifest` or `-manifest-dir` flag. This is like `-snapshot`, but instead of asking the registry for images information, it only looks at the image information in `-manifest`/`-manifest-dir` (the promoter manifest(s)), filtered by `<registry>`. The use case is to generate a sort of "global" manifest given a set of smaller manifests (case in point: the new staging repos). The YAML output here can be diffed to create a delta with the output of `-snapshot` for auditing purposes.

`-minimal-snapshot`: This can be paired with `-manifest-based-snapshot-of` or -snapshot`; it removes all images in the output YAML that are children of DockerManifestLists. So if a manifest list "X" refers to A, B, and C (and A/B/C are all untagged to purposely discourage architecture-specific images), using this flag removes A/B/C (if they are tagless) from the output of the snapshotting flags.

/assign @justinsb

### DIFF
--- a/cip.go
+++ b/cip.go
@@ -93,6 +93,10 @@ func main() {
 		"snapshot-service-account",
 		"",
 		"service account to use for -snapshot")
+	manifestBasedSnapshotOf := flag.String(
+		"manifest-based-snapshot-of",
+		"",
+		"read all images in either -manifest or -manifest-dir and print all images that will be promoted to the given registry; this is like -snapshot, but instead of reading from a registry, it reads from the manifests those images that need to be promoted to the given registry")
 	noSvcAcc := false
 	flag.BoolVar(&noSvcAcc, "no-service-account", false,
 		"do not pass '--account=...' to all gcloud calls (default: false)")
@@ -149,6 +153,15 @@ func main() {
 		}
 	}
 
+	if len(*manifestBasedSnapshotOf) > 0 {
+		srcRegistry = &reg.RegistryContext{
+			Name:           reg.RegistryName(*manifestBasedSnapshotOf),
+			ServiceAccount: *snapshotSvcAccPtr,
+			Src:            true,
+		}
+	}
+
+	doingPromotion := false
 	if *manifestPtr != "" {
 		mfest, err = reg.ParseManifestFromFile(*manifestPtr)
 		if err != nil {
@@ -167,6 +180,7 @@ func main() {
 		if err != nil {
 			klog.Fatal(err)
 		}
+		doingPromotion = true
 	} else if *manifestDirPtr != "" {
 		mfests, err = reg.ParseManifestsFromDir(*manifestDirPtr)
 		if err != nil {
@@ -187,6 +201,7 @@ func main() {
 		if err != nil {
 			klog.Fatal(err)
 		}
+		doingPromotion = true
 	}
 
 	if *parseOnlyPtr {
@@ -196,7 +211,7 @@ func main() {
 	// If there are no images in the manifest, it may be a stub manifest file
 	// (such as for brand new registries that would be watched by the promoter
 	// for the very first time).
-	if len(*snapshotPtr) == 0 {
+	if doingPromotion && len(*manifestBasedSnapshotOf) == 0 {
 		promotionEdges, err = reg.ToPromotionEdges(mfests)
 		if err != nil {
 			klog.Exitln(err)
@@ -224,26 +239,48 @@ func main() {
 		}
 	}
 
-	if len(*snapshotPtr) > 0 {
-		sc, err = reg.MakeSyncContext(
-			mfests,
-			*verbosityPtr,
-			*threadsPtr,
-			*dryRunPtr,
-			!noSvcAcc)
-		if err != nil {
-			klog.Fatal(err)
-		}
-		sc.ReadRegistries(
-			[]reg.RegistryContext{*srcRegistry},
-			// Read all registries recursively, because we want to produce a
-			// complete snapshot.
-			true,
-			reg.MkReadRepositoryCmdReal)
+	if len(*snapshotPtr) > 0 || len(*manifestBasedSnapshotOf) > 0 {
+		rii := make(reg.RegInvImage)
+		if len(*manifestBasedSnapshotOf) > 0 {
+			promotionEdges, err = reg.ToPromotionEdges(mfests)
+			if err != nil {
+				klog.Exitln(err)
+			}
+			rii = reg.EdgesToRegInvImage(promotionEdges,
+				*manifestBasedSnapshotOf)
 
-		rii := sc.Inv[mfests[0].Registries[0].Name]
-		if snapshotTag != "" {
-			rii = reg.FilterByTag(rii, snapshotTag)
+			if *minimalSnapshotPtr {
+				sc.ReadRegistries(
+					[]reg.RegistryContext{*srcRegistry},
+					// No need to read all registries recursively because if we
+					// read in manifests, by default we treat each image entry
+					// as its own repository.
+					true,
+					reg.MkReadRepositoryCmdReal)
+				sc.ReadGManifestLists(reg.MkReadManifestListCmdReal)
+				rii = sc.RemoveChildDigestEntries(rii)
+			}
+		} else {
+			sc, err = reg.MakeSyncContext(
+				mfests,
+				*verbosityPtr,
+				*threadsPtr,
+				*dryRunPtr,
+				!noSvcAcc)
+			if err != nil {
+				klog.Fatal(err)
+			}
+			sc.ReadRegistries(
+				[]reg.RegistryContext{*srcRegistry},
+				// Read all registries recursively, because we want to produce a
+				// complete snapshot.
+				true,
+				reg.MkReadRepositoryCmdReal)
+
+			rii = sc.Inv[mfests[0].Registries[0].Name]
+			if snapshotTag != "" {
+				rii = reg.FilterByTag(rii, snapshotTag)
+			}
 		}
 
 		snapshot := rii.ToYAML()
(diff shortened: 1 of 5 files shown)


cip.go
```diff
@@ -89,10 +89,18 @@ func main() {
 		"read all images in a repository and print to stdout")
 	snapshotTag := ""
 	flag.StringVar(&snapshotTag, "snapshot-tag", snapshotTag, "only snapshot images with the given tag")
+	minimalSnapshotPtr := flag.Bool(
+		"minimal-snapshot",
```
justinsb: I was thinking about "better" names, but I actually like that this is a statement of intent rather than just being narrowly focused on manifest lists.  :+1:

cip.go
```diff
@@ -149,6 +157,15 @@ func main() {
 		}
 	}
 
+	if len(*manifestBasedSnapshotOf) > 0 {
+		srcRegistry = &reg.RegistryContext{
+			Name:           reg.RegistryName(*manifestBasedSnapshotOf),
+			ServiceAccount: *snapshotSvcAccPtr,
+			Src:            true,
+		}
+	}
```
justinsb: It's surprising (at first glance) that we don't set mfests here, but we do in the `*snapshotPtr` case.  Deliberate?

cip.go
```diff
@@ -196,7 +215,7 @@ func main() {
 	// If there are no images in the manifest, it may be a stub manifest file
 	// (such as for brand new registries that would be watched by the promoter
 	// for the very first time).
-	if len(*snapshotPtr) == 0 {
+	if doingPromotion && len(*manifestBasedSnapshotOf) == 0 {
```
justinsb: Is `doingPromotion` is actually doing anything here?  This looks like the only place you read the var, and it looks like it must be true when we get here anyway

lib/dockerregistry/inventory.go
```diff
@@ -872,6 +880,63 @@ func getRegistryTagsFrom(req stream.ExternalRequest) (*google.Tags, error) {
...
+
+		// We never return an error (err) in the second part of our return
+		// argument, because we don't want to prematurely stop the
+		// ExponentialBackoff() loop; we want it to continue looping until
+		// either we get a well-formed value, or until it hits
+		// ErrWaitTimeout. This is how ExponentialBackoff() uses the
+		// ConditionFunc type.
+		if err == nil &&
+			gManifestList != nil &&
+			len(gManifestList.Manifests) > 0 {
```
justinsb: Aside: is it possible to have a manifest list with no manifests?  Maybe this is what you're hitting on the odd real-world cases.

lib/dockerregistry/inventory.go
```diff
@@ -872,6 +880,63 @@ func getRegistryTagsFrom(req stream.ExternalRequest) (*google.Tags, error) {
...
+
+			return true, nil
+		}
+		return false, nil
```
justinsb: If you get an error, I think you should log it, even if only at `klog.V(2)`

lib/dockerregistry/inventory.go
```diff
@@ -872,6 +880,63 @@ func getRegistryTagsFrom(req stream.ExternalRequest) (*google.Tags, error) {
...
+		return nil, err
+	}
+
+	return gManifestList, nil
+}
+
+func getGManifestListFrom(req stream.ExternalRequest) (*GManifestList, error) {
+	reader, _, err := req.StreamProducer.Produce()
+	if err != nil {
+		klog.Warning("error reading from stream:", err)
```
justinsb: Ah - you're logging it here.  It's typically better to return the error and let the caller choose to log or not, IMO

lib/dockerregistry/inventory.go
```diff
@@ -1171,6 +1236,143 @@ func (sc *SyncContext) ReadRegistries(
 	sc.ExecRequests(populateRequests, processRequest)
 }
 
+// ReadGManifestLists reads all manifest lists and populates the ParentDigest
```
justinsb: _If_ we could afford to give up parallel execution, this function would be much easier to understand!

lib/dockerregistry/inventory.go
```diff
@@ -1171,6 +1236,143 @@ func (sc *SyncContext) ReadRegistries(
...
+
+// RemoveChildDigestEntries removes all tagless images in RegInvImage that are
+// referenced by ManifestLists in the Registries.
+func (sc *SyncContext) RemoveChildDigestEntries(rii RegInvImage) RegInvImage {
+	filtered := make(RegInvImage)
+	for imageName, digestTags := range rii {
+		for digest, tagSlice := range digestTags {
+			_, hasParent := sc.ParentDigest[digest]
+			// If this digest is referenced by a parent ManfestList, skip over
+			// it (do not add to "filtered") only if it is also tagless.
```
justinsb: You could simplify the language here e.g. "If this image digest is only referenced as part of a parent ManifestList (i.e. not directly tagged), we filter it out"

lib/dockerregistry/inventory.go
```diff
@@ -1229,6 +1431,53 @@ func MkReadRepositoryCmdReal(
...
+		gmlc.Tag)
+
+	httpReq, err := http.NewRequest("GET", endpoint, nil)
+
+	// Without this, GCR responds as we had used the "Accept:
+	// application/vnd.docker.distribution.manifest.v1+prettyjws" header.
+	httpReq.Header.Add("Accept", "*/*")
+
+	if err != nil {
+		klog.Fatalf(
```
justinsb: Idea for the future: Maybe tackling stream.Producer is a reasonable place to start in terms of making the code simpler - can we just return a `([]byte, error)`

lib/dockerregistry/inventory.go
```diff
@@ -1305,6 +1554,25 @@ func extractRegistryTags(reader io.Reader) (*google.Tags, error) {
...
+	decoder := json.NewDecoder(reader)
+	decoder.DisallowUnknownFields()
+
+	for {
+		err := decoder.Decode(&gManifestList)
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			fmt.Println("DECODING ERROR:", err)
```
justinsb: Nit: use klog instead of stdout

lib/dockerregistry/inventory.go
```diff
@@ -1561,6 +1829,57 @@ func (sc *SyncContext) FilterPromotionEdges(
...
+// EdgesToRegInvImage takes the destination endpoints of all edges and converts
+// their information to a RegInvImage type. It uses only those edges that are
+// trying to promote to the given destination registry.
+func EdgesToRegInvImage(
+	edges map[PromotionEdge]interface{},
+	destRegistry string) RegInvImage {
+
+	rii := make(RegInvImage)
+
+	destRegistry = strings.TrimRight(destRegistry, "/")
```
justinsb: Don't you want to consistently add the `/`, so `gcr.io/foo` doesn't match `gcr.io/foobar` ?

lib/dockerregistry/types.go
```diff
@@ -250,6 +251,15 @@ type RegistryContext struct {
 	Src            bool         `yaml:"src,omitempty"`
 }
 
+// GManifestListContext is used only for reading GManifestList information from
```
justinsb: Nit: GCR might be a better prefix than G

justinsb:
A few nits / queries, but overall LGTM

cip.go
listx: Oops, this is an oversight. Ack

cip.go
listx: Hmm it's only true if we pass in `-manifest` or `-manifest-dir`.

lib/dockerregistry/inventory.go
listx: Good point. I wrote down your thoughts in my debugging notebook!

lib/dockerregistry/inventory.go
listx: Hmm I do log errors encountered by the child function `getGManifestListFrom()`... Still, I could log oddities such as not having encountered an error but finding that for example the child manifests list is empty. Ack

lib/dockerregistry/inventory.go
listx: lol Github hid your comment here so I didn't see it until l unfolded this convo... Anyway ya I will change the logging convention to make the caller log it (it feels cleaner that way to me actually, dunno why I didn't do it that way the first time)

listx: Oh wait, I just remembered. It's because of the way ExponentialBackoff behaves. If I return an error here, then the backoff loop will exit (on instead of letting the backoff mechanism exit the backoff loop due to timeout). See https://godoc.org/k8s.io/apimachinery/pkg/util/wait#ConditionFunc

That explains the non-idiomatic code... still, I can add logging to make the various different error cases more explicit.

lib/dockerregistry/inventory.go
listx: Yes... I can think of a way to simplify it but let's hold off on that until later (will add a TODO for myself here). Ack.

lib/dockerregistry/inventory.go
listx: Ack.

lib/dockerregistry/inventory.go
listx: Ack.

lib/dockerregistry/inventory.go
listx: Ack.

lib/dockerregistry/inventory.go
listx: This is actually sanitizing the `*manifestBasedSnapshotOf` pointer value as we pass that on directly to this function from cip.go.

lib/dockerregistry/types.go
listx: Ack

PR #139, opened 2019-10-10
remove renames feature

listx:
The ability to rename images (not just the image names but the entire
path altogether) was never really used in practice. The only time that
has happened is when images path prefixes were renamed, as in

https://github.com/kubernetes/k8s.io/commit/1d14e8ec918ec2c71c30b6b0a06ee3ea9f724d01#diff-6d388aca90f2a61817328dbc3db1a123

but that change was short-lived as it was superseded by

https://github.com/kubernetes/k8s.io/commit/946c4d9d1e34f6054af7ac75e84075facfdeb801.

/assign @justinsb


justinsb:
Thanks @listx - agree that this is simpler & safer

/approve
/lgtm

PR #141, opened 2019-10-11
add -thin-manifest-dir flag

### LINKED ISSUE #138: promoter manifest: move "registries" field out to a separate YAML location
This will improve the security posture because it would separate credentials from the image details.

listx:
Thin manifests are like regular manifests but instead of having `images: ...` defined directly, they have a `imagesPath: ...` field that links to an images YAML that resides in a separate file.

I could have added a corresponding `-thin-manifest` flag, but decided not to because this is the minimal change necessary for the actual use case in k8s.io repo.

Fixes #138

/hold
/assign @justinsb


listx:
I plan to add an e2e test for this flag later todayl; let's hold off until we can get that working as well.

lib/dockerregistry/inventory.go
```diff
@@ -163,17 +225,12 @@ func ParseManifestsFromDir(dir string) ([]Manifest, error) {
...
 
 		// First try to parse the path as a manifest file. The only requirement
-		// is that the file must end with a ".yaml" extension. We can be more
-		// restrictive in the future (maybe it will be all files with a certain
-		// pattern, passable from the CLI as an option), but let's cross that
-		// bridge when we get there.
-
-		// Skip non-YAML files.
-		if !strings.HasSuffix(path, ".yaml") {
+		// is that the file must be named "promoter-manifest.yaml".
```
justinsb: Very Henry Ford of you to describe this as the "only" restriction ;-)

lib/dockerregistry/inventory.go
```diff
@@ -163,17 +225,12 @@ func ParseManifestsFromDir(dir string) ([]Manifest, error) {
...
+		if !strings.HasSuffix(path, "/promoter-manifest.yaml") {
```
justinsb: Checking info.Name() would avoid problems on Windows or if we didn't have a parent dir

justinsb:
LGTM - two very minor nits that you can fix separately if you agree / think it's worth it.

/approve
/lgtm

listx:
Thanks Justin, will address those nits in a new PR!

/hold cancel

PR #153, opened 2019-11-17
replace '-no-service-account' flag with '-use-service-account'

### LINKED ISSUE #150: reverse behavior of -no-service-account flag
This way, service accounts are not attached by default to the request in case someone forgets to attach `-no-service-account` flag.

listx:
Also, do not activate service accounts (by default).

Fixes #150.


listx:
This fixes #150

dims:
/lgtm

please cancel hold

Issue #151, opened 2019-11-14
ParseThinManifestFromFile: limit path traversal


listx:
Make the traversal path (1) relative-only and (2) limit its scope to only be within the `-thin-manifest-dir` directory.

bartsmykla:
@listx I'm not sure if `ParseManifestFromFile` also should be limited, so I'll create PR with limiting both of them (it sounds to me as more reasonable), but if this one shouldn't be limited I'll change the PR.

listx:
Closing this in favor of https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/161, which is more comprehensive (and stricter!). Please take a look @bartsmykla

PR #156, opened 2019-12-05
Limit path traversal when parsing mfest from file

bartsmykla:
Closes: https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/151

imagesPath from thin manifests is now limited to only
--thin-manifest-dir directory

Signed-off-by: Bart Smykla <bartek@smykla.com>

### DIFF
--- a/lib/dockerregistry/inventory.go
+++ b/lib/dockerregistry/inventory.go
@@ -107,7 +107,7 @@ func MakeSyncContext(
 }
 
 // ParseManifestFromFile parses a Manifest from a filepath.
-func ParseManifestFromFile(filePath string) (Manifest, error) {
+func ParseManifestFromFile(filePath string, _ string) (Manifest, error) {
 
 	var mfest Manifest
 	var empty Manifest
@@ -134,7 +134,9 @@ func ParseManifestFromFile(filePath string) (Manifest, error) {
 
 // ParseThinManifestFromFile parses a ThinManifest from a filepath and generates
 // a Manifest.
-func ParseThinManifestFromFile(filePath string) (Manifest, error) {
+func ParseThinManifestFromFile(
+	filePath string,
+	thinManifestDir string) (Manifest, error) {
 
 	var thinManifest ThinManifest
 	var mfest Manifest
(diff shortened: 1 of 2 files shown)


lib/dockerregistry/inventory.go
```diff
@@ -106,12 +106,23 @@ func MakeSyncContext(
 	return sc, nil
 }
 
+func isPathRelative(rootPath string, relativePath string) error {
```
listx: Can you just make this return a bool?

lib/dockerregistry/inventory.go
```diff
@@ -106,12 +106,23 @@ func MakeSyncContext(
...
+	_, err := filepath.Rel(rootPath, relativePath)
```
listx: The docs for `Rel` https://golang.org/pkg/path/filepath/#Rel have an example, and in one case `Rel("/a", "/b/c")` returns `../b/c` (no error, even though the two paths are outside of each other).

lib/dockerregistry/inventory.go
```diff
@@ -208,9 +223,15 @@ func (m *Manifest) finalize() error {
 // (there can only be 1 source registry).
 func ParseManifestsFromDir(
```
listx: https://golang.org/pkg/path/filepath/#Walk already only walks recursively inside the given directory and does not follow symlinks, so changing this function is unnecessary.

bartsmykla:
I have updated the PR to limit the imagesPath to the `--thin-manifest-dir` directory as initially stated in the issue, because it would be bigger update if we want to hardcode the path like this: `../../images/<name-of-staging-subproject>/images.yaml` because we don't have access to the `name-of-staging-subproject` directly in thin manifests, and it would be a little hacky (I think) to parse the array of images and then the string with item which has property `src` inside (I can see some potential edge cases here). @listx if you still think it wold be better to hardcode it I can change the PR. :-)

cc. @mkumatag

listx:
> I have updated the PR to limit the imagesPath to the `--thin-manifest-dir` directory as initially stated in the issue, because it would be bigger update if we want to hardcode the path like this: `../../images/<name-of-staging-subproject>/images.yaml` because we don't have access to the `name-of-staging-subproject` directly in thin manifests, and it would be a little hacky (I think) to parse the array of images and then the string with item which has property `src` inside (I can see some potential edge cases here). @listx if you still think it wold be better to hardcode it I can change the PR. :-)
> 
> cc. @mkumatag

I think it's better to hardcode, mainly because we already follow the strict naming convention in the promoter manifests we have today. Let's start with strict rules first where we can, and only allow power/flexibility when we need them later.

bartsmykla:
@listx where I should take the `<name-of-staging-subproject>` from?

listx:
> @listx where I should take the `<name-of-staging-subproject>` from?

It is in the name of the subfolder, e.g. all the ones here: https://github.com/kubernetes/k8s.io/tree/37cdf1dd6f527f9d6215c4941957eaf2b0e58a30/k8s.gcr.io/manifests

listx:
Closing this in favor of #161, which is more comprehensive (and stricter!). Please take a look @bartsmykla

PR #158, opened 2019-12-11
disallow tag moves (immutable tags)

### LINKED ISSUE #157: add a utility to check tag moves at the manifest level (independent of GCR state)
Given two versions of promoter manifests, ensure that tags are never moved. That is, going from manifest 1 to 2, there should only be *new* tags. All other cases, such as tag deletions (which can happen for legitimate reasons) or moving tags (such as swapping tags maliciously as in https://github.com/kubernetes/k8s.io/pull/499), should result in an error.

This check (let's call it delta checker as it will check the changes between two manifests) will be run as another presubmit check for changes to promoter manifests. If it fails, some humans must manually approve the change (maybe force push or something manually).

listx:
Because tags should always be immutable, moving tags are no longer
supported. We leave the code around detecting such events in tact, in
case we want to revert this behavior in the future.

Note that this means if the destination registries get into a bad state
because of a system independent of the promoter, the promoter can no
longer be used to rectify the situation.

Fixes #157

### DIFF
--- a/cip.go
+++ b/cip.go
@@ -329,7 +329,12 @@ func main() {
 			tp)
 		return &sp
 	}
-	promotionEdges = sc.FilterPromotionEdges(promotionEdges, true)
+	promotionEdges, ok := sc.FilterPromotionEdges(promotionEdges, true)
+	// If any funny business was detected during a comparison of the manifests
+	// with the state of the registries, then exit immediately.
+	if !ok {
+		klog.Exitln(err)
+	}
 	err = sc.Promote(promotionEdges, mkProducer, nil)
 	if err != nil {
 		klog.Exitln(err)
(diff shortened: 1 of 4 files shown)


listx:
/hold

Holding this for any discussions that may arise.

thockin:
Can you expand on what you mean by the bad situation?

listx:
> Can you expand on what you mean by the bad situation?

If an attacker mutates images (no image deletions, just additions and/or moves to the tags), fixing the situation will require N tag moves (back to the original image contents that are still in GCR). The promoter can no longer be used for this "fixup" case, because all tag moves are disallowed with this PR.

thockin:
Ah, so a human would have to fix it.  BUt a human would have caused it, so
not egregious, IMO

listx:
> Ah, so a human would have to fix it. BUt a human would have caused it, so not egregious, IMO
> […](#)

Yup, exactly. And for the record, the fix would be to restore from one of our backups.

PR #159, opened 2019-12-11
add auditing mechanism implementation

listx:
The meat of it is in the commit `add basic auditing mechanism`. This just extends the cip binary so that it can be run inside Cloud Run as an HTTP server.

In the spirit of time I am just putting this up now so that it can gather some review comments. I still need to create a test framework to test changes to this auditing mechanism. That test will involve creating images in a fake prod GCR, then mutating those images, then catching the Pub/Sub messages of those mutation events with the auditor, and making sure that we get certain logs (with the idea that particular log events will be monitored as a log-based metric in GCP stackdriver).

/hold for any discussions

/cc @justinsb @thockin @bartsmykla


listx:
This is part of https://github.com/kubernetes/k8s.io/issues/394

listx:
I've tested this with Error Reporting and it works with local testing. Here's what I've done (all with my own GCP project for testing purposes):

- started a Cloud Run application of the promoter, using the new `-audit` flag
- pushed a new image in a GCR that was being audited that was *not* accounted for in promoter manifests (this created an error on [Stackdriver Error Reporting](https://cloud.google.com/error-reporting/))
- pushed a new image that *was* accounted for in the manifests (this GCR "transaction" was logged as verified)
- deleted an image (this created an error as well, because we treat all deletions as errors)

To summarize this new feature, the "auditor" that this feature introduces works as follows:

- cloning a Git repo that holds promoter manifests
- accepts HTTP reqs that contain a particular "GCR transaction payload" (the HTTP request is created by Cloud Run, which in turn transforms GCR Pub/Sub messages into these HTTP reqs)
- upon each request, the auditor compares it with the Git repo holding the promoter manifests (it updates the Git repo on each request)
- if it passes muster, nothing happens (we log "TRANSACTION VERIFIED...")
- if it does not pass muster, we report it to Stackdriver Error Reporting, and a new error gets created and an email gets sent to the GCP project owner

I am not completely sure if the email that gets sent to the GCP project owner is configurable, but that is the observed behavior I have seen.

We can go back to the original idea of creating log-based metrics and alerting on the rise of those metrics, but I find the Error Reporting UI easier to grok for our use case.

PTAL

cip.go
```diff
@@ -108,10 +109,30 @@ func main() {
 	manifestBasedSnapshotOf := flag.String(
 		"manifest-based-snapshot-of",
 		"",
-		"read all images in either -manifest or -manifest-dir and print all images that will be promoted to the given registry; this is like -snapshot, but instead of reading from a registry, it reads from the manifests those images that need to be promoted to the given registry")
+		"read all images in either -manifest or -manifest-dir and print all images that should be promoted to the given registry (assuming the given registry is empty); this is like -snapshot, but instead of reading over the network from a registry, it reads from the local manifests only")
 	useServiceAccount := false
 	flag.BoolVar(&useServiceAccount, "use-service-account", false,
 		"pass '--account=...' to all gcloud calls (default: false)")
+	auditorPtr := flag.Bool(
```
thockin: Without knowing this code very well - I'd expect the audit-mode to be roughly the same as for non-audit.  Trigger, read manifests, do work (alert if audit, else promote).

I guess non-audit mode reals files directly and audit mode reads the URL - is that why they are different?  Would it make sense (medium term?) to converge those?

lib/audit/auditor.go
```diff
@@ -0,0 +1,362 @@
...
+	branch string,
+) (*git.Repository, string, error) {
+	tdir, err := ioutil.TempDir("", "k8s.io-")
+	if err != nil {
+		return nil, "", err
+	}
+
+	r, err := git.PlainClone(tdir, false, &git.CloneOptions{
+		URL:           repoURL.String(),
+		ReferenceName: (plumbing.ReferenceName)("refs/heads/" + branch),
```
thockin: is there a depth option?  Save clone time...

lib/audit/auditor.go
```diff
@@ -0,0 +1,362 @@
...
+	if err == nil {
+		klog.Infof("updated %v to revision %v", tdir, sha)
+	}
+
+	return r, tdir, nil
+}
+
+// Check if the given path already has a git repository. If it does, just use
+// that. Otherwise, clone a repo to a new temporary path.
+func cloneOrPull(repoURL fmt.Stringer,
```
thockin: I'm worried aboutthe sorts of corner-cases git-sync experiences here.  I wonder if it would be simpler and more successful to just do a clone with --depth=1 on every alert?

cip.go
listx: The auditing code shares some code with non-audit code, but the behavior is different:

- non-audit code: read manifests on disk, perform GCR promotions (image copies) using go-containerregistry API
- audit code: read manifests by URL, read HTTP req with GCR state change, check whether state change is pre-approved by the manifests, alert (Stackdriver Error Reporting API) if change is not verifiable

The convergence of the "read manifests" bit at the start would be tricky, mainly because of the way the non-audit code works in harmony with Github PRs (which could have manifests on different branches). I don't think it's worth the effort to converge them.

lib/audit/auditor.go
listx: Ah yes, indeed. ACK

lib/audit/auditor.go
listx: It would be simpler. Re-reading this code, it looks like my treatment of ErrNonFastForwardUpdate is wrong (I should always fetch and then reset-hard to the latest branch HEAD, because if I do hit a ErrNonFastForwardUpdate error the library will not actually update the branch HEAD).

I will just clone every time and take care to delete the tmp directory after I'm done reading the manifests. ACK

listx:
I will create new PRs for the code review fixups.

/hold cancel

Regarding notifications, looks like Stackdriver Error Reporting (which is what this PR is using) page has some info about Slack integration: https://cloud.google.com/error-reporting/docs/notifications

There is also a Cloud Console mobile app...!!

PR #161, opened 2020-01-02
Strict path traversal

### LINKED ISSUE #151: ParseThinManifestFromFile: limit path traversal

listx:
This fixes #151. Notably, the `-manifest-dir` flag is now deprecated (no one is using it anyway, at least not in k8s/test-infra).

The directory holding the manifests must look like this:

```
	//       foo
	//       ├── images
	//       │   ├── a
	//       │   │   └── images.yaml
	//       │   ├── b
	//       │   │   └── images.yaml
	//       │   ├── c
	//       │   │   └── images.yaml
	//       │   └── d
	//       │       └── images.yaml
	//       └── manifests
	//           ├── a
	//           │   └── promoter-manifest.yaml
	//           ├── b
	//           │   └── promoter-manifest.yaml
	//           ├── c
	//           │   └── promoter-manifest.yaml
	//           └── d
	//               └── promoter-manifest.yaml
```

Anything (that I could think of) that deviates from this is rejected.

/cc @justinsb @thockin @bartsmykla


cip.go
```diff
@@ -193,7 +189,7 @@ func main() {
 			},
 		}
 	} else {
-		if *manifestPtr == "" && *manifestDirPtr == "" && *thinManifestDirPtr == "" {
+		if *manifestPtr == "" && *thinManifestDirPtr == "" {
 			klog.Fatal(fmt.Errorf("one of -manifest, -manifest-dir, or -thin-manifest-dir is required"))
```
justinsb: Nit: technically this is still true, but one of those is no longer an option :-)

lib/dockerregistry/inventory.go
```diff
@@ -150,13 +151,12 @@ func ParseThinManifestFromFile(filePath string) (Manifest, error) {
...
-	// Handle a relative path for "ImagesPath", if necessary.
-	if strings.HasPrefix(thinManifest.ImagesPath, ".") {
-		imagesPath = filepath.Join(
-			filepath.Dir(filePath),
-			thinManifest.ImagesPath)
-	}
+	// Get directory name holding this thin manifest.
+	subProject := filepath.Base(filepath.Dir(filePath))
+	imagesPath := filepath.Join(filepath.Dir(filePath),
+		"../../images",
```
justinsb: Nesting images under the manifest _might_ be less confusing, and _might_ still satisfy the OWNERs rules; but if this is what we're doing today then we should stick to it, I guess!

lib/dockerregistry/inventory.go
```diff
@@ -232,7 +241,22 @@ func ParseManifestsFromDir(
...
+		// inside a subfolder within "manifests/<dir>" --- any other paths are
+		// forbidden.
+		shortened := strings.TrimPrefix(path, dir)
+		shortenedList := strings.Split(shortened, "/")
+		if len(shortenedList) != ThinManifestDepth {
+			return fmt.Errorf("unexpected manifest path %q",
+				path)
+		}
+
+		if !reflect.DeepEqual(shortenedList[0:2], []string{"", "manifests"}) {
```
justinsb: Nit: This might be overkill for reflect!  Also, if we made sure that `path` always had a `/` at the end would we avoid the first "" ?

lib/dockerregistry/inventory.go
```diff
@@ -255,6 +279,76 @@ func ParseManifestsFromDir(
...
+func ValidateThinManifestDirectoryStructure(
+	dir string,
+) error {
+	// First, enforce that there are directories named "images" and "manifests".
+	dirExists, err := directoryExists(filepath.Join(dir, "images"))
+	if !dirExists || err != nil {
+		return err
+	}
+
+	mfestDir := filepath.Join(dir, "manifests")
```
justinsb: Nit: clearer just to spell it out - `manifestsDir` or `manifestsPath`

lib/dockerregistry/inventory.go
```diff
@@ -255,6 +279,76 @@ func ParseManifestsFromDir(
...
+// named "foo/manifests/bar/promoter-manifest.yaml" exists, that a corresponding
+// file named "foo/images/bar/promoter-manifest.yaml" must also exist.
+//
+// nolint[gocyclo]
+func ValidateThinManifestDirectoryStructure(
+	dir string,
+) error {
+	// First, enforce that there are directories named "images" and "manifests".
+	dirExists, err := directoryExists(filepath.Join(dir, "images"))
+	if !dirExists || err != nil {
```
justinsb: Nit: You can combine these onto one line, to avoid shadowing issues.

lib/dockerregistry/inventory.go
```diff
@@ -255,6 +279,76 @@ func ParseManifestsFromDir(
...
+		return err
```
justinsb: Blocker (?): If !dirExists but err==nil, we might return no error?  Though it is an error I believe.

lib/dockerregistry/inventory.go
```diff
@@ -255,6 +279,76 @@ func ParseManifestsFromDir(
...
+		}
+
+		// Only check for directory files.
+		if p.IsDir() {
+			// Check if "promoter-manifest.yaml" exists here.
+			if _, err := os.Stat(
+				filepath.Join(mfestDir,
+					file.Name(),
+					"promoter-manifest.yaml")); os.IsNotExist(err) {
+				continue
```
justinsb: Blocker (?): This doesn't match the comment - the comment says that promoter-manifest.yaml must exist, but this tolerates it not existing.  It looks like it actually tolerates _any_ error.

lib/dockerregistry/inventory.go
```diff
@@ -255,6 +279,76 @@ func ParseManifestsFromDir(
...
+			}
+
+			// Check for corresponding images file.
+			imagesPath := filepath.Join(dir,
+				"images",
+				file.Name(), // wrong; include all parent dirs up to dir
+				"images.yaml")
+			if _, err := os.Stat(imagesPath); os.IsNotExist(err) {
+				return fmt.Errorf("corresponding file %q does not exist",
```
justinsb: We should also return error on other errors, I believe.

lib/dockerregistry/inventory.go
```diff
@@ -255,6 +279,76 @@ func ParseManifestsFromDir(
...
+	return nil
+}
+
+func directoryExists(dir string) (bool, error) {
+	p, err := os.Stat(filepath.Join(dir))
+	if err != nil {
+		return false, err
+	}
+	if !p.IsDir() {
+		return false, fmt.Errorf("%q does not exist", dir)
```
justinsb: Might save some head-scratching to say "%q is not a directory"

lib/dockerregistry/inventory.go
```diff
@@ -255,6 +279,76 @@ func ParseManifestsFromDir(
...
+				return fmt.Errorf("corresponding file %q does not exist",
+					imagesPath)
+			}
+		}
+	}
+
+	return nil
+}
+
+func directoryExists(dir string) (bool, error) {
```
justinsb: Suggest renaming this to `validateIsDirectory(string) error` to avoid the contract that `false => err != nil` that we hit above.

justinsb:
Generally LGTM, though I thought there were a few blockers around strictness of validation.

lib/dockerregistry/inventory.go
listx: My thoughts exactly. I can't find a strong enough reason to change it. Maybe that change can come in the future if we deem it necessary.

listx:
@justinsb PTAL; I've added all changes to a new commit to make changes easier to see. (I assume this is preferred on GitHb over the more traditional squashing of changes piecemeal back into the originating commits where they would have belonged).

justinsb:
Thanks for the changes - lgtm!

/approve
/lgtm

Failure looks like a timeout

/retest

PR #162, opened 2020-01-07
workspace_status.sh: do not rely solely on Git

listx:
In some environments (such as is cloudbuild.yaml which Prow uses), the
source code does not contain the .git folder and instead we are provided
with the _GIT_TAG variable. Use this variable instead of relying on
Git (which would otherwise result in a build failure).

This should fix https://github.com/kubernetes/kubernetes/issues/86880

/cc @dims @justinsb @thockin


dims:
/approve
/lgtm

let's try this!

Issue #165, opened 2020-01-08
auditor: make sure that it can handle information about child images in manifest list images


listx:
Problem:

- user adds fat manifest digest (without children) to promoter manifests,
- user pushes fat manifest list to GCR
  - Docker pushes one child image at a time, until all are pushed
  - then the fat manifest is created at the end
- the child image push events will come in from Pub/Sub, but the auditor will only knows about parent images, so the child image push events will not be verified

We could either:

1. force users to always add child image information in promoter manifests (ugly workaround),
2. make the auditor read GCR state to populate child image information when it first boots up,
3. make the auditor read GCR state for child image's subproject

The 1st solution is just a workaround and is actually not what users want (users only want to track the *minimal* valid information, which is the fat manifest list digest, not the child images).

The 2nd solution is not ideal, because reading GCR state will take N seconds (could depend on network throttling also) and during this time we might miss out on reading child images for certain images that were pushed *during* the read.

I think the 3rd solution is what we should go with. For more context, all production images should reside under a prefix named after the subproject they are touching. This means that if we get a Pub/Sub message about a child image for subproject FOO, we can just read everything under the FOO prefix (something like `gcr.io/k8s-artifacts-prod/FOO/...`). IIRC we already have logic in the `-minimal-snapshot` flag that removes child images of fat manifest lists. So we could (1) read the subproject prefix in prod, (2) get the minimal snapshot, (3) check which images were discarded from (1) and (2), and finally check if the child image digest is in the set of discarded images from (3).

listx:
#170 addressed this

PR #170, opened 2020-01-18
Auditor e2e test

listx:
add e2e test for auditing mechanism
    
    This essentially involves running the auditor in Cloud Run, mutating
    state in GCR, and seeing if we get logs from the auditor we launched
    saying whether it verified or rejected the change.

Once https://github.com/kubernetes/test-infra/pull/15924 lands, we can just iterate on this PR if necessary (probably to iron out any permission kinks with `k8s-infra-gcr-promoter@k8s-gcr-audit-test-prod.iam.gserviceaccount.com`).

We can of course add more tests (currently this only performs 2 mutations to GCR state --- adding and deleting a basic image) in the future.

/cc @thockin @justinsb @dims @bartsmykla 

Holding for any necessary discussions.
/hold


listx:
I've added a test case for fat manifests (#165).

listx:
I am working with the test-infra team to try to use Workload Identity for https://github.com/kubernetes/test-infra/pull/15924. That would make the need to pass in the service account key unnecessary, so that might have to change in the future but that would only require a small tweak.

/hold cancel

PR #172, opened 2020-01-22
fix perms for pull-cip-auditor-e2e test

listx:
This is just a test to invoke the new pull-cip-auditor-e2e test (it
should pass --- if it doesn't GCP permissions are messed up somewhere).

/hold

### DIFF
--- a/test-e2e/cip-auditor/cip-auditor-e2e.go
+++ b/test-e2e/cip-auditor/cip-auditor-e2e.go
@@ -133,6 +133,9 @@ func runE2ETests(testsFile, repoRoot string) {
 	if err := enableStackdriverAPI(projectID); err != nil {
 		klog.Fatal("error enabling Stackdriver API", err)
 	}
+	if err := enableStackdriverErrorReportingAPI(projectID); err != nil {
+		klog.Fatal("error enabling Stackdriver Error Reporting API", err)
+	}
 
 	// Allow Pub/Sub to create auth tokens for the project.
 	if err := enablePubSubTokenCreation(projectNumber, projectID); err != nil {
@@ -815,6 +818,16 @@ func enableStackdriverAPI(
 	args := getCmdEnableService(
 		projectID,
 		"stackdriver.googleapis.com")
+	_, _, err := execCommand("", args...)
+	return err
+}
+
+func enableStackdriverErrorReportingAPI(
+	projectID string,
+) error {
+	args := getCmdEnableService(
+		projectID,
+		"clouderrorreporting.googleapis.com")
 	_, _, err := execCommand("", args...)
 	return err
 }
--- a/test-e2e/cip-auditor/cip-auditor-e2e.go
+++ b/test-e2e/cip-auditor/cip-auditor-e2e.go
@@ -136,6 +136,9 @@ func runE2ETests(testsFile, repoRoot string) {
 	if err := enableStackdriverErrorReportingAPI(projectID); err != nil {
 		klog.Fatal("error enabling Stackdriver Error Reporting API", err)
 	}
+	if err := enableCloudRunAPI(projectID); err != nil {
+		klog.Fatal("error enabling Cloud Run API", err)
+	}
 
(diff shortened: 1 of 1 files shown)


listx:
The pull-cip-auditor-e2e test is now passing!
For the record, I had to do 2 things manually because I felt that putting them in the auditor test code was not the best thing:

(1) create a Pub/Sub topic named `gcr` (this is used by GCR's Pub/Sub integration)
(2) make the "gcr.io" registry host Public instead of Private (for some reason, `{asia,eu,us}.gcr.io/k8s-gcr-audit-test-prod` all had Public visibility but just `gcr.io/k8s-gcr-audit-test-prod` was Private, leading to errors when the promoter was trying to read from it).

We would need to do the above things in the k8s-artifacts-prod project before launching the cip-auditor to Cloud Run. 

/cc @thockin @justinsb @dims 
/hold

listx:
@bartsmykla FYI I created https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/174 to add you to OWNERS.

test-e2e/cip-auditor/cip-auditor-e2e.go
```diff
@@ -647,19 +703,22 @@ func getCmdsDeployCloudRun(
 			),
 			"--platform=managed",
 			"--no-allow-unauthenticated",
+			"--region=us-central1",
```
justinsb: I expected we're going to parameterize this region eventually, but fine for now

test-e2e/cip-auditor/cip-auditor-e2e.go
```diff
@@ -554,21 +600,24 @@ func getCmdEmpowerServiceAccount(invokerServiceAccount string) []string {
...
-func getCmdPurgePubSubMessages() []string {
+func getCmdPurgePubSubMessages(projectID string) []string {
 	return []string{
 		"gcloud",
 		"pubsub",
 		"subscriptions",
 		"seek",
 		subscriptionName,
 		"--time=+p1y",
+		fmt.Sprintf("--project=%s", projectID),
```
justinsb: Might be worth thinking about introducing a class to help with this, but ... can be done later.

test-e2e/cip-auditor/cip-auditor-e2e.go
```diff
@@ -755,26 +820,77 @@ func enablePubSubTokenCreation(
...
-func empowerServiceAccount(invokerServiceAccount string) error {
-	args := getCmdEmpowerServiceAccount(invokerServiceAccount)
+func enableCloudResourceManagerAPI(
+	projectID string,
+) error {
+	args := getCmdEnableService(
+		projectID,
+		"cloudresourcemanager.googleapis.com")
+	_, _, err := execCommand("", args...)
+	return err
```
justinsb: This could also be refactored to call a shared method `func enableService(projectID, service string)` ... or a class.

test-e2e/cip-auditor/cip-auditor-e2e.go
```diff
@@ -755,26 +820,77 @@ func enablePubSubTokenCreation(
...
+	args := getCmdEmpowerServiceAccount(projectID, invokerServiceAccount)
 	_, _, err := execCommand("", args...)
 	return err
 }
 
-func createPubSubSubscription(invokerServiceAccount string) error {
-	args := getCmdCloudRunPushEndpoint()
+func clearPubSubMessages(projectID string) error {
+	args := getCmdPurgePubSubMessages(projectID)
+	_, _, err := execCommand("", args...)
```
justinsb: Another way to refactor: have `getCmdPurgePubSubMessages` call exec.

justinsb:
This could probably be simplified with a refactor, but ... not in this PR

/approve
/lgtm

listx:
@justinsb I agree that the code can be cleaned up, certainly. However I would prefer to do that in a later PR.

BTW I think you mean structs/methods instead of classes?

/hold cancel

PR #175, opened 2020-02-11
WIP: Files: add a command to upload files while also computing the manifest

justinsb:
This should be useful in building a staging workflow.

### DIFF
--- a/cmd/promobot-stage-and-hash-files/README.md
+++ b/cmd/promobot-stage-and-hash-files/README.md
@@ -1,0 +1,41 @@
+# File Promoter
+
+The file promoter (aka "promobot-files") copies files from source GCS buckets to
+one or more destination buckets, by reading a Manifest file (in YAML).
+
+The Manifest lists the files and their hashes that should be copied from src to
+dest.
+
+Example Manifest for files:
+
+```
+filestores:
+- base: gs://staging/
+  src: true
+- base: gs://prod/subdir
+  service-account: foo@google-containers.iam.gserviceaccount.com
+files:
+- name: vegetables/artichoke
+  sha256: 2d4f26491e0e470236f73a0b8d6828db017eab988cd102fc19afe31f1f56aff7
+- name: vegetables/beetroot
+  sha256: 160b98e27ec99f77efe01e2996fa386f2b2aec552599f8bd861be0a857e7f29f
+```
+
+`filestores` is the equivalent of the container manifest `registries`, and lists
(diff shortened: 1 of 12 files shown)


cmd/promobot-stage-and-hash-files/README.md
```diff
@@ -0,0 +1,41 @@
...
+the buckets from which the promoter should read or write files.  `files` is the
+equivalent of `images`, and lists the files that should be promoted.
+
+`filestores` supports `service-account`, and it also supports relative paths -
+note that the source files in the example above are in the root of the bucket,
+but they are copied into a subdirectory of the target bucket.
+
+`files` is a list of files to be copied.  The `name` is appended to the base of
+the filestore, and then the files are copied.  If the source file does not have
+the matching sha256, it will not be copied.
```
listx: I assume it will also print a warning for the hash mismatch?

pkg/filepromoter/file.go
```diff
@@ -119,10 +138,20 @@ func (o *copyFileOp) Run(ctx context.Context) error {
...
 }
 
-func computeSHA256ForFile(filename string) (string, error) {
+func ComputeSHA256ForFile(filename string) (string, error) {
+	hasher := sha256.New()
+	return computeHashForFile(filename, hasher)
+}
+
+func ComputeMD5ForFile(filename string) (string, error) {
+	hasher := md5.New()
```
listx: Nit: If md5.New() never fails and always returns a hasher, you might as well just pass that into the invocation at 151 and do away with this 'hasher' var.

But at that point, I think it would be reasonable to just use `computeHashForFile` directly and do away with these wrapper functions. /shrug

cmd/promobot-stage-and-hash-files/main.go
```diff
@@ -0,0 +1,140 @@
...
+	})
+
+	promoteOptions.Manifest = manifest
+
+	if err := cmd.RunPromoteFiles(ctx, promoteOptions); err != nil {
+		return err
+	}
+
+	// Print manifest to stdout
+	if _, err := os.Stdout.Write(manifestYAML); err != nil {
```
listx: I think it's better to write the manifest first and then ask the user if it's OK to run the promotion (dry-run or otherwise). It just seems backwards to have the promotion happen first before the manifest is first observed. User could pass in `--quiet` like for gcloud if they desire not to get prompted.

bartsmykla:
Hi @justinsb. What is the status of that PR? How much do we need to make it "merge" ready? :-)

PR #176, opened 2020-02-12
Files: add a command to generate manifests from a local path

justinsb:
This is helpful when uploading files, or when validating a list of
files.


cmd/promobot-generate-manifest/main.go
```diff
@@ -0,0 +1,84 @@
...
+
+func main() {
+	ctx := context.Background()
+
+	if err := run(ctx); err != nil {
+		fmt.Fprintf(os.Stderr, "%v\n", err)
+		// nolint[gomnd]
+		os.Exit(1)
+	}
+	os.Exit(0)
```
listx: Is this necessary?

cmd/promobot-generate-manifest/main.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	s, err := filepath.Abs(src)
+	if err != nil {
+		return xerrors.Errorf("cannot resolve %q to absolute path: %w", src, err)
+	}
+	src = s
+
+	var opt cmd.GenerateManifestOptions
+	opt.BaseDir = src
```
listx: Maybe just do `opt.BaseDir = s` and delete line 64?

pkg/cmd/hash.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+// GenerateManifestOptions holds the parameters for a hash-files operation
+type GenerateManifestOptions struct {
+	// BaseDir is the directory containing the files to hash
+	BaseDir string
+}
+
+// PopulateDefaults sets the default values for GenerateManifestOptions
+func (o *GenerateManifestOptions) PopulateDefaults() {
```
listx: Stub?

listx:
Feel free to cancel hold @justinsb

cmd/promobot-generate-manifest/main.go
```diff
@@ -0,0 +1,84 @@
...
+	"k8s.io/klog"
+	"sigs.k8s.io/k8s-container-image-promoter/pkg/cmd"
+	"sigs.k8s.io/yaml"
+)
+
+func main() {
+	ctx := context.Background()
+
+	if err := run(ctx); err != nil {
+		fmt.Fprintf(os.Stderr, "%v\n", err)
```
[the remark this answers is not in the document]
justinsb: Not strong reasons, but 3 reasons:

1) gives us a little more control over the output, both format and destination
2) we can specify a different exit code
3) it's a valid codepath - the task failed, we're not really `panic`ing

cmd/promobot-generate-manifest/main.go
justinsb: No.  Removing it - thanks!

cmd/promobot-generate-manifest/main.go
justinsb: Good call - done

pkg/cmd/hash.go
justinsb: Yeah, there currently aren't any fields that need initializing.  But we still want to follow the pattern. 
 Added a comment to make that clearer.

justinsb:
Thanks for looking @listx ... cleaned up the original commit and added a simple golden test.

PR #177, opened 2020-02-12
files: split manifests into filestores & files

justinsb:
This follows the same approach that images are also using.


listx:
Looks fine, but can you explain what filestores and files actually look like (in a readme doc)? Or alternatively, add a golden test?

listx:
Also, looks like you have to run the gazelle command to add in the xerrors dependency. I usually do `bazel run //:gazelle` to auto-add Bazel rules for new deps. You might have to run `make update`.

As for the lint check failures that all complain about `lll`, you might just want to add an exclusion rule for `lll` for the promotefiles sources.

listx:
> Looks fine, but can you explain what filestores and files actually look like (in a readme doc)? Or alternatively, add a golden test?

Ah, looks like you added docs in a separate PR https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/175. Nvm

justinsb:
Thanks for the tip about excluding `[lll]` at the file level!  I'll try to keep it inside the github line length (which I believe to be 120 characters) :-)

I added a golden image test here as well!

PR #179, opened 2020-02-13
WIP: files: support file:// as a source

justinsb:
This is helpful because then we can upload to GCS exactly as the
promoter will upload (e.g. compression).


bartsmykla:
Hi @justinsb, if you'll need someone to do osme quick review today I should be able to do so, so don't hessitate to write to me.

Issue #185, opened 2020-02-26
cip-auditor: fix log formatting error


listx:
During the initial prod deployment of the auditor (and some real-world testing) as described [here](https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/183), we discovered the following log printing error: 
![image](https://user-images.githubusercontent.com/725613/75305958-75537400-57fc-11ea-8393-d3128b1b9585.png)

This should be fixed!

listx:
I will work on this tomorrow to get it fixed ASAP.

listx:
This was fixed in #187.

/close

Issue #184, opened 2020-02-26
cip-auditor: always generate UUID for instance


listx:
cip-auditor runs with a UUID that is generated for each invocation. This UUID is injected into every message that is logged, to ease fetching logs from Stackdriver's web console. It currently fetches the UUID from the `CIP_AUDIT_TESTCASE_UUID` env var, but for production deployments, this var is not defined. And so we get log messages like `() s.ThinManifestDirPath: k8s.gcr.io` where the `()` is missing the UUID --- it should be inside the parentheses.

Instead, each time the auditor starts up, it should generate a new UUID for itself. We cannot use the UUID tied to the Cloud Run revision, because in the event that the auditor crashes, we want it to generate a new UUID.

listx:
Fixed with #187.

PR #187, opened 2020-03-04
Fix auditor

listx:
These address #184 and #185.

### DIFF
--- a/lib/audit/auditor.go
+++ b/lib/audit/auditor.go
@@ -392,7 +392,7 @@ func (s *ServerContext) Audit(w http.ResponseWriter, r *http.Request) {
 	// If we can't find the source registry for this image, then reject the
 	// transaction.
 	if string(srcRegistry.Name) == "" {
-		msg := fmt.Sprintf("(%s) TRANSACTION REJECTED: could not determine source registry: %v", s.ID)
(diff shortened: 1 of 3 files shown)


listx:
After this merges, I have to (1) promote the `cip-auditor` image to production (probably the `cip` image as well), and then (2) redeploy the auditor on Cloud Run.

lib/audit/auditor.go
```diff
@@ -392,7 +392,7 @@ func (s *ServerContext) Audit(w http.ResponseWriter, r *http.Request) {
 	// If we can't find the source registry for this image, then reject the
 	// transaction.
 	if string(srcRegistry.Name) == "" {
-		msg := fmt.Sprintf("(%s) TRANSACTION REJECTED: could not determine source registry: %v", s.ID)
+		msg := fmt.Sprintf("(%s) TRANSACTION REJECTED: could not determine source registry: %v", s.ID, gcrPayload)
```
tpepper: Is there any chance of over-logging with gcrPayload?  Sharing something that shouldn't be in the log?

listx: No, the  GCR payload only has very sparse details --- the image path, digest, and action (e.g., INSERT, DELETE)

PR #196, opened 2020-03-10
ServerContext: use a ReportingFacility wrapper for errorreporting

listx:
Instead of hardcoding the Stackdriver Error Reporting client directly
into the Auditor's ServerContext, use a "ReportingFacility" interface
wrapper so that we can use FakeReportingClient for unit tests.

/cc @justinsb

### DIFF
--- a/lib/report/gcp.go
+++ b/lib/report/gcp.go
@@ -1,0 +1,44 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


lib/report/gcp.go
```diff
@@ -0,0 +1,44 @@
...
+func NewGcpErrorReportingClient(
+	projectID, serviceName string,
+) *errorreporting.Client {
+
+	ctx := context.Background()
+
+	erc, err := errorreporting.NewClient(ctx, projectID, errorreporting.Config{
+		ServiceName: serviceName,
+		OnError: func(err error) {
+			klog.Errorf("Could not log error: %v", err)
```
justinsb: Might be nice to say "Could not log error to GCP: %v", or else this is going to look very confusing...

justinsb:
One nit, but lgtm

/approve
/lgtm

lib/report/gcp.go
listx: ACK

PR #197, opened 2020-03-10
gcp error reporting: clarify error

listx:
/cc @justinsb


justinsb:
Thanks - this will save some head-scratching!

/approve
/lgtm

PR #198, opened 2020-03-10
ServerContext: use a RemoteManifestFacility wrapper for cloning

listx:
Instead of hardcoding the Git operations directly into the Auditor's
ServerContext, use a "RemoteManifestFacility" interface wrapper so that
we can use a Fake for unit tests.

/cc @justinsb


justinsb:
Thanks @listx 

/approve
/lgtm

Issue #191, opened 2020-03-07
auditor fails to parse valid image references


listx:
During the backfill https://github.com/kubernetes/k8s.io/pull/632, a buggy behavior was observed where the auditor could not parse certain image references.

The following images all triggered a `TRANSACTION REJECTED: could not determine source registry:` error:

```
gcr.io/k8s-artifacts-prod/cluster-api-aws/cluster-api-aws-controller@sha256:050848d5a92c8c794af31726b982e8c29aff67e6043e9861e6c71d22d282aa6d
gcr.io/k8s-artifacts-prod/cluster-api-aws/cluster-api-aws-controller@sha256:465842186bfa79cf037163d2c632f544bf8ef0f42065351e6fef7fca7ae58fde
gcr.io/k8s-artifacts-prod/cluster-api-aws/cluster-api-aws-controller@sha256:6600f3e68b8139babd91d52339103c2f9cf0f61d87abcfb08140059047a05f50
gcr.io/k8s-artifacts-prod/cluster-api-aws/cluster-api-aws-controller@sha256:cdf2581188c5de863f6e3bf3180c4d537e5bac25998ceb17f38076d7a28d08ad
gcr.io/k8s-artifacts-prod/cluster-api-aws/cluster-api-aws-controller@sha256:dc5d27778cb5e341d7703001e6009c34ddd89ddcdcf63efaf8edeef22b2aa8ee
gcr.io/k8s-artifacts-prod/e2e-test-images/agnhost@sha256:1dfec5637a7010d6c0955c26f0a752266fa2646ed2bf8e6ad745cdcfcb611db8
gcr.io/k8s-artifacts-prod/e2e-test-images/agnhost@sha256:435e92e8fa45c3bfadb2c12374ea5bea2173d1afbeff572cdca4ffee115a1be7
gcr.io/k8s-artifacts-prod/e2e-test-images/agnhost@sha256:bfd170a7683c58cb23c08a8ed3516fcfaa3180503be43a2be9a9deb0e1dfb0f2
gcr.io/k8s-artifacts-prod/e2e-test-images/agnhost@sha256:d3c06db4ef3ac9f309819e52a401ce2393aef76d2e9657e60daab9027c281855
gcr.io/k8s-artifacts-prod/e2e-test-images/agnhost@sha256:d9d5a169f27bac0c58cce541eeef3db7c2ddd70eca09b48da484a04564a6b1f9
[...]

PR #199, opened 2020-03-12
fix child manifest handling

listx:
This should fix https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/191

The problem was that the parsing around child manifests was not done correctly. I have added a unit test to cover this case.

Putting this up now for feedback. I still have to (1) add more unit tests for Audit() and (2) add an e2e test for the bug in #191. But I would like to do those in separate PRs as this PR is already pretty big.

/cc @justinsb @thockin @dims @bartsmykla

### DIFF
--- a/lib/audit/BUILD.bazel
+++ b/lib/audit/BUILD.bazel
@@ -13,6 +13,7 @@ go_library(
         "//lib/logclient:go_default_library",
         "//lib/remotemanifest:go_default_library",
         "//lib/report:go_default_library",
+        "//lib/stream:go_default_library",
         "@com_google_cloud_go//errorreporting:go_default_library",
         "@io_k8s_klog//:go_default_library",
     ],
@@ -22,5 +23,10 @@ go_test(
     name = "go_default_test",
     srcs = ["auditor_test.go"],
     embed = [":go_default_library"],
-    deps = ["//lib/dockerregistry:go_default_library"],
+    deps = [
+        "//lib/dockerregistry:go_default_library",
+        "//lib/logclient:go_default_library",
+        "//lib/remotemanifest:go_default_library",
+        "//lib/report:go_default_library",
+    ],
 )
--- a/lib/audit/auditor.go
+++ b/lib/audit/auditor.go
@@ -190,9 +190,10 @@ func (s *ServerContext) Audit(w http.ResponseWriter, r *http.Request) {
 	logInfo.Printf("(%s) gcrPayload: %v", s.ID, gcrPayload)
 	logInfo.Printf("(%s) RemoteManifestFacility: %v", s.ID, s.RemoteManifestFacility)
 
-	// (3) Compare GCR state change with the intent of the promoter manifests.
+	// (3) Compare GCR state change with the intent of the promoter manifests
+	// (exact digest match on a tagged or tagless image).
 	for _, manifest := range manifests {
-		if manifest.Contains(*gcrPayload) {
+		if manifest.Contains(*gcrPayload)&(reg.FlagDigestMatched|reg.FlagTagMatched) != 0 {
 			msg := fmt.Sprintf(
 				"(%s) TRANSACTION VERIFIED: %v: agrees with manifest\n", s.ID, gcrPayload)
 			logInfo.Println(msg)
--- a/lib/audit/auditor.go
+++ b/lib/audit/auditor.go
@@ -253,7 +253,6 @@ func (s *ServerContext) Audit(w http.ResponseWriter, r *http.Request) {
 		true,
 		s.GcrReadingFacility.ReadRepo)
 	sc.ReadGCRManifestLists(s.GcrReadingFacility.ReadManifestList)
-	klog.Infof("sc.ParentDigest is: %v", sc.ParentDigest)
 	var childDigest reg.Digest
 	childImageParts := strings.Split(gcrPayload.Digest, "@")
 	if len(childImageParts) != 2 {
--- a/lib/audit/auditor.go
+++ b/lib/audit/auditor.go
@@ -261,8 +261,8 @@ func (s *ServerContext) Audit(w http.ResponseWriter, r *http.Request) {
 	sc.ReadRegistries(
 		[]reg.RegistryContext{srcRegistry},
 		true,
-		reg.MkReadRepositoryCmdReal)
-	sc.ReadGCRManifestLists(reg.MkReadManifestListCmdReal)
+		s.GcrReadingFacility.ReadRepo)
+	sc.ReadGCRManifestLists(s.GcrReadingFacility.ReadManifestList)
 	klog.Infof("sc.ParentDigest is: %v", sc.ParentDigest)
 	var childDigest reg.Digest
 	childImageParts := strings.Split(gcrPayload.Digest, "@")
--- a/lib/audit/auditor.go
+++ b/lib/audit/auditor.go
@@ -128,11 +128,17 @@ func ParsePubSubMessage(r *http.Request) (*reg.GCRPubSubPayload, error) {
 		return nil, fmt.Errorf(
 			"%v: deletions are prohibited", gcrPayload)
 	case "INSERT":
-		return &gcrPayload, nil
+		break
 	default:
 		return nil, fmt.Errorf(
 			"gcrPayload: unknown action %q", gcrPayload.Action)
 	}
+
+	if err := gcrPayload.PopulateExtraFields(); err != nil {
+		return nil, err
+	}
+
+	return &gcrPayload, nil
 }
 
 // Audit receives and processes a Pub/Sub push message. It has 3 parts: (1)
--- a/lib/audit/auditor.go
+++ b/lib/audit/auditor.go
@@ -111,9 +111,9 @@ func ParsePubSubMessage(r *http.Request) (*reg.GCRPubSubPayload, error) {
 		return nil, fmt.Errorf("json.Unmarshal: %v", err)
 	}
 
-	if len(gcrPayload.Digest) == 0 && len(gcrPayload.Tag) == 0 {
+	if len(gcrPayload.FQIN) == 0 && len(gcrPayload.PQIN) == 0 {
 		return nil, fmt.Errorf(
-			"gcrPayload: neither Digest nor Tag was specified")
+			"gcrPayload: neither 'digest' nor 'tag' was specified")
 	}
 
 	switch gcrPayload.Action {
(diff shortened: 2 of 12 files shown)


lib/dockerregistry/inventory.go
```diff
@@ -2547,22 +2570,41 @@ func (m Manifest) Contains(gcrPayload GCRPubSubPayload) bool {
...
 			}
+			r |= FlagPathMatched
 			for digest, tags := range image.Dmap {
 				fqin := ToFQIN(rc.Name, image.ImageName, digest)
-				if fqinOnly {
+				if len(gcrPayload.Tag) == 0 {
+					// The gcrPayload.Digest field actually holds the FQIN.
 					if gcrPayload.Digest == fqin {
-						return true
+						r |= FlagDigestMatched
```
bartsmykla: the same comment about bitwise operator as above, let's discuss if maybe we could provide some information about how it will behave for less experienced contributors?

lib/dockerregistry/inventory.go
```diff
@@ -2547,22 +2570,41 @@ func (m Manifest) Contains(gcrPayload GCRPubSubPayload) bool {
...
+			// If we have just a path match, return early because we should
+			// limit the scope of the search to just 1 image.
+			if r&FlagPathMatched != 0 {
+				return r
+			}
 		}
 	}
-	return false
+	return r
 }
```
bartsmykla: this method is VERY complex, do you think we can maybe simplify it somehow? Maybe spliting it into smaller functions would help?

lib/audit/BUILD.bazel
```diff
@@ -22,5 +23,10 @@ go_test(
     name = "go_default_test",
     srcs = ["auditor_test.go"],
     embed = [":go_default_library"],
-    deps = ["//lib/dockerregistry:go_default_library"],
+    deps = [
+        "//lib/dockerregistry:go_default_library",
+        "//lib/logclient:go_default_library",
+        "//lib/remotemanifest:go_default_library",
+        "//lib/report:go_default_library",
+    ],
```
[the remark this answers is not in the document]
listx: The `deps` is just a list of dependencies. The `auditor.go` file (which is part of `package audit`) now depends on these (new) local dependencies. BTW I didn't manually put this in, I just ran `bazel run //:gazelle` and it does all of these Bazel imports for me automatically.

lib/audit/auditor.go
```diff
@@ -190,9 +191,10 @@ func (s *ServerContext) Audit(w http.ResponseWriter, r *http.Request) {
 	logInfo.Printf("(%s) gcrPayload: %v", s.ID, gcrPayload)
 	logInfo.Printf("(%s) RemoteManifestFacility: %v", s.ID, s.RemoteManifestFacility)
 
-	// (3) Compare GCR state change with the intent of the promoter manifests.
+	// (3) Compare GCR state change with the intent of the promoter manifests
+	// (exact digest match on a tagged or tagless image).
 	for _, manifest := range manifests {
-		if manifest.Contains(*gcrPayload) {
+		if manifest.Contains(*gcrPayload)&(reg.FlagDigestMatched|reg.FlagTagMatched) != 0 {
```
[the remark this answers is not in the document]
listx: I can add maybe more explanations where I declare FlagDigestMatched instead. I'd prefer adding comments in a central place (where they are declared) over where they are used.

lib/dockerregistry/inventory.go
listx: I can split it up at least 1 more level. ACK

listx:
I think this PR is ready for merge. I'll work on adding more test cases in subsequent PRs.

/cc @justinsb

listx:
FYI gonna sync up with @justinsb on Monday to try to simplify this. Setting this to work-in-progres to reflect that.

/wip

lib/audit/auditor.go
```diff
@@ -190,9 +191,12 @@ func (s *ServerContext) Audit(w http.ResponseWriter, r *http.Request) {
 	logInfo.Printf("(%s) gcrPayload: %v", s.ID, gcrPayload)
 	logInfo.Printf("(%s) RemoteManifestFacility: %v", s.ID, s.RemoteManifestFacility)
 
-	// (3) Compare GCR state change with the intent of the promoter manifests.
+	// (3) Compare GCR state change with the intent of the promoter manifests
+	// (exact digest match on a tagged or tagless image).
 	for _, manifest := range manifests {
-		if manifest.Contains(*gcrPayload) {
+		m := gcrPayload.Match(manifest)
+		if m&(reg.FlagDigestMatched|reg.FlagTagMatched) != 0 &&
```
justinsb: Idea: I do wonder if the flags would be better expressed as a struct with bool members, particularly as some of them have true for match and some true for mismatch.

But ... what if match results `(bool, Info)`, where info was the bitset or struct with more info, and the bool was true if it's a match.

lib/audit/auditor.go
```diff
@@ -232,37 +236,23 @@ func (s *ServerContext) Audit(w http.ResponseWriter, r *http.Request) {
...
-		// Find the subproject key from the pub/sub message.
-		klog.Infof("subproject name: %s", rcNameParts[2])
-		if strings.HasSuffix(string(rc.Name), rcNameParts[2]) {
-			rcBound := rc // Avoid loop reference variable.
-			srcRegistry = rcBound
-			break
-		}
-	}
-	// If we can't find the source registry for this image, then reject the
+	// Find the subproject repository responsible for the GCRPubSubPayload. This
```
justinsb: It might be nice to split out PubSub (a transport) from the idea of a "Change" we are validating.  You could also then split this function into two, one that parses pubsub and one that validates a change.

lib/audit/types.go
```diff
@@ -17,18 +17,30 @@ limitations under the License.
...
 	"sigs.k8s.io/k8s-container-image-promoter/lib/remotemanifest"
 	"sigs.k8s.io/k8s-container-image-promoter/lib/report"
+	"sigs.k8s.io/k8s-container-image-promoter/lib/stream"
 )
 
+// GcrReadingFacility holds functions used to create streams for reading the
+// repository and manifest list.
+//
+// nolint[lll]
+type GcrReadingFacility struct {
```
justinsb: Nit: I don't love the name, but I'm not sure I understand it well enough to propose a better one.  `GcrReader`?

lib/dockerregistry/inventory.go
```diff
@@ -2527,42 +2527,121 @@ func GetDeleteCmd(
...
+type GcrPayloadMatch uint
+
+const (
+	// FlagPathMatched is set if the registry + image path (everything leading
+	// up to either the digest or a tag) matches.
+	FlagPathMatched GcrPayloadMatch = 1 << iota
+	// FlagDigestMatched is set if the digest in the payload matches a digest in
+	// the promoter manifest. This is ONLY matched if the path also matches.
+	FlagDigestMatched
+	// FlagTagMatched is ONLY matched if the digest also matches.
```
justinsb: I do think these rules suggest it isn't a bitset after all, hence why I suggested just returning `(bool, Info)`

lib/dockerregistry/inventory.go
```diff
@@ -2527,42 +2527,121 @@ func GetDeleteCmd(
...
+func (payload GCRPubSubPayload) Match(m Manifest) GcrPayloadMatch {
+	var r GcrPayloadMatch
 	for _, rc := range m.Registries {
-		if rc.Src {
+		r = payload.matchImages(rc, m.Images)
+		if r&FlagPathMatched != 0 {
+			return r
+		}
+	}
+	return r
```
justinsb: Is this ever non-zero?

lib/dockerregistry/inventory.go
```diff
@@ -2527,42 +2527,121 @@ func GetDeleteCmd(
...
+	// We do not look at source registries, because the payload will only
+	// contain the image name as it appears on the destination (production).
+	// So the prefix and substring checks only make sense for the
+	// destination registries.
+	if rc.Src {
+		return r
+	}
+	// Speed up the search by skipping over registry names whose leading
+	// characters do not match.
+	if !strings.HasPrefix(payload.Digest, (string)(rc.Name)) {
```
justinsb: We're comparing `Digest` against name - is that deliberate?

lib/dockerregistry/inventory.go
```diff
@@ -2527,42 +2527,121 @@ func GetDeleteCmd(
...
+
+	for _, image := range images {
+		r = payload.matchImage(rc, image)
+		// If we have just a path match, return early because we should
+		// limit the scope of the search to just 1 image.
+		if r&FlagPathMatched != 0 {
+			return r
+		}
+	}
+	return r
```
justinsb: Same question here - is this ever non-zero?  It makes the code harder to reason about if there's the potential for results carrying over

lib/dockerregistry/inventory.go
```diff
@@ -2527,42 +2527,121 @@ func GetDeleteCmd(
...
+}
+
+func (payload GCRPubSubPayload) matchImage(
+	rc RegistryContext,
+	image Image,
+) GcrPayloadMatch {
+
+	var r GcrPayloadMatch
+
+	if !strings.Contains(payload.Digest, (string)(image.ImageName)) {
```
justinsb: Again - digest vs image name?  Needs a comment if it's right :-)

lib/dockerregistry/inventory.go
```diff
@@ -2527,42 +2527,121 @@ func GetDeleteCmd(
...
+		return r
+	}
+	r |= FlagPathMatched
+	for digest, tags := range image.Dmap {
+		fqin := ToFQIN(rc.Name, image.ImageName, digest)
+		// The payload.Digest field actually holds the FQIN.
```
justinsb: Ah I see - and I presume that's an artifact of the GCR pubsub format?  If so, let's definitely convert that notification into something that is easier to work with...  even if we just rename Digest to `FQIN`

lib/dockerregistry/inventory.go
```diff
@@ -2527,42 +2527,121 @@ func GetDeleteCmd(
...
+	image Image,
+) GcrPayloadMatch {
+
+	var r GcrPayloadMatch
+
+	if !strings.Contains(payload.Digest, (string)(image.ImageName)) {
+		return r
+	}
+	r |= FlagPathMatched
+	for digest, tags := range image.Dmap {
```
justinsb: Do we need to loop?  It looks like we are only interested in one key value...

justinsb:
Two concrete suggestions that I think aren't too big changes and could make this much clearer:

1) Replace the bitset with a struct, and return `(bool, InfoStruct)`.  If there's a match, the first value is true and that's that.  If there's no match, we look at InfoStruct to see how bad things are.

2) Parse the pubsub message into the fields we want, and isolate ourselves from that rather than letting that flow into our code.  It looks like splitting out the digest and the tags from the name would simplify things.

listx:
> Two concrete suggestions that I think aren't too big changes and could make this much clearer:
> 
> 1. Replace the bitset with a struct, and return `(bool, InfoStruct)`.  If there's a match, the first value is true and that's that.  If there's no match, we look at InfoStruct to see how bad things are.

I think just replacing the bitset with a struct would be enough, but I'll try out your suggestion. I imagine this will just be a struct with some `bool`s in it.

> 2. Parse the pubsub message into the fields we want, and isolate ourselves from that rather than letting that flow into our code.  It looks like splitting out the digest and the tags from the name would simplify things.

This... Yes, it would make things a lot cleaner. I'll prioritize this change first.

listx:
Thinking about the `(bool, InfoStruct)` suggestion again, I'm not sure if it will make it clearer.

The problem here is that there are 2 dimensions of matching:
1. does payload match the promoter manifest, field-for-field (first the path, then digest, then tag (if present))?
2. if the payload matches up to the digest, and the payload has a tag field, does the payload's tag match the promoter manifest's tag? (FlagTagMismatch)

> Replace the bitset with a struct, and return (bool, InfoStruct). If there's a match, the first value is true and that's that. If there's no match, we look at InfoStruct to see how bad things are.

It's not clear what `bool` here would mean in the first part of the tuple.

listx:
@justinsb I've gone ahead with your second suggestion to rename the misnamed `Digest` field in the payload to be `FQIN` (also `PQIN` in place of `Tag`).

PR #209, opened 2020-04-29
enable testpackage linter

listx:


justinsb:
I do have my concerns over whether this linter is a good idea, but the changes aren't too tricky so we should investigate!

/approve
/lgtm

Issue #214, opened 2020-05-23
auditor: read from all matching source repos for parent/child manifest info


listx:
Recently, we've started creating promotion relationships where 2 different promoter manifests promote into the same destination repo. For example, consider https://github.com/kubernetes/k8s.io/pull/889 which made it so that both `gcr.io/k8s-staging-etcd` and `gcr.io/google-containers` promote into `{asia,eu,us}.gcr.io/k8s-artifacts-prod` (prod). This means that 2 different promoter manifests (the `k8s-staging-etcd` one (non-root-level) and the `gcr.io/google-containers` one (root-level)) can both potentially promote into prod.

Now consider the perspective of the auditor. When it receives a Pub/Sub payload for a child image that was promoted into prod, it only has information about the child digest, which is not in any of the promoter manifests (because the promoter manifests can operate only on parent digests). So the auditor has to look up the parent/child digest information from the *correct* source repo to verify that the child image was intentional. Also note that the Pub/Sub payload will only ever have information coming from prod (that is, GCR does not tell us from *which staging repo* a new image has been inserted into prod).

In the existing code, the promoter simply iterates through all of the promoter manifests and selects the staging repo from the *first* matching promoter manifest. "Matching" here means matching the prod insertion path (e.g., `us.gcr.io/k8s-artifacts-prod/etcd@sha256:...`) with a promoter manifest (looking at which destination repos the manifest intends to promote into). Recall from above that we have 2 different promoter manifests for the same prod repo (at the root level). So now, depending on the non-deterministic behavior of which one of the promoter manifests we choose first, we will select one staging repo or the other to read parent/child info from. This would not matter if both repos had the same content, but this is not guaranteed. If we are unlucky and select the "wrong" staging repo, the Pub/Sub message won't get verified and we'll get hit with a false positive alert.

The solution is to just read from all staging repos for all promoter manifests that match the Pub/Sub payload's insertion path, not just the first one that matches.

ETA for this bugfix is next week.

/cc @justinsb @thockin

PR #215, opened 2020-05-23
fix auditor behavior for redundant promotions

listx:
This addresses #214. The actual commit that does the fix is much simpler than expected and is the last commit in this PR (https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/215/commits/d19284e841056c3b51bbe47bc6c6d224a9567e33).

/cc @thockin @justinsb

### DIFF
--- a/lib/audit/auditor_test.go
+++ b/lib/audit/auditor_test.go
@@ -823,6 +823,23 @@ func TestAudit(t *testing.T) {
 				alert:  nil,
 			},
 		},
+		{
+			"child image promoted (two manifests both promote to the same destination, but only one source repo actually has the child's parent image referenced in the promoter manifest; payload refers to root insertion)",
+			manifests2,
+			reg.GCRPubSubPayload{
+				Action: "INSERT",
+				FQIN:   "us.gcr.io/k8s-artifacts-prod/etcd@sha256:0873d877318546c6569e1abfafd75e0625c202d24435299c4d2e57eeebea52ee",
+				PQIN:   "",
+			},
+			readRepo2,
+			readManifestList2,
+			expectedPatterns{
+				report: nil,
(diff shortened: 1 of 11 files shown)


lib/audit/auditor_test.go
```diff
@@ -352,6 +352,309 @@ func TestAudit(t *testing.T) {
...
 	}
 
+	// Regression test for the case where we are promoting
+	// bcdd5657b1edc1a2eb27356f33dd66b9400d4a084209c33461c7a7da0a32ebb3 (etcd
+	// tag "3.4.7-2"). It is crucial that we read fom *ALL* source registries,
+	// not just the first one that has a match for the destination registry
+	// "us.gcr.io/k8s-artifacts-prod" (because both manifests below promote to
+	// that same destination, but only one of them have the child images of
+	// 3.4.7-2).
+	manifests2 := []reg.Manifest{
```
thockin, line 362: You could make this a lot more readable with a couple `makeFoo()` helpers that take the params you want to populate and produce these structs - save on a bunch of curlies and lines.

lib/audit/auditor_test.go
```diff
@@ -352,6 +352,309 @@ func TestAudit(t *testing.T) {
...
+					ImageName: "etcd",
+					Dmap: reg.DigestTags{
+						"sha256:bcdd5657b1edc1a2eb27356f33dd66b9400d4a084209c33461c7a7da0a32ebb3": {"3.4.7-2"},
+					},
+				},
+			},
+		},
+	}
+
+	readRepo2 := map[string]string{
```
thockin, line 414: Is there any of this that can be elided, just to make it more readable?  It may be not..

listx: I could add some helpers to make test data like this easier to write (and more succinct), but I'm uncomfortable adding another layer of abstraction. Something to think about though, for sure.

thockin:
I meant just for test data.  See k/k pkg/proxy/iptables/proxier_test.go
makeTestService() for example.

PR #217, opened 2020-06-05
Using .bazelrc instead of BAZEL_BUILD_OPTS in Makefile

yodahekinsew:
Using .bazelrc instead of BAZEL_BUILD_OPTS in Makefile

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,0 +1,3 @@
+common --workspace_status_command=$PWD/workspace_status.sh
+common --host_force_python=PY2 
+common --stamp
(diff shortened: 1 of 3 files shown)


workspace_status.sh
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env bash
-
```
listx, line 2: Undo?

.bazelrc
```diff
@@ -0,0 +1,3 @@
+common --workspace_status_command=$PWD/workspace_status.sh
+common --host_force_python=PY2 
+common --stamp
```
listx, line 3: Can you add a newline here and also make sure your editor adds a trailing newline?

Makefile
```diff
@@ -1,45 +1,42 @@
 REPO_ROOT:=$(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
```
listx, line 1: Do we still need REPO_ROOT around? If not, let's delete it as well.

listx: Oops, looks like we still use it, nvm.

PR #218, opened 2020-06-05
Added new Make target for testing on Mac

yodahekinsew:
Current test target contains the "//..." invocation which causes issues on Mac, so added a new test-mac target to get around this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,14 @@ lint-ci: download
 	make lint
 test: build
 	bazel test --test_output=all //...
+# test-mac make target is a workaround for the following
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -26,6 +26,12 @@ lint-ci: download
 	make lint
 test: build
 	bazel test $(BAZEL_BUILD_OPTS) --test_output=all //...
+test-mac:
```
listx, line 29: Please add a comment just above this saying that this make target is a workaround for https://github.com/bazelbuild/rules_go/issues/2013

Issue #145, opened 2019-10-25
exit with an error if errors (auth errors, etc) are encountered during the promotion step


listx:
E.g., this job should have exited with an error: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-k8sio-cip/1187747140168847360 because the promotions failed due to an authorization error.

listx:
We just need to keep around a state variable that is globally written-to (use mutexes/locks to protect it across the various goroutines), which stores the state of the promoter, and exit with it.

listx:
/close

See #219

PR #219, opened 2020-06-05
Exiting with error if any errors are encountered in the promotion step

yodahekinsew:
Fix for Issue #145 
Added a state variable to check if any errors are encountered, and if so then the promotion function returns an error instead of nil. Also added a test (TestExecResults) in inventory_test.go.

### DIFF
--- a/lib/dockerregistry/inventory.go
+++ b/lib/dockerregistry/inventory.go
@@ -1619,7 +1619,7 @@ func MkReadManifestListCmdReal(
 // determines the number of workers to spawn.
 func (sc *SyncContext) ExecRequests(
 	populateRequests PopulateRequests,
-	processRequest ProcessRequest) {
+	processRequest ProcessRequest) error {
 	// Run requests.
 	MaxConcurrentRequests := 10
 	if sc.Threads > 0 {
@@ -1631,10 +1631,16 @@ func (sc *SyncContext) ExecRequests(
 	// We have to use a WaitGroup, because even though we know beforehand the
 	// number of workers, we don't know the number of jobs.
 	wg := new(sync.WaitGroup)
+
+	var err error
 	// Log any errors encountered.
 	go func() {
 		for reqRes := range requestResults {
 			if len(reqRes.Errors) > 0 {
+				(*mutex).Lock()
+				err = fmt.Errorf("Encountered an error during the" +
+					" promotion step")
+				(*mutex).Unlock()
 				klog.Errorf(
 					"Request %v: error(s) encountered: %v\n",
 					reqRes.Context,
@@ -1670,6 +1676,8 @@ func (sc *SyncContext) ExecRequests(
 	// goroutine (no need to put the error into a channel for consumption from a
 	// single point).
 	close(requestResults)
+
+	return err
 }
 
 func extractRegistryTags(reader io.Reader) (*ggcrV1Google.Tags, error) {
@@ -1855,8 +1863,11 @@ func SplitRegistryImagePath(
 		fmt.Errorf("could not determine registry name for '%v'", registryImagePath)
 }
 
+// MKPopulateRequestsForPromotionEdges takes in a map of PromotionEdges to promote
+// and a PromotionContext and returns a PopulateRequests which can generate
+// requests to be processed
 // nolint[lll]
-func mkPopulateRequestsForPromotionEdges(
+func MKPopulateRequestsForPromotionEdges(
 	toPromote map[PromotionEdge]interface{},
 	mkProducer PromotionContext) PopulateRequests {
 	return func(sc *SyncContext, reqs chan<- stream.ExternalRequest, wg *sync.WaitGroup) {
@@ -2055,7 +2066,7 @@ func (sc *SyncContext) Promote(
 		klog.Infof("  %v\n", edge)
 	}
 
-	var populateRequests = mkPopulateRequestsForPromotionEdges(
+	var populateRequests = MKPopulateRequestsForPromotionEdges(
 		edges,
 		mkProducer)
 
@@ -2158,13 +2169,14 @@ func (sc *SyncContext) Promote(
 	if customProcessRequest != nil {
 		processRequest = *customProcessRequest
 	}
-	sc.ExecRequests(populateRequests, processRequest)
+
+	err := sc.ExecRequests(populateRequests, processRequest)
 
 	if sc.DryRun {
 		sc.PrintCapturedRequests(&captured)
 	}
 
-	return nil
+	return err
 }
 
 // PrintCapturedRequests pretty-prints all given PromotionRequests.
@@ -2329,7 +2341,11 @@ func (sc *SyncContext) GarbageCollect(
 	if customProcessRequest != nil {
 		processRequest = *customProcessRequest
 	}
-	sc.ExecRequests(populateRequests, processRequest)
+
+	err := sc.ExecRequests(populateRequests, processRequest)
+	if err != nil {
+		klog.Info(err)
+	}
 
 	if sc.DryRun {
 		sc.PrintCapturedRequests(&captured)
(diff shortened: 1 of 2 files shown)


lib/dockerregistry/inventory.go
```diff
@@ -2158,13 +2164,19 @@ func (sc *SyncContext) Promote(
...
+	sc.ExecRequests(populateRequests, processRequest, &errorTracker)
 
 	if sc.DryRun {
 		sc.PrintCapturedRequests(&captured)
 	}
 
-	return nil
+	if errorTracker {
+		return fmt.Errorf("encountered an error during the promotion step")
+	} else {
```
listx, line 2177: No need to keep the `return nil` in the `else` clause.

lib/dockerregistry/inventory.go
```diff
@@ -1342,7 +1342,8 @@ func (sc *SyncContext) ReadRegistries(
 			wg.Add(-1)
 		}
 	}
-	sc.ExecRequests(populateRequests, processRequest)
+	errorTracker := false
```
listx, line 1345: s/errorTracker/errorEncountered

lib/dockerregistry/inventory.go
```diff
@@ -1856,7 +1862,7 @@ func SplitRegistryImagePath(
 }
 
 // nolint[lll]
-func mkPopulateRequestsForPromotionEdges(
+func MKPopulateRequestsForPromotionEdges(
```
listx, line 1865: Please add a comment for this function, now that it is exported.

lib/dockerregistry/inventory.go
```diff
@@ -2158,13 +2164,19 @@ func (sc *SyncContext) Promote(
 	if customProcessRequest != nil {
 		processRequest = *customProcessRequest
 	}
-	sc.ExecRequests(populateRequests, processRequest)
+	errorTracker := false
+
+	sc.ExecRequests(populateRequests, processRequest, &errorTracker)
```
listx, line 2169: I think you can just make ExecRequests return an error type. That way, you don't have to pass in `errorTracker` by reference to modify it. And also, you don't need to instantiate the variable outside of ExecRequests.

lib/dockerregistry/inventory.go
```diff
@@ -2329,7 +2341,9 @@ func (sc *SyncContext) GarbageCollect(
...
```
listx, line 2346: Let's at least log a "there was an error" or some such message if errorTracker is true in `GarbageCollect`. Likewise for `ClearRepository`.

lib/dockerregistry/inventory_test.go
```diff
@@ -2458,6 +2458,80 @@ func TestPromotion(t *testing.T) {
...
+				ImageName: "b",
+				Dmap: reg.DigestTags{
+					"sha256:111": {"0.9"}}}},
+		SrcRegistry: &srcRC}})
+	populateRequestsFakeError := reg.MKPopulateRequestsForPromotionEdges(
+		errorEdges,
+		nopStream)
+
+	captured := make(reg.CapturedRequests)
+	processRequestFake := reg.MkRequestCapturer(&captured)
```
listx, line 2519: Your `processRequestFake` function here needs to (always) populate the requestResults channel with an error. This will force your new ExecRequests function to always set the boolean you added to `true`, enabling you to test it here. Likewise your your sucessEdges case, you would have a `processRequestFakeSucceed` function that always processes the requests by populating requestResults with non-errors (probably just `reqRes.Context`).

lib/dockerregistry/inventory.go
```diff
@@ -1631,10 +1631,18 @@ func (sc *SyncContext) ExecRequests(
...
 	// number of workers, we don't know the number of jobs.
 	wg := new(sync.WaitGroup)
+
+	errorEncountered := false
+
 	// Log any errors encountered.
 	go func() {
 		for reqRes := range requestResults {
 			if len(reqRes.Errors) > 0 {
+				if !errorEncountered {
```
listx, line 1641: Please remove this optimization for now, unless we deem it necessary in the future.

lib/dockerregistry/inventory.go
listx: Maybe something like "... returns a PopulateRequests which can generate requests to be processed down the road."

lib/dockerregistry/inventory_test.go
```diff
@@ -2458,6 +2458,105 @@ func TestPromotion(t *testing.T) {
...
+		tag reg.Tag,
+		tp reg.TagOp) stream.Producer {
+
+		// We don't even need a stream producer, because we are not creating
+		// subprocesses that generate JSON or any other output; the vanilla
+		// "mkReq" in Promote() already stores all the info we need to check.
+		return nil
+	}
+
+	// successEdges, _ := reg.ToPromotionEdges([]reg.Manifest{reg.Manifest{}})
```
listx, line 2494: Debug? Please delete :)

lib/dockerregistry/inventory_test.go
```diff
@@ -2458,6 +2458,105 @@ func TestPromotion(t *testing.T) {
...
+			errors = append(errors, reg.Error{
+				Context: "Running TestExecRequests",
+				Error:   fmt.Errorf("This request results in an error")})
+			reqRes.Errors = errors
+			requestResults <- reqRes
+			wg.Add(-1)
+		}
+	}
+
+	successEncountered := sc.ExecRequests(populateRequests, processRequestSuccess)
```
listx, line 2545: Rename `successEncountered` to `errorEncountered`?

lib/dockerregistry/inventory_test.go
```diff
@@ -2458,6 +2458,93 @@ func TestPromotion(t *testing.T) {
...
+		}
+	}
+
+	errorEncountered := sc.ExecRequests(populateRequests, processRequestSuccess)
+	err := checkEqual(errorEncountered, nil)
+	checkError(t, err, fmt.Sprintf("checkError: test: %v (Error Tracking)\n",
+		"Error tracking for successful promotion"))
+
+	errorEncountered = sc.ExecRequests(populateRequests, processRequestError)
+	err = nil
```
listx, line 2539: Not sure why this line is necessary.

lib/dockerregistry/inventory_test.go
```diff
@@ -2458,6 +2458,93 @@ func TestPromotion(t *testing.T) {
...
+	if errorEncountered == nil {
```
listx, line 2540: You should just reuse the checkEqual() function here instead.

lib/dockerregistry/inventory.go
```diff
@@ -2329,7 +2344,11 @@ func (sc *SyncContext) GarbageCollect(
 	if customProcessRequest != nil {
 		processRequest = *customProcessRequest
 	}
-	sc.ExecRequests(populateRequests, processRequest)
+
+	errorEncountered := sc.ExecRequests(populateRequests, processRequest)
+	if errorEncountered != nil {
+		klog.Infof(errorEncountered.Error())
```
listx, line 2350: You could use `klog.Info(err)` here directly, probably. No use in using `Infof` if you are not using any formatting string like `"%v"`.

lib/dockerregistry/inventory.go
```diff
@@ -1631,10 +1631,16 @@ func (sc *SyncContext) ExecRequests(
 	// We have to use a WaitGroup, because even though we know beforehand the
 	// number of workers, we don't know the number of jobs.
 	wg := new(sync.WaitGroup)
+
+	var err error = nil
```
listx, line 1635: FYI in Golang just `var err error` does the same thing. If you just declare vars without providing the value, it just gets assigned the default value.

lib/dockerregistry/inventory.go
```diff
@@ -2329,7 +2341,11 @@ func (sc *SyncContext) GarbageCollect(
 	if customProcessRequest != nil {
 		processRequest = *customProcessRequest
 	}
-	sc.ExecRequests(populateRequests, processRequest)
+
+	err := sc.ExecRequests(populateRequests, processRequest)
+	if err != nil {
+		klog.Info(err.Error())
```
listx, line 2347: I guess we don't get the same thing if we just do `klog.Info(err)` ?

PR #221, opened 2020-06-17
Removed unused 'verbosity' flag from the promoter

yodahekinsew:
Removed unused 'verbosity' flag from the promoter and any use of verbosity in SyncContext type since they weren't being used. Verbosity levels from now on will be set using klog's expected "v" flag since we already use klog to handle all logging. This will also allow for future use of the "klog.V(int)" syntax to determine whether something should be logged.


listx:
Thanks!

/lgtm
/approve

PR #222, opened 2020-06-17
Added a PreCheck Interface for adding future checks against pull requests

yodahekinsew:
The PullCheck interface will make it easier for engineers in the future to add checks for the promoter against pull requests that attempt to modify manifests.

### DIFF
--- a/cip.go
+++ b/cip.go
@@ -345,6 +345,14 @@ func main() {
 		os.Exit(0)
 	}
 
+	// Check the pull request
+	if *dryRunPtr {
+		err = sc.RunChecks(promotionEdges, []reg.PreCheck{})
+		if err != nil {
+			klog.Exitln(err)
+		}
+	}
+
 	// Promote.
 	mkProducer := func(
 		srcRegistry reg.RegistryName,
--- a/lib/dockerregistry/inventory.go
+++ b/lib/dockerregistry/inventory.go
@@ -1952,6 +1952,28 @@ func MKPopulateRequestsForPromotionEdges(
 	}
 }
 
+// RunChecks runs defined PreChecks in order to check the promotion.
+func (sc *SyncContext) RunChecks(
+	edges map[PromotionEdge]interface{},
+	preChecks []PreCheck,
+) error {
+
+	var errors []error
+	for _, preCheck := range preChecks {
+		err := preCheck.Run(edges)
+		if err != nil {
+			klog.Error(err)
+			errors = append(errors, err)
+		}
+	}
+
+	if errors != nil {
+		return fmt.Errorf("%v error(s) encountered during the prechecks",
+			len(errors))
+	}
+	return nil
+}
+
 // FilterPromotionEdges generates all "edges" that we want to promote.
 func (sc *SyncContext) FilterPromotionEdges(
 	edges map[PromotionEdge]interface{},
--- a/lib/dockerregistry/types.go
+++ b/lib/dockerregistry/types.go
@@ -60,6 +60,15 @@ type SyncContext struct {
 	ParentDigest      ParentDigest
 }
 
+// PreCheck represents a check function to run against a pull request that
+// modifies the promoter manifests before oking promotion of the changes.
+//
+// Run runs the defined check on a set of PromotionEdges and returns an error
+// if the check fails, returns nil otherwise.
+type PreCheck interface {
+	Run(edges map[PromotionEdge]interface{}) error
+}
+
 // PromotionEdge represents a promotion "link" of an image repository between 2
 // registries.
 type PromotionEdge struct {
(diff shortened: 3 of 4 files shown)


cip.go
```diff
@@ -345,6 +345,12 @@ func main() {
 		os.Exit(0)
 	}
 
+	// Check the pull request
```
listx, line 348: This check can't run unconditionally under main(). It should only run when we are doing a PR on dry run when comparing the PR vs master. You should use the PULL_PULL_REF and other env vars to determine whether you are running in a PR or not. Or just pass an explicit flag in.

cip.go
```diff
@@ -345,6 +345,12 @@ func main() {
...
+	pullRequestOK := sc.RunChecks(promotionEdges, []reg.PullCheck{})
```
listx, line 349: You should pass in a non-empty list of checks to iterate over.

cip.go
```diff
@@ -345,6 +345,12 @@ func main() {
...
+	if pullRequestOK != nil {
+		klog.Exitln("encountered error with pull request")
```
listx, line 351: You should log pullRequestOK as well.

lib/dockerregistry/inventory.go
```diff
@@ -1952,6 +1952,20 @@ func MKPopulateRequestsForPromotionEdges(
...
 
+// RunChecks runs input PullChecks in order to check the promotion.
+func (sc *SyncContext) RunChecks(
+	edges map[PromotionEdge]interface{},
+	pullChecks []PullCheck,
+) error {
+	for _, pullCheck := range pullChecks {
+		err := pullCheck.Run(edges)
+		if err != nil {
+			return err
```
listx, line 1963: You should collect all errors and bubble up all failures, not just the first one to fail.

cip.go
yodahekinsew: I'm passing in an anonymous list right now that can be filled with checks later on as they are added

listx: SG

lib/dockerregistry/inventory.go
```diff
@@ -1952,6 +1952,26 @@ func MKPopulateRequestsForPromotionEdges(
...
+
+	var errors []error
+	for _, preCheck := range preChecks {
+		err := preCheck.Run(edges)
+		if err != nil {
+			errors = append(errors, err)
+		}
+	}
+
+	if errors != nil {
```
listx, line 1969: Can you iterate over each error here and print it?

lib/dockerregistry/inventory.go
```diff
@@ -1952,6 +1952,26 @@ func MKPopulateRequestsForPromotionEdges(
...
+		return fmt.Errorf("An error was encountered in one of the prechecks")
```
listx, line 1970: This message is no longer accurate because you may have more than 1 error encountered.

cip.go
```diff
@@ -345,6 +345,14 @@ func main() {
 		os.Exit(0)
 	}
 
+	// Check the pull request
+	if *dryRunPtr {
+		checksOK := sc.RunChecks(promotionEdges, []reg.PreCheck{})
```
listx, line 350: `checksOK` should be named `err` because RunChecks() returns an error.

lib/dockerregistry/inventory.go
```diff
@@ -1952,6 +1952,38 @@ func MKPopulateRequestsForPromotionEdges(
 	}
 }
 
+// CheckPromotion is where PreChecks to be run are defined.
```
listx, line 1955: I think you can delete `CheckPromotion` and just use sc.RunChecks directly at line 350.

lib/dockerregistry/inventory.go
```diff
@@ -1952,6 +1952,38 @@ func MKPopulateRequestsForPromotionEdges(
...
+	var errors []error
+	for _, preCheck := range preChecks {
+		err := preCheck.Run(edges)
+		if err != nil {
+			errors = append(errors, err)
+		}
+	}
+
+	if errors != nil {
+		for _, err := range errors {
```
listx, line 1978: You can just log the errors as you append them in the for-loop above this. Then you only need 1 for-loop instead of 2.

lib/dockerregistry/types.go
```diff
@@ -60,6 +60,12 @@ type SyncContext struct {
 	ParentDigest      ParentDigest
 }
 
+// PreCheck represents a check function to run against a pull request that
+// modifies the promoter manifests before oking promotion of the changes.
```
listx, line 64:
```
// PreCheck is the interface that wraps the basic Run method.
//
// <... explain in more detail here what Run() does>
```

Take a look at the official Golang library docs to get a sense of the commenting style, for example https://golang.org/src/io/io.go#L115

prathak07:
testing

PR #223, opened 2020-06-18
auditor: use shorter ExponentialBackoff for Cloud Run

listx:
Cloud Run has a 15 minute quota (or time limit) for requests to finish
"running". Otherwise the request is "terminated" [1]. Because we don't
want our requests to be rudely interrupted by Cloud Run, we use a
shorter (~13 minute) timeout, named BackoffCloudRun.

This in itself doesn't get us much, but it does take us out of
poorly-documented behavior (for example, it is unclear what exactly
happens to the container when the request is terminated abruptly by
Cloud Run).

[1]: https://github.com/kubernetes/k8s.io/pull/955#issuecomment-645794886


listx:
/hold

I have a better idea about this actually. It doesn't involve setting time limits like this at all...!

listx:
The idea is: since we're already using the Pub/Sub subscription to hold messages (with retry capability for the duration of the maximum retention period of each message, which is currently set to 7 days), we don't have to bother with our own network retry function. Instead we can just return HTTP 400 or 500 (per https://cloud.google.com/run/docs/tutorials/pubsub#looking_at_the_code) which results in the message being retried. I believe doing this will also reduce the number of messages sent at a time by 1/2 from PubSub to us (as described in the Slow-Start Algorithm in this video https://youtu.be/k8lBOV-nRXI), which is similar to the idea of an ExponentialBackoff. The only difference is that ultimately our auditor will get just 1 PubSub message at a time, which is fine. 

In other words, using Pub/Sub already gives us the ability to retry requests --- so the use of ExponentialBackoff here is unnecessary. So if we see a GCR read error of any kind, we should just return HTTP 504 (which isn't entirely accurate but is better than a generic HTTP 500 response code). This will (1) make PubSub retry the request for us for up to 7 days, and (2) reduce the concurrent number of messages being sent by PubSub to us by 1/2. Doing this alone should at least limit the number of alerts we have seen before we get into batching PubSub requests.

dims:
Do we still needs this @listx ?

listx:
> Do we still needs this @listx ?

Yes, this is still needed. It's all part of the plan to make the auditor scale better.

PR #224, opened 2020-06-19
WIP: Adding a JSON Summary for Promoter

yodahekinsew:
The Promoter will collect errors that are generated throughout the job and then log them as a prettified JSON when the job ends.

### DIFF
--- a/cip.go
+++ b/cip.go
@@ -53,6 +53,10 @@ func main() {
 	threadsPtr := flag.Int(
 		"threads",
 		10, "number of concurrent goroutines to use when talking to GCR")
+	jsonLogSummaryPtr := flag.Bool(
+		"json-log-summary",
+		false,
+		"only log a json summary of important errors")
 	parseOnlyPtr := flag.Bool(
 		"parse-only",
 		false,
@@ -334,6 +338,10 @@ func main() {
 		fmt.Print(snapshot)
 		os.Exit(0)
 	}
+
+	if *jsonLogSummaryPtr {
+		defer sc.LogJSONSummary()
+    }
 
 	// Check the pull request
 	if *dryRunPtr {
--- a/lib/dockerregistry/inventory.go
+++ b/lib/dockerregistry/inventory.go
@@ -106,6 +106,17 @@ func MakeSyncContext(
 	return sc, nil
 }
 
+// LogJSONSummary logs the SyncContext's Logs as a prettified JSON.
+func (sc *SyncContext) LogJSONSummary() {
+	json, err := json.MarshalIndent(sc.Logs, "", "  ")
+	if err != nil {
+		klog.Infof("There was a problem generating the JSON summary: %v",
+			err)
+	} else {
+		klog.Info(string(json))
+	}
+}
+
 // ParseManifestFromFile parses a Manifest from a filepath.
 func ParseManifestFromFile(filePath string) (Manifest, error) {
 
@@ -1616,6 +1627,8 @@ func MkReadManifestListCmdReal(
 
 // ExecRequests uses the Worker Pool pattern, where MaxConcurrentRequests
 // determines the number of workers to spawn.
+//
+// nolint[funlen]
 func (sc *SyncContext) ExecRequests(
 	populateRequests PopulateRequests,
 	processRequest ProcessRequest) error {
@@ -1639,6 +1652,7 @@ func (sc *SyncContext) ExecRequests(
 				(*mutex).Lock()
 				err = fmt.Errorf("Encountered an error during the" +
 					" promotion step")
+				sc.Logs.Errors = append(sc.Logs.Errors, reqRes.Errors...)
 				(*mutex).Unlock()
 				klog.Errorf(
 					"Request %v: error(s) encountered: %v\n",
--- a/lib/dockerregistry/types.go
+++ b/lib/dockerregistry/types.go
@@ -45,6 +45,11 @@ type Error struct {
 // is used for both -dry-run and testing.
 type CapturedRequests map[PromotionRequest]int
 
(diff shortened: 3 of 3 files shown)


lib/dockerregistry/inventory.go
```diff
@@ -106,6 +106,30 @@ func MakeSyncContext(
...
+	for _, err := range sc.Logs.Errors {
+		JSONString += "\t\t{\n" +
+			"\t\t\tError: " + err.Error.Error() + "\n" +
+			"\t\t\tComment: " + err.Context + "\n" +
+			"\t\t}\n"
+	}
+	JSONString += "\t]\n"
+	JSONString += "}"
+	klog.Info(JSONString)
+}
```
yodahekinsew, line 131: this is what I was using to prettify the JSON before I learned of the existence of Marshall functions, so will be changed in future

cip.go
```diff
@@ -335,6 +339,10 @@ func main() {
 		os.Exit(0)
 	}
 
+	if *jsonLogSummaryPtr {
+		defer sc.LogJSONSummary()
+  }
```
listx, line 344: Looks like you didn't run the formatter for this (bracket looks off).

lib/dockerregistry/inventory.go
```diff
@@ -106,6 +106,23 @@ func MakeSyncContext(
...
+// stored in the SyncContext's Logs field.
+func (sc *SyncContext) CollectErrors(errors ...Error) {
+	for _, err := range errors {
+		sc.Logs.Errors = append(sc.Logs.Errors, err)
+	}
+}
+
+// LogJSONSummary logs the SyncContext's Logs as a prettified JSON.
+func (sc *SyncContext) LogJSONSummary() {
+	json, err := json.MarshalIndent(sc.Logs, "", "\t")
```
listx, line 119: Can you use 2-spaces instead of the tab character?

lib/dockerregistry/inventory.go
```diff
@@ -106,6 +106,23 @@ func MakeSyncContext(
...
+	if err != nil {
+		klog.Info("There was a problem generating the JSON summary.")
```
listx, line 121: Always always always log the error as well. It is very bad practice to hide information from errors, unless of course such errors contain sensitive information (which is not the case here).

lib/dockerregistry/inventory.go
```diff
@@ -1636,14 +1653,14 @@ func (sc *SyncContext) ExecRequests(
 	go func() {
 		for reqRes := range requestResults {
 			if len(reqRes.Errors) > 0 {
+				klog.Errorf(
+					"Request %v: error(s) encountered: %v\n",
+					reqRes.Context, reqRes.Errors)
 				(*mutex).Lock()
 				err = fmt.Errorf("Encountered an error during the" +
 					" promotion step")
+				sc.CollectErrors(reqRes.Errors...)
```
listx, line 1662: I think you can replace this line with a call to `append()` directly. I'm pretty sure it can accept a variable number of arguments.

lib/dockerregistry/inventory.go
```diff
@@ -1636,14 +1653,14 @@ func (sc *SyncContext) ExecRequests(
 	go func() {
 		for reqRes := range requestResults {
 			if len(reqRes.Errors) > 0 {
+				klog.Errorf(
```
listx, line 1656: Not sure why you moved this block up here. Please move back?

lib/dockerregistry/types.go
```diff
@@ -39,12 +39,18 @@ type Errors []Error
 type Error struct {
 	Context string
 	Error   error
+	Info    string
```
listx, line 42: Are we using this?

yodahekinsew: The plan was to use it as a field for extra comments that aren't included in the Error or Context field. For example, telling a user to remove the following images which are too large or something.

listx: OK, but if we're not using it in this PR then I suggest not adding it until we actually do need it.

lib/dockerregistry/inventory.go
```diff
@@ -1639,11 +1650,11 @@ func (sc *SyncContext) ExecRequests(
 				(*mutex).Lock()
 				err = fmt.Errorf("Encountered an error during the" +
 					" promotion step")
-				(*mutex).Unlock()
+				sc.Logs.Errors = append(sc.Logs.Errors, reqRes.Errors...)
 				klog.Errorf(
```
listx, line 1654: This call to klog shouldn't be in the critical section (locked area).

lib/dockerregistry/inventory.go
```diff
@@ -1639,11 +1650,11 @@ func (sc *SyncContext) ExecRequests(
 				(*mutex).Lock()
 				err = fmt.Errorf("Encountered an error during the" +
 					" promotion step")
+				sc.Logs.Errors = append(sc.Logs.Errors, reqRes.Errors...)
 				(*mutex).Unlock()
 				klog.Errorf(
 					"Request %v: error(s) encountered: %v\n",
-					reqRes.Context,
-					reqRes.Errors)
+					reqRes.Context, reqRes.Errors)
```
listx, line 1657: This change isn't necessary. Let's try to keep the diff clean...!

listx: The way I would do this is: squash all 6 commits into 1, then use `git reset ...` to let git perform the "undo" for me.

listx: Undo?

listx:
There is still 1 unresolved comment, PTAL

listx:
/lgtm
/approve

Thanks!

PR #226, opened 2020-06-23
WIP: Adding image removal check

yodahekinsew:
This is the groundwork for a new check against pull requests that attempt to remove images from the promoter manifests.

### DIFF
--- a/cip.go
+++ b/cip.go
@@ -337,7 +337,7 @@ func main() {
 
 	// Check the pull request
 	if *dryRunPtr {
-		err = sc.RunChecks(promotionEdges, []reg.PreCheck{})
+		err = sc.RunChecks([]reg.PreCheck{})
 		if err != nil {
 			klog.Exitln(err)
 		}
--- a/lib/dockerregistry/checks.go
+++ b/lib/dockerregistry/checks.go
@@ -1,0 +1,166 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package inventory
+
+import (
+	"encoding/hex"
+	"fmt"
+	"os"
+	"strings"
+
+	gogit "gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+)
+
+// ImageRemovalCheck implements the PreCheck interface and checks against
+// pull requests that attempt to remove any images from the promoter manifests.
+type ImageRemovalCheck struct {
+	GitRepoPath    string
+	MasterSHA      plumbing.Hash
+	PullRequestSHA plumbing.Hash
+	PullEdges      map[PromotionEdge]interface{}
+}
+
+func getGitShaFromEnv(envVar string) (plumbing.Hash, error) {
+	potenitalSHA := os.Getenv(envVar)
+	const gitShaLength = 40
+	if len(potenitalSHA) != gitShaLength {
+		return plumbing.Hash{},
+			fmt.Errorf("Length of SHA is %d characters, should be %d",
+				len(potenitalSHA), gitShaLength)
+	}
+	_, err := hex.DecodeString(potenitalSHA)
+	if err != nil {
+		return plumbing.Hash{}, fmt.Errorf("Not a valid SHA: %v", err)
+	}
+	return plumbing.NewHash(potenitalSHA), nil
+}
+
+// MKRealImageRemovalCheck returns an instance of ImageRemovalCheck.
+func MKRealImageRemovalCheck(
+	gitRepoPath string,
+	edges map[PromotionEdge]interface{},
+) (*ImageRemovalCheck, error) {
+	// The "PULL_BASE_SHA" and "PULL_PULL_SHA" environment variables are given
+	// by the PROW job running the promoter container and represent the Git SHAs
+	// for the master branch and the pull request branch respectively.
+	masterSHA, err := getGitShaFromEnv("PULL_BASE_SHA")
+	if err != nil {
+		return nil, fmt.Errorf("The PULL_BASE_SHA environment variable "+
+			"is invalid: %v", err)
+	}
+	pullRequestSHA, err := getGitShaFromEnv("PULL_PULL_SHA")
+	if err != nil {
+		return nil, fmt.Errorf("The PULL_PULL_SHA environment variable "+
+			"is invalid: %v", err)
+	}
+	return &ImageRemovalCheck{
+		gitRepoPath,
+		masterSHA,
+		pullRequestSHA,
+		edges,
+	}, nil
+}
+
+// Run executes ImageRemovalCheck on a set of promotion edges.
+// Returns an error if the pull request removes images from the
+// promoter manifests.
+func (check *ImageRemovalCheck) Run() error {
+	r, err := gogit.PlainOpen(check.GitRepoPath)
+	if err != nil {
+		return fmt.Errorf("Could not open the Git repo: %v", err)
+	}
+	w, err := r.Worktree()
+	if err != nil {
+		return fmt.Errorf("Could not create Git worktree: %v", err)
+	}
+
+	// The Prow job that this check is running in has already cloned the
+	// git repo for us so we can just checkout the master branch to get the
+	// master branch's version of the promoter manifests.
+	err = w.Checkout(&gogit.CheckoutOptions{
+		Hash:  check.MasterSHA,
+		Force: true,
+	})
+	if err != nil {
+		return fmt.Errorf("Could not checkout the master branch of the Git"+
+			" repo: %v", err)
+	}
+
+	mfests, err := ParseThinManifestsFromDir(check.GitRepoPath)
+	if err != nil {
+		return fmt.Errorf("Could not parse manifests from the directory: %v",
+			err)
+	}
+	masterEdges, err := ToPromotionEdges(mfests)
+	if err != nil {
+		return fmt.Errorf("Could not generate promotion edges from promoter"+
+			" manifests: %v", err)
+	}
+
+	// Reset the current directory back to the pull request branch so that this
+	// check doesn't leave lasting effects that could affect subsequent checks.
+	err = w.Checkout(&gogit.CheckoutOptions{
+		Hash:  check.PullRequestSHA,
+		Force: true,
+	})
+	if err != nil {
+		return fmt.Errorf("Could not checkout the pull request branch of the"+
+			" Git repo %v: %v",
+			check.GitRepoPath, err)
+	}
+
+	return check.Compare(masterEdges, check.PullEdges)
+}
+
+// Compare is a function of the ImageRemovalCheck that handles
+// the comparison of the pull requests's set of promotion edges and
+// the master branch's set of promotion edges.
+func (check *ImageRemovalCheck) Compare(
+	edgesMaster map[PromotionEdge]interface{},
+	edgesPullRequest map[PromotionEdge]interface{},
+) error {
+	// Generate a set of all destination images that appear in
+	// the pull request's set of promotion edges.
+	destinationImages := make(map[PromotionEdge]interface{})
+	for edge := range edgesPullRequest {
+		destinationImages[PromotionEdge{
+			DstImageTag: edge.DstImageTag,
+			Digest:      edge.Digest,
+		}] = nil
+	}
+
+	// Check that every destination image in the master branch's set of
+	// promotion edges exists in the pull request's set of promotion edges.
+	removedImages := make([]string, 0)
+	for edge := range edgesMaster {
+		_, found := destinationImages[PromotionEdge{
+			DstImageTag: edge.DstImageTag,
+			Digest:      edge.Digest,
+		}]
+		if !found {
+			removedImages = append(removedImages,
+				string(edge.DstImageTag.ImageName))
+		}
+	}
+
+	if len(removedImages) > 0 {
+		return fmt.Errorf("The following images were removed in this pull "+
+			"request: %v", strings.Join(removedImages, ", "))
+	}
+	return nil
+}
--- a/lib/dockerregistry/checks_test.go
+++ b/lib/dockerregistry/checks_test.go
@@ -1,0 +1,170 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package inventory_test
+
+import (
+	"fmt"
+	"testing"
+
+	reg "sigs.k8s.io/k8s-container-image-promoter/lib/dockerregistry"
+)
+
+func TestImageRemovalCheck(t *testing.T) {
+	srcRegName := reg.RegistryName("gcr.io/foo")
+	srcRegName2 := reg.RegistryName("gcr.io/foo2")
+	destRegName := reg.RegistryName("gcr.io/bar")
+	destRC := reg.RegistryContext{
+		Name:           destRegName,
+		ServiceAccount: "robot",
+	}
+	srcRC := reg.RegistryContext{
+		Name:           srcRegName,
+		ServiceAccount: "robot",
+		Src:            true,
+	}
+	srcRC2 := reg.RegistryContext{
+		Name:           srcRegName2,
+		ServiceAccount: "robot",
+		Src:            true,
+	}
+	registries := []reg.RegistryContext{destRC, srcRC}
+	registries2 := []reg.RegistryContext{destRC, srcRC, srcRC2}
+
+	imageA := reg.Image{
+		ImageName: "a",
+		Dmap: reg.DigestTags{
+			"sha256:000": {"0.9"}}}
+	imageA2 := reg.Image{
+		ImageName: "a",
+		Dmap: reg.DigestTags{
+			"sha256:111": {"0.9"}}}
+	imageB := reg.Image{
+		ImageName: "b",
+		Dmap: reg.DigestTags{
+			"sha256:000": {"0.9"}}}
+
+	var tests = []struct {
+		name            string
+		check           reg.ImageRemovalCheck
+		masterManifests []reg.Manifest
+		pullManifests   []reg.Manifest
+		expected        error
+	}{
+		{
+			"Empty manifests",
+			reg.ImageRemovalCheck{},
+			[]reg.Manifest{},
+			[]reg.Manifest{},
+			nil,
+		},
+		{
+			"Same manifests",
+			reg.ImageRemovalCheck{},
+			[]reg.Manifest{
+				{
+					Registries: registries,
+					Images: []reg.Image{
+						imageA,
+					},
+					SrcRegistry: &srcRC},
+			},
+			[]reg.Manifest{
+				{
+					Registries: registries,
+					Images: []reg.Image{
+						imageA,
+					},
+					SrcRegistry: &srcRC},
+			},
+			nil,
+		},
+		{
+			"Different manifests",
+			reg.ImageRemovalCheck{},
+			[]reg.Manifest{
+				{
+					Registries: registries,
+					Images: []reg.Image{
+						imageA,
+					},
+					SrcRegistry: &srcRC},
+			},
+			[]reg.Manifest{
+				{
+					Registries: registries,
+					Images: []reg.Image{
+						imageB,
+					},
+					SrcRegistry: &srcRC},
+			},
+			fmt.Errorf("The following images were removed in this pull " +
+				"request: a"),
+		},
+		{
+			"Promoting same image from different registry",
+			reg.ImageRemovalCheck{},
+			[]reg.Manifest{
+				{
+					Registries: registries2,
+					Images: []reg.Image{
+						imageA,
+					},
+					SrcRegistry: &srcRC},
+			},
+			[]reg.Manifest{
+				{
+					Registries: registries2,
+					Images: []reg.Image{
+						imageA,
+					},
+					SrcRegistry: &srcRC2},
+			},
+			nil,
+		},
+		{
+			"Promoting image with same name and different digest",
+			reg.ImageRemovalCheck{},
+			[]reg.Manifest{
+				{
+					Registries: registries,
+					Images: []reg.Image{
+						imageA,
+					},
+					SrcRegistry: &srcRC},
+			},
+			[]reg.Manifest{
+				{
+					Registries: registries,
+					Images: []reg.Image{
+						imageA2,
+					},
+					SrcRegistry: &srcRC},
+			},
+			fmt.Errorf("The following images were removed in this pull " +
+				"request: a"),
+		},
+	}
+
+	for _, test := range tests {
+		masterEdges, _ := reg.ToPromotionEdges(test.masterManifests)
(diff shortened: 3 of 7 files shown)


cip.go
```diff
@@ -337,7 +337,12 @@ func main() {
 
 	// Check the pull request
 	if *dryRunPtr {
-		err = sc.RunChecks(promotionEdges, []reg.PreCheck{})
+		err = sc.RunChecks(promotionEdges, []reg.PreCheck{
+			reg.ImageRemovalCheck{
+				*auditManifestRepoUrlPtr,
+				os.Getenv("PULL_BASE_SHA"),
```
listx, line 343: You should move out the call to os.Getenv() out of this argument list. As a general rule, unless you know for a fact that the function cannot fail, always execute such dependencies early and then pass along the results as an argument to another function call.

cip.go
```diff
@@ -337,7 +337,12 @@ func main() {
 
 	// Check the pull request
 	if *dryRunPtr {
-		err = sc.RunChecks(promotionEdges, []reg.PreCheck{})
+		err = sc.RunChecks(promotionEdges, []reg.PreCheck{
+			reg.ImageRemovalCheck{
```
listx, line 341: You should instead create a helper function that returns the correct ImageRemovalCheck{} structure. Something like `func mkRealImageRemovalCheck(...) reg.ImageRemovalCheck { ... }`. The reason is because we should not just do things ad-hoc if we can avoid it.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,86 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
listx, line 2: 2020

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,86 @@
...
+	"os"
+	"strings"
+
+	gogit "gopkg.in/src-d/go-git.v4"
+)
+
+// ImageRemovalCheck implements the PreCheck interface and checks against
+// pull requests that attempt to remove any images from the promoter manifests.
+type ImageRemovalCheck struct {
+	ManifestRepo string
```
listx, line 31: Probably `GitRepoURL` is better. Even better if it's not a `string` and actualy a URL type (see https://golang.org/pkg/net/url/#URL).

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,86 @@
...
+	MasterSHA    string
```
listx, line 32: A string is fine here but since you're using gogit, isn't there some SHA type you can use? I just found https://godoc.org/gopkg.in/src-d/go-git.v4/plumbing#Hash

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// Run executes ImageRemovalCheck on a set of promotion edges.
+func (check ImageRemovalCheck) Run(
+	edges map[PromotionEdge]interface{},
+) error {
+	var err error
+	// Clone the manifest dir into a temporary directory
```
listx, line 40: This is not necessary --- you already have the manifest repo cloned for you by Prow.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,86 @@
...
+	_, err = gogit.PlainClone(tempDir, false, &gogit.CloneOptions{
+		URL: check.ManifestRepo,
+	})
+	if err != nil {
+		return fmt.Errorf("Could not clone Git repo: %v",
+			err)
+	}
+
+	// Parse the manifests and generate the promotion edges
+	mfests, _ := ParseThinManifestsFromDir(tempDir)
```
listx, line 55: Ouch, you should always check for errors and exit early.

You should almost never ignore errors, unless you know that ignoring them is harmless because some external system is already taking care of such errors.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,86 @@
...
+	masterEdges, _ := ToPromotionEdges(mfests)
+
+	destinationImages := make(map[ImageTag]interface{})
```
listx, line 58: Can you comment what you are doing here? Esp. in the for-loop just below this?

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+// ImageRemovalCheck implements the PreCheck interface and checks against
+// pull requests that attempt to remove any images from the promoter manifests.
+type ImageRemovalCheck struct {
+	ManifestRepo string
+	MasterSHA    string
+}
+
+// Run executes ImageRemovalCheck on a set of promotion edges.
```
listx, line 35: You should re-iterate here what you are actually checking.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,86 @@
...
+	// Parse the manifests and generate the promotion edges
+	mfests, _ := ParseThinManifestsFromDir(tempDir)
+	masterEdges, _ := ToPromotionEdges(mfests)
+
+	destinationImages := make(map[ImageTag]interface{})
+	for edge := range edges {
+		destinationImages[edge.DstImageTag] = nil
+	}
+
+	// Check that every destination image in the master set of promotion edges
```
listx, line 63: Let's not call it "master set" because that sounds like something else.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,86 @@
...
+	removedImages := make([]string, 0)
+	for edge := range masterEdges {
+		_, found := destinationImages[edge.DstImageTag]
+		if !found {
+			removedImages = append(removedImages,
+				string(edge.DstImageTag.ImageName))
+		}
+	}
+
+	// Remove the temporary directory
```
listx, line 74: As stated above, since you don't need to clone, you don't need this either. In general though, you would have probably used a `defer` here to do this instead of calling this directly in-between the removal check logic.

lib/dockerregistry/checks.go
yodahekinsew: The function I use GitRepoURL for (gogit.PlainOpen) actually expects a string directory path

listx: Even so, it's better to not use strings for any internal data representation. You can easily convert a URL to a string if a library function expects a string, but it does not mean that we should also store strings for ourselves.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,133 @@
...
+type ImageRemovalCheck struct {
+	GitRepoURL     string
+	MasterSHA      plumbing.Hash
+	PullRequestSHA plumbing.Hash
+}
+
+// MKRealImageRemovalCheck returns an instance of ImageRemovalCheck.
+func MKRealImageRemovalCheck(gitRepoURL string) *ImageRemovalCheck {
+	masterSHAStr := os.Getenv("PULL_BASE_SHA")
+	pullRequestSHAStr := os.Getenv("PULL_PULL_SHA")
```
listx, line 39: What happens if this is an empty var? Fail early if possible...

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,133 @@
...
+	"strings"
+
+	gogit "gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+)
+
+// ImageRemovalCheck implements the PreCheck interface and checks against
+// pull requests that attempt to remove any images from the promoter manifests.
+type ImageRemovalCheck struct {
+	GitRepoURL     string
```
listx, line 31: Since you're opening an already-existing repo on disk, you should rename this variable...

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,133 @@
...
+		Hash:  check.PullRequestSHA,
+		Force: true,
+	})
+	if err != nil {
+		return fmt.Errorf("Could not checkout the pull request branch of the"+
+			" Git repo %v: %v",
+			check.GitRepoURL, err)
+	}
+
+	return check.CompareEdges(edges, masterEdges)
```
listx, line 100: I would prefer to flip the order of arguments, and also drop `Edges` from the name. So the call should look like `check.Compare(masterEdges, edges)`.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,133 @@
...
+	// Check that every destination image in the master branch's set of
+	// promotion edges exists in the pull request's set of promotion edges
+	removedImages := make([]string, 0)
+	for edge := range edgesMaster {
+		_, found := destinationImages[edge.DstImageTag]
+		if !found {
+			removedImages = append(removedImages,
+				string(edge.DstImageTag.ImageName))
+		}
+	}
```
listx, line 126: This check alone is not enough. Consider the case where we want to promote

(master edges) sha256:000 -> ImageTagX
(PR edges) sha256:111 -> ImageTagX

Now even though the PR is trying to promote a *different* digest, it is not detected by this check. That is, we actually *did* remove an image, with digest sha256:000 but this check did not detect it because it only looks at the ImageTag info and not the digests.

lib/dockerregistry/inventory_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Kubernetes Authors.
+Copyright 2020 The Kubernetes Authors.
```
listx, line 2: Hmm, you don't have to update this. Instead you should update the year for your `checks.go`.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,159 @@
...
+	"strings"
+
+	gogit "gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+)
+
+// ImageRemovalCheck implements the PreCheck interface and checks against
+// pull requests that attempt to remove any images from the promoter manifests.
+type ImageRemovalCheck struct {
+	CurrentGitRepo url.URL
```
listx, line 32: I don't think this will be a URL, because again, Prow gives it to you on the filesystem.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,159 @@
...
+
+// MKRealImageRemovalCheck returns an instance of ImageRemovalCheck.
+func MKRealImageRemovalCheck(
+	currentGitRepoStr string,
+) (*ImageRemovalCheck, error) {
+	currentGitRepo, err := url.Parse(currentGitRepoStr)
+	if err != nil {
+		return nil, fmt.Errorf("Could not parse the git repo url: %v", err)
+	}
+	masterSHAStr := os.Getenv("PULL_BASE_SHA")
```
listx, line 45: I would go 1 step farther and actually verify that this value looks like a SHA.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,159 @@
...
+	if masterSHAStr == "" {
+		return nil, fmt.Errorf("The PULL_BASE_SHA environment variable " +
+			"doesn't exist")
```
listx, line 48: WRT "doesn't exist" --- this is not always true. It could be that the environment variable is set to the 0-length (empty) string. Easier to just say that it is invalid.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,159 @@
...
+		return fmt.Errorf("Could not parse manifests from the directory: %v",
+			err)
+	}
+	masterEdges, err := ToPromotionEdges(mfests)
+	if err != nil {
+		return fmt.Errorf("Could not generate promotion edges from promoter"+
+			" manifests: %v", err)
+	}
+
+	// Reset the working Git repo by checking out the pull request
```
listx, line 104: Please clarify *why* you are doing this. Why is this necessary? (You know the answer now, but your fellow coders in the future won't necessarily know the context.)

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// Run executes ImageRemovalCheck on a set of promotion edges.
+// Returns an error if the pull request removes images from the
+// promoter manifests.
+func (check *ImageRemovalCheck) Run(
+	edges map[PromotionEdge]interface{},
+) error {
+	var err error
+	// Get the Git repo the current working directory
```
listx, line 71: It would be great to summarize all of the Git operations and what you are doing in a short summary at the top of this area; the interspersed comments are fine, but it's hard to see why this function is doing the things it's doing. That is, focus on the forest, not so much the trees...

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// Compare is a function of the ImageRemovalCheck that handles
+// the comparison of the pull requests's set of promotion edges and
+// the master branch's set of promotion edges.
+func (check *ImageRemovalCheck) Compare(
+	edgesMaster map[PromotionEdge]interface{},
+	edgesPullRequest map[PromotionEdge]interface{},
+) error {
+	type ImageInfo struct {
```
listx, line 125: This shouldn't be necessary... why can't we just compare edges directly?

In general you should be very conservative about introducing new data structures, even temporary ones. Imagine now testing for `ImageInfo` --- one disadvantage (among others) is you can't test it unless you move it out of this function.

lib/dockerregistry/checks_test.go
```diff
@@ -0,0 +1,170 @@
...
+					Images: []reg.Image{
+						imageB,
+					},
+					SrcRegistry: &srcRC},
+			},
+			fmt.Errorf("The following images were removed in this pull " +
+				"request: a"),
+		},
+		{
+			"Promoting Same Image From Different Registry",
```
listx, line 118: No need to capitalize every word here and in the test below.

lib/dockerregistry/inventory_test.go
yodahekinsew: yeah, this was a mistake. I was intending to update the year for the checks_test.go

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,169 @@
...
+// ImageRemovalCheck implements the PreCheck interface and checks against
+// pull requests that attempt to remove any images from the promoter manifests.
+type ImageRemovalCheck struct {
+	GitRepoPath    string
+	MasterSHA      plumbing.Hash
+	PullRequestSHA plumbing.Hash
+}
+
+func isValidSha(potenitalSHA string) error {
+	sha256Length := 64
```
listx, line 38: This should probably be a const; wdyt?

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,169 @@
...
+	if len(potenitalSHA) == sha256Length {
```
listx, line 39: Typically, the idiom I've seen in imperative languages (Python, Java, also Go) is to avoid nesting if-conditions too much. This is the style you've already used in the Run() function with its various Git ops and error-checking for each result.

E.g.
```
if a {
  return err
}

if b {
  return err
}

return nil
```

Can you apply it here?

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,169 @@
...
+
+// ImageRemovalCheck implements the PreCheck interface and checks against
+// pull requests that attempt to remove any images from the promoter manifests.
+type ImageRemovalCheck struct {
+	GitRepoPath    string
+	MasterSHA      plumbing.Hash
+	PullRequestSHA plumbing.Hash
+}
+
+func isValidSha(potenitalSHA string) error {
```
listx, line 37: Are there any hash validation functions in git-go? If not, this function should be named `isValidSha256` because that's the only one you check for.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	return fmt.Errorf("Length of SHA is %d characters, should be 64",
+		len(potenitalSHA))
+}
+
+// MKRealImageRemovalCheck returns an instance of ImageRemovalCheck.
+func MKRealImageRemovalCheck(
+	gitRepoPath string,
+) (*ImageRemovalCheck, error) {
+	masterSHAStr := os.Getenv("PULL_BASE_SHA")
```
listx, line 54: So there are 2 types of SHAs in our problem domain.

One is the digests for the images. These are sha256. The other is a commit SHA, which is a 40-character string, used by Git. You're checking for Git SHAs to have a 64-char string here (wrong SHA type).

FYI there are many kinds of SHA algorithms. See the wikipedia article for an overview of SHA-1, SHA-2, SHA-3...

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,169 @@
...
+	err := isValidSha(masterSHAStr)
```
listx, line 55: I would combine the os.Getenv and isValidateSha functions into 1. It's more DRY that way.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// Run executes ImageRemovalCheck on a set of promotion edges.
+// Returns an error if the pull request removes images from the
+// promoter manifests.
+func (check *ImageRemovalCheck) Run(
+	edges map[PromotionEdge]interface{},
+) error {
+	var err error
+	// Open the Git repo of the current working directory, check
```
listx, line 82: The first sentence is the most important, and it doesn't explain the *why*, so I think you should remove it.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,169 @@
...
+	// edges from the manifests, and then reset the working directory
+	// by checking out the pull request branch of the Git repo. This
+	// is done to make sure that this check is not leaving any lasting
+	// effects on the working directory when the check finishes.
+	r, err := gogit.PlainOpen(check.GitRepoPath)
+	if err != nil {
+		return fmt.Errorf("Could not open the Git repo: %v", err)
+	}
+
+	// Check out the master branch of the repo
```
listx, line 93: This comment is a little too low-level to be useful. Instead try to think one level higher (explain what you're doing here to someone unfamiliar with the codebase). When in doubt, try to explain *why* something is done; if that's difficult to express, then next best thing is to explain a high-level overview of why some piece of code *exists* at all. What's its *purpose*? Another way to think about it is --- if you were to move this section of code to its own function, what would be the best name for the function?

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,169 @@
...
+	err = w.Checkout(&gogit.CheckoutOptions{
+		Hash:  check.MasterSHA,
+		Force: true,
+	})
+	if err != nil {
+		return fmt.Errorf("Could not checkout the master branch of the Git"+
+			" repo: %v", err)
+	}
+
+	// Parse the manifests and generate the promotion edges
```
listx, line 107: This comment is also too low-level to be useful.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,169 @@
...
+// Compare is a function of the ImageRemovalCheck that handles
+// the comparison of the pull requests's set of promotion edges and
+// the master branch's set of promotion edges.
+func (check *ImageRemovalCheck) Compare(
+	edgesMaster map[PromotionEdge]interface{},
+	edgesPullRequest map[PromotionEdge]interface{},
+) error {
+	// Generate a set of all destination images that appear in
+	// the pull request's set of promotion edges
+	destinationImages := make(map[PromotionEdge]interface{})
```
listx, line 142: This is better, but I wonder if it's cleaner to just clear the unused parts in the edgesMaster and edgesPullRequest directly... any pros/cons you can think of?

lib/dockerregistry/checks.go
yodahekinsew: This is a really good trick!

lib/dockerregistry/types.go
```diff
@@ -65,7 +65,7 @@ type SyncContext struct {
 // Run runs the defined check on a set of PromotionEdges and returns an error
```
listx, line 65: Update comment.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,168 @@
...
+		edges,
+	}, nil
+}
+
+// Run executes ImageRemovalCheck on a set of promotion edges.
+// Returns an error if the pull request removes images from the
+// promoter manifests.
+func (check *ImageRemovalCheck) Run() error {
+	var err error
+	// To compare the master and pull request promoter manifests, check
```
listx, line 81: Ping? I commented here but it looks like my old comment is gone. Basically, state *why* in comments, not so much the *how*. The *what*-type of comments here is the least useful.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,168 @@
...
+	"gopkg.in/src-d/go-git.v4/plumbing"
+)
+
+// ImageRemovalCheck implements the PreCheck interface and checks against
+// pull requests that attempt to remove any images from the promoter manifests.
+type ImageRemovalCheck struct {
+	GitRepoPath    string
+	MasterSHA      plumbing.Hash
+	PullRequestSHA plumbing.Hash
+	PromotionEdges map[PromotionEdge]interface{}
```
listx, line 35: PrEdges is probably better (it's the edges that the PR is trying to go for).

yodahekinsew: Done

lib/dockerregistry/checks.go
yodahekinsew: done

lib/dockerregistry/types.go
yodahekinsew: done

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// MKRealImageRemovalCheck returns an instance of ImageRemovalCheck.
+func MKRealImageRemovalCheck(
+	gitRepoPath string,
+	edges map[PromotionEdge]interface{},
+) (*ImageRemovalCheck, error) {
+	// The "PULL_BASE_SHA" and "PULL_PULL_SHA" environment variables are given
+	// by the PROW job running the promoter container and represent the Git SHAs
+	// for the master branch and the pull request branch respectively
```
listx, line 60: Missing period.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,166 @@
...
+			err)
+	}
+	masterEdges, err := ToPromotionEdges(mfests)
+	if err != nil {
+		return fmt.Errorf("Could not generate promotion edges from promoter"+
+			" manifests: %v", err)
+	}
+
+	// Reset the current directory back to the pull request branch so that this
+	// check doesn't leave lasting effects that could affect subsequent checks
```
listx, line 116: Missing period.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,166 @@
...
+
+// Compare is a function of the ImageRemovalCheck that handles
+// the comparison of the pull requests's set of promotion edges and
+// the master branch's set of promotion edges.
+func (check *ImageRemovalCheck) Compare(
+	edgesMaster map[PromotionEdge]interface{},
+	edgesPullRequest map[PromotionEdge]interface{},
+) error {
+	// Generate a set of all destination images that appear in
+	// the pull request's set of promotion edges
```
listx, line 138: Missing period.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,166 @@
...
+	destinationImages := make(map[PromotionEdge]interface{})
+	for edge := range edgesPullRequest {
+		destinationImages[PromotionEdge{
+			DstImageTag: edge.DstImageTag,
+			Digest:      edge.Digest,
+		}] = nil
+	}
+
+	// Check that every destination image in the master branch's set of
+	// promotion edges exists in the pull request's set of promotion edges
```
listx, line 148: Missing period.

lib/dockerregistry/checks.go
```diff
@@ -0,0 +1,166 @@
...
+		return fmt.Errorf("Could not open the Git repo: %v", err)
+	}
+	w, err := r.Worktree()
+	if err != nil {
+		return fmt.Errorf("Could not create Git worktree: %v", err)
+	}
+
+	// The Prow job that this check is running in has already cloned the
+	// git repo for us so we can just checkout the master branch to get the
+	// master branch's version of the promoter manifests
```
listx, line 94: Missing period.

yodahekinsew: done

lib/dockerregistry/checks.go
yodahekinsew: done

lib/dockerregistry/checks.go
yodahekinsew: done

lib/dockerregistry/checks.go
yodahekinsew: done

lib/dockerregistry/checks.go
yodahekinsew: done

listx:
Thanks!

/lgtm
/approve

PR #227, opened 2020-06-24
WIP: Adding Image size check

yodahekinsew:
We need to add an image size check to make sure that the promoter isn't promoting unnecessarily large images, since larger images can pose a security risk and are generally bad practice.

### DIFF
--- a/cip.go
+++ b/cip.go
@@ -123,10 +123,10 @@ func main() {
 		"audit-gcp-project-id",
 		os.Getenv("CIP_AUDIT_GCP_PROJECT_ID"),
 		"GCP project ID (name); used for labeling error reporting logs to GCP")
-	maxImageSizePtr := flag.Float64(
+	maxImageSizePtr := flag.Int(
 		"max-image-size",
-		2,
-		"The maximum image size in GB allowed for promotion (defaults to 2GB)")
+		2048,
+		"The maximum image size (MB) allowed for promotion. Default is 2048MB")
 	flag.Parse()
 
 	if len(os.Args) == 1 {
--- a/cip.go
+++ b/cip.go
@@ -341,7 +341,12 @@ func main() {
 
 	// Check the pull request
 	if *dryRunPtr {
-		err = sc.RunChecks(promotionEdges, []reg.PreCheck{})
+		err = sc.RunChecks(promotionEdges, []reg.PreCheck{
+			reg.MKRealImageSizeCheck(
+				*maxImageSizePtr,
+				sc.DigestImageSize,
+			),
+		})
 		if err != nil {
 			klog.Exitln(err)
 		}
--- a/cip.go
+++ b/cip.go
@@ -126,7 +126,7 @@ func main() {
 	maxImageSizePtr := flag.Int(
 		"max-image-size",
 		2048,
-		"The maximum image size (MB) allowed for promotion. Default is 2048MB")
+		"The maximum image size (MiB) allowed for promotion. Default is 2048MiB")
 	flag.Parse()
 
 	if len(os.Args) == 1 {
--- a/cip.go
+++ b/cip.go
@@ -126,7 +126,10 @@ func main() {
 	maxImageSizePtr := flag.Int(
 		"max-image-size",
 		2048,
-		"The maximum image size (MiB) allowed for promotion. Default is 2048MiB")
+		"The maximum image size (MiB) allowed for promotion and must be a positive value (othwerise set to the default value of 2048 MiB)")
+	if *maxImageSizePtr <= 0 {
+		*maxImageSizePtr = 2048
+	}
 	flag.Parse()
 
 	if len(os.Args) == 1 {
@@ -341,9 +344,10 @@ func main() {
 
 	// Check the pull request
 	if *dryRunPtr {
-		err = sc.RunChecks(promotionEdges, []reg.PreCheck{
+		err = sc.RunChecks([]reg.PreCheck{
 			reg.MKRealImageSizeCheck(
 				*maxImageSizePtr,
+				promotionEdges,
 				sc.DigestImageSize,
 			),
 		})
--- a/cip.go
+++ b/cip.go
@@ -123,6 +123,10 @@ func main() {
 		"audit-gcp-project-id",
 		os.Getenv("CIP_AUDIT_GCP_PROJECT_ID"),
 		"GCP project ID (name); used for labeling error reporting logs to GCP")
+	maxImageSizePtr := flag.Float64(
+		"max-image-size",
+		2,
+		"The maximum image size in GB allowed for promotion (defaults to 2GB)")
 	flag.Parse()
 
 	if len(os.Args) == 1 {
--- a/lib/dockerregistry/inventory.go
+++ b/lib/dockerregistry/inventory.go
@@ -1984,19 +1984,19 @@ func (sc *SyncContext) RunChecks(
 func (check *ImageSizeCheck) Run(
 	edges map[PromotionEdge]interface{},
 ) error {
-	byteConvert := 1024.0
-	gbToByte := byteConvert * byteConvert * byteConvert
+	const mbToByteShift = 20
+	maxImageSizeByte := check.MaxImageSize << mbToByteShift
 	oversizedImages := make([]string, 0)
 	for edge := range edges {
 		imageSize := check.DigestImageSize[edge.Digest]
-		if imageSize > int(check.MaxImageSize*gbToByte) {
+		if imageSize > maxImageSizeByte {
 			oversizedImages = append(oversizedImages,
 				string(edge.DstImageTag.ImageName))
 		}
 	}
 	if len(oversizedImages) > 0 {
 		return fmt.Errorf("The following images were over the max file size"+
-			" of %fGB: %v",
+			" of %dMB: %v",
 			check.MaxImageSize,
 			strings.Join(oversizedImages, ", "))
 	}
@@ -2006,7 +2006,7 @@ func (check *ImageSizeCheck) Run(
 // MKRealImageSizeCheck returns an instance of ImageSizeCheck which
 // checks that all images to be promoted are under a max size.
 func MKRealImageSizeCheck(
-	maxImageSize float64,
+	maxImageSize int,
 	digestImageSize DigestImageSize,
 ) *ImageSizeCheck {
 	return &ImageSizeCheck{
--- a/lib/dockerregistry/inventory.go
+++ b/lib/dockerregistry/inventory.go
@@ -1973,6 +1973,13 @@ func (sc *SyncContext) RunChecks(
 	return nil
 }
 
+func (check *ImageSizeCheck) Run(
+	edges map[PromotionEdge]interface{},
+) error {
+	var err error
+	return err
+}
+
 // FilterPromotionEdges generates all "edges" that we want to promote.
 func (sc *SyncContext) FilterPromotionEdges(
 	edges map[PromotionEdge]interface{},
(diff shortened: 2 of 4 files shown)


cip.go
```diff
@@ -123,6 +123,10 @@ func main() {
 		"audit-gcp-project-id",
 		os.Getenv("CIP_AUDIT_GCP_PROJECT_ID"),
 		"GCP project ID (name); used for labeling error reporting logs to GCP")
+	maxImageSizePtr := flag.Float64(
+		"max-image-size",
+		2,
```
listx, line 128: I would just use MiB instead of GiB. It is more granular with no loss of functionaliy.

lib/dockerregistry/inventory.go
```diff
@@ -1973,6 +1979,42 @@ func (sc *SyncContext) RunChecks(
 	return nil
 }
 
+// Run is a function of ImageSizeCheck and checks that all
+// images to be promoted are under the max file size.
+func (check *ImageSizeCheck) Run(
+	edges map[PromotionEdge]interface{},
+) error {
+	byteConvert := 1024.0
```
listx, line 1987: This variable should be a const, but also, its name doesn't really say much.

lib/dockerregistry/inventory_test.go
```diff
@@ -2049,6 +2049,133 @@ func TestRunChecks(t *testing.T) {
...
+	image2 := reg.Image{
+		ImageName: "bar",
+		Dmap: reg.DigestTags{
+			"sha256:000": {"0.9"}}}
+
+	var tests = []struct {
+		name       string
+		check      reg.ImageSizeCheck
+		manifests  []reg.Manifest
+		imageSizes []int
```
listx, line 2095: This is a little bit confusing if you have more than 1 image; why not use a map[Digest]int ?

lib/dockerregistry/types.go
```diff
@@ -68,6 +69,14 @@ type PreCheck interface {
 	Run(edges map[PromotionEdge]interface{}) error
 }
 
+// ImageSizeCheck implements the PreCheck interface and checks against
+// images that are larger than a size threshold (controlled by the
+// max-image-size flag).
+type ImageSizeCheck struct {
+	MaxImageSize    int
+	DigestImageSize DigestImageSize
```
listx, line 77: We already store this information inside SyncContext... why do we need it here as well?

lib/dockerregistry/inventory.go
```diff
@@ -1276,6 +1277,11 @@ func (sc *SyncContext) ReadRegistries(
 					fmt.Printf("digest %s: %s\n", digest, err)
 				}
 				sc.DigestMediaType[Digest(digest)] = mediaType
+
+				// Store ImageSize
+				if sc.DigestImageSize != nil {
```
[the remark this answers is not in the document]
yodahekinsew: Because some tests make a SyncContext without the DigestImageSize field which causes the line "sc.DigestImageSize[Digest(digest)] = int(mfestInfo.Size)" to fail

lib/dockerregistry/types.go
yodahekinsew: because the sync context isn't accessible within the check.Run() function. would it be better to just pass in the SyncContext itself?

lib/dockerregistry/inventory_test.go
```diff
@@ -2049,6 +2049,133 @@ func TestRunChecks(t *testing.T) {
...
+		manifests []reg.Manifest
+		expected  error
+	}{}
+
+	for _, test := range tests {
+		edges, _ := reg.ToPromotionEdges(test.manifests)
+		got := test.check.Run(edges)
+		err := checkEqual(got, test.expected)
+		checkError(t, err,
+			fmt.Sprintf("checkError: test: %v (Error Tracking)\n", test.name))
```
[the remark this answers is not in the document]
yodahekinsew: Yeah it is

lib/dockerregistry/inventory_test.go
```diff
@@ -2042,13 +2039,123 @@ func TestRunChecks(t *testing.T) {
...
+			[]reg.Manifest{
+				{
+					Registries: registries,
+					Images: []reg.Image{
+						image1,
+					},
+					SrcRegistry: &srcRC},
+			},
+			map[reg.Digest]int{
+				"sha256:000": .5 * 1024 * 1024,
```
listx, line 2094: Use a helper function instead of 1024 * 1024.

Or, use a constant and multiply by it (toMB).

lib/dockerregistry/inventory_test.go
```diff
@@ -2042,13 +2039,123 @@ func TestRunChecks(t *testing.T) {
...
 	for _, test := range tests {
-		got := sc.RunChecks(edges, test.checks)
+		got := sc.RunChecks(test.checks)
 		err := checkEqual(got, test.expected)
 		checkError(t, err,
 			fmt.Sprintf("checkError: test: %v (Error Tracking)\n", test.name))
 	}
 }
 
+func TestImageSizeCheck(t *testing.T) {
```
listx, line 2049: You should also add cases for negative image size, what happens when you pass in -max-image-size=0, etc.

Always consider as many edge cases as possible...!

lib/dockerregistry/inventory.go
```diff
@@ -1973,6 +1976,60 @@ func (sc *SyncContext) RunChecks(
...
+	if len(oversizedImages) > 0 {
+		sort.Strings(oversizedImages)
+		errStr += fmt.Sprintf("The following images were over the max file "+
+			"size of %dMiB: %v",
+			check.MaxImageSize,
+			strings.Join(oversizedImages, ", "))
+	}
+	if len(invalidImages) > 0 {
+		sort.Strings(invalidImages)
+		errStr += fmt.Sprintf("The following images had an invalid file size "+
```
listx, line 2008: You should just display the actual image file size found for the particular image.

lib/dockerregistry/inventory.go
```diff
@@ -1973,6 +1976,60 @@ func (sc *SyncContext) RunChecks(
...
+			oversizedImages = append(oversizedImages,
+				string(edge.DstImageTag.ImageName))
+		}
+		if imageSize <= 0 {
+			invalidImages = append(invalidImages,
+				string(edge.DstImageTag.ImageName))
+		}
+	}
+
+	errStr := ""
```
listx, line 1998: You can just use fmt.Errorf() directly and return it when you use it, instead of appending into this errStr.

lib/dockerregistry/inventory.go
```diff
@@ -1973,6 +1976,60 @@ func (sc *SyncContext) RunChecks(
...
+	if len(oversizedImages) > 0 {
+		sort.Strings(oversizedImages)
+		errStr += fmt.Sprintf("The following images were over the max file "+
```
listx, line 2001: Better to display the actual image size for the offending image (probably in bytes but also in MiB).

lib/dockerregistry/inventory.go
```diff
@@ -1973,6 +1976,60 @@ func (sc *SyncContext) RunChecks(
...
+			check.MaxImageSize,
+			strings.Join(oversizedImages, ", "))
+	}
+	if len(invalidImages) > 0 {
+		sort.Strings(invalidImages)
+		errStr += fmt.Sprintf("The following images had an invalid file size "+
+			"of 0 bytes or less: %v",
+			strings.Join(invalidImages, ", "))
+	}
+	if errStr != "" {
```
listx, line 2012: This branch is not necessary if you don't use errStr.

lib/dockerregistry/inventory_test.go
```diff
@@ -2042,13 +2039,187 @@ func TestRunChecks(t *testing.T) {
...
+	image1 := reg.Image{
+		ImageName: "foo",
+		Dmap: reg.DigestTags{
+			"sha256:000": {"0.9"}}}
+	image2 := reg.Image{
+		ImageName: "bar",
+		Dmap: reg.DigestTags{
+			"sha256:111": {"0.9"}}}
+
+	const mbToBytes = 1024 * 1024
```
listx, line 2072: Another way to do this is to do a bitwise left-shift. But this is fine.

lib/dockerregistry/inventory_test.go
```diff
@@ -2042,13 +2039,187 @@ func TestRunChecks(t *testing.T) {
...
+			},
+			map[reg.Digest]int{
+				"sha256:000": 5 * mbToBytes,
+				"sha256:111": 10 * mbToBytes,
+			},
+			fmt.Errorf("The following images were over the max file size" +
+				" of 1MiB: bar, foo"),
+		},
+		{
+			"Max image size is zero",
```
listx, line 2143: I don't think it's reasonable to allow a max image size of 0... This is a case where we have to guard against misbehaving humans who put in nonsense values.

lib/dockerregistry/inventory_test.go
```diff
@@ -2042,13 +2039,187 @@ func TestRunChecks(t *testing.T) {
...
+			map[reg.Digest]int{
+				"sha256:000": 5 * mbToBytes,
+			},
+			fmt.Errorf("The following images were over the max file size" +
+				" of 0MiB: foo"),
+		},
+		{
+			"Max image size is negative",
+			reg.ImageSizeCheck{
+				MaxImageSize:    -10,
```
listx, line 2165: Same here, we should probably guard our image size flag value so that only positive integers are accepted. Then we don't have to test it down the line here during the check itself.

lib/dockerregistry/inventory_test.go
```diff
@@ -2042,13 +2039,187 @@ func TestRunChecks(t *testing.T) {
...
+					Registries: registries,
+					Images: []reg.Image{
+						image1,
+						image2,
+					},
+					SrcRegistry: &srcRC},
+			},
+			map[reg.Digest]int{
+				"sha256:000": 0,
+				"sha256:111": -5 * mbToBytes,
```
listx, line 2199: Great, this is what I was thinking of as edge cases.

lib/dockerregistry/inventory.go
```diff
@@ -1953,13 +1957,12 @@ func MKPopulateRequestsForPromotionEdges(
 
 // RunChecks runs defined PreChecks in order to check the promotion.
 func (sc *SyncContext) RunChecks(
-	edges map[PromotionEdge]interface{},
```
listx, line 1956: I didn't check directly, but can you make sure that these diffs that are not related to your image size check are in a separate commit?

lib/dockerregistry/inventory_test.go
```diff
@@ -1993,23 +1995,18 @@ func TestCheckOverlappingEdges(t *testing.T) {
 
 type FakeCheckAlwaysSucceed struct{}
 
-func (c *FakeCheckAlwaysSucceed) Run(
```
listx, line 1996: This stuff around edges should also be in a separate commit (I didn't check, can you verify?). I wonder if you rebased too far back into your history?

lib/dockerregistry/inventory.go
yodahekinsew: I'm using an errStr so that I can bubble up errors for both oversized images and invalid images. if I use fmt.Errorf() directly how would I bubble up the two errors into one

lib/dockerregistry/inventory.go
listx: Ping?

lib/dockerregistry/inventory_test.go
listx: Ping? This change is also in your other PR for image removal. You should only have it in 1 of them.

yodahekinsew:
Moved to PR #233

Issue #225, opened 2020-06-19
add "--timeout" option to pull-cip-lint job 


yodahekinsew:
The pull-cip-lint job will randomly fail on some pull requests due to timeout. Issue should be fixed by simply adding the "--timeout" option to the job with an arbitrary amount of time (something like 2 seconds).

listx:
We invoke the pull-cip-lint job here: https://github.com/kubernetes/test-infra/blob/157a58118f01633f69d1623a09a49fb39f771d79/config/jobs/kubernetes/sig-release/cip/container-image-promoter.yaml#L122 with just "make lint-ci". Can you add the flag to our Makefile?

yodahekinsew:
fixed with pull request #229

PR #229, opened 2020-06-24
added --timeout option to make lint

yodahekinsew:
The pull-cip-lint job sometimes fails on pull requests due to timeout issues. In order to fix this, the "--timeout" flag has been added to ensure that the job runs for long enough to finish. Fixing issue #225.


yodahekinsew:
\cc @listx @prathak07

PR #231, opened 2020-06-25
increased timeout of golangci-lint and added verbosity flag

yodahekinsew:
Despite adding the --timeout flag to the make lint target, the pull-cip-lint job is still failing due to timeouts. Increased the timeout of the linter from 2 seconds to 5 minutes, and added a verbosity flag in order to see why failures might be occurring.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,9 @@ image-load-cip-auditor-e2e:
 image-push-cip-auditor-e2e:
 	bazel run //test-e2e/cip-auditor:push-cip-auditor-test
 lint:
-	GO111MODULE=on golangci-lint --timeout=2s run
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -22,7 +22,7 @@ image-load-cip-auditor-e2e:
 image-push-cip-auditor-e2e:
 	bazel run //test-e2e/cip-auditor:push-cip-auditor-test
 lint:
-	GO111MODULE=on golangci-lint --timeout=2s run
+	GO111MODULE=on golangci-lint run -v --timeout=5m
```
listx, line 25: Can you move each flag to its own line with `\`? I don't like it when we have 1-line diffs and more than 1 thing is changed in it.

PR #233, opened 2020-06-30
Add ImageSizeCheck to the promoter

yodahekinsew:
We need to add an image size check to make sure that the promoter isn't promoting unnecessarily large images, since larger images can pose a security risk and are generally bad practice. Moved from PR #227.

### DIFF
--- a/lib/dockerregistry/checks.go
+++ b/lib/dockerregistry/checks.go
@@ -27,6 +27,18 @@ import (
 	"gopkg.in/src-d/go-git.v4/plumbing"
 )
 
+// MBToBytes converts a value from MiB to Bytes.
+func MBToBytes(value int) int {
+	const mbToBytesShift = 20
+	return value << mbToBytesShift
+}
+
+// BytesToMB converts a value from Bytes to MiB.
+func BytesToMB(value int) int {
+	const bytesToMBShift = 20
+	return value >> bytesToMBShift
+}
+
 func getGitShaFromEnv(envVar string) (plumbing.Hash, error) {
 	potenitalSHA := os.Getenv(envVar)
 	const gitShaLength = 40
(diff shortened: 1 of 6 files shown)


lib/dockerregistry/checks.go
```diff
@@ -164,3 +156,73 @@ func (check *ImageRemovalCheck) Compare(
...
+	imageSizes map[string]int,
+) string {
+	imageSizesStr := ""
+	imageNames := make([]string, 0)
+	for k := range imageSizes {
+		imageNames = append(imageNames, k)
+	}
+	const mbToByteShift = 20
+	sort.Strings(imageNames)
+	for i, imageName := range imageNames {
```
listx, line 184: It would be nicer to consume this data if each image was on its own separate line (assuming that humans will the main consumers of this message).

lib/dockerregistry/checks_test.go
```diff
@@ -168,3 +168,137 @@ func TestImageRemovalCheck(t *testing.T) {
...
+						image1,
+						image2,
+					},
+					SrcRegistry: &srcRC},
+			},
+			map[reg.Digest]int{
+				"sha256:000": 0,
+				"sha256:111": -5 << mbToByteShift,
+			},
+			fmt.Errorf("The following images had an invalid file size of 0" +
```
listx, line 284: It might be worth creating a custom error type. Here you're checking for the string representation of the error, which can get a bit unwieldy during comparisons because of how strings can be formatted. If you use a custom error type then you can compare the custom structs directly against each other.

lib/dockerregistry/checks.go
```diff
@@ -164,3 +156,73 @@ func (check *ImageRemovalCheck) Compare(
...
+
+func (check *ImageSizeCheck) joinImageSizesToString(
+	imageSizes map[string]int,
+) string {
+	imageSizesStr := ""
+	imageNames := make([]string, 0)
+	for k := range imageSizes {
+		imageNames = append(imageNames, k)
+	}
+	const mbToByteShift = 20
```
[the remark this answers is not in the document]
yodahekinsew: Done, the functions have been added in checks.go

lib/dockerregistry/checks.go
yodahekinsew: Done

lib/dockerregistry/checks_test.go
yodahekinsew: Done. I've added an ImageSizeError with an Error() function in checks.go

listx:
Thanks!

/lgtm
/approve

Issue #235, opened 2020-06-30
add a check against PRs that attempt add images with known vulnerabilities 


yodahekinsew:
Currently, none of the promoter’s checks check against known vulnerabilities that may exist in an image at the time of promotion. This means that the promoter may be pushing vulnerable images to production. The promoter should have a very high standard for the images that it pushes to production, since images that are pushed to production are never removed. So a vulnerability check needs to be added that makes sure new images are stable before being promoted.
(The design for this vulnerability check is written out [here](https://docs.google.com/document/d/1o50At7WoS2YUqvkYGUkVvtRsFzLdgC7kPvl9t2p7Ytg/edit#heading=h.oxbgr73uqhru).

- [x] Enable the Container Scanning API (and subsequently the Container Analysis API) for all image staging projects
- [x] Create a new service account and authenticate it with the "Container Analysis Occurrences Viewer" role for all staging projects
- [x] Create a new Kubernetes service account for the vulnerability check to authenticate with using Workload Identity
- [ ] Create new Prow job (pull-cip-vuln) to run the ImageVulnerabilityCheck (add ServiceAccountName field and --use-service-account flag) 
- [x] Add Container Analysis Service client library to the promoter
- [x] Add ImageVulnerabilityCheck implementation of the PreCheck interface
- [x] Using the ExecRequests framework, query vulnerability info for each new image to be promoted from CAS
- [x] Add unit tests for the ImageVulnerabilityCheck using stubbing

PR #236, opened 2020-07-07
WIP: add a vulnerability check for the promoter

yodahekinsew:
Adding a check against PRs that attempt to add images with known vulnerabilities (Issue #235). The vulnerability detection is handled using the Container Analysis Service.

### DIFF
--- a/cip.go
+++ b/cip.go
@@ -134,6 +134,10 @@ func main() {
 	if *maxImageSizePtr <= 0 {
 		*maxImageSizePtr = 2048
 	}
+	severityThresholdPtr := flag.Int(
+		"vuln-severity-threshold",
+		-1,
+		"Using this flag will cause the promoter to only run the vulnerability check. Found vulnerabilities at or above this threshold will result in the vulnerability check failing [severity levels beteen 0 and 5; 0 - UNSPECIFIED, 1 - MINIMAL, 2 - LOW, 3 - MEDIUM, 4 - HIGH, 5 - CRITICAL]")
 	flag.Parse()
 
 	if len(os.Args) == 1 {
@@ -280,7 +284,9 @@ func main() {
 		// Print version to make Prow logs more self-explanatory.
 		printVersion()
 
-		if *dryRunPtr {
+		if *severityThresholdPtr >= 0 {
+			klog.Info("********** START (VULN CHECK) **********")
+		} else if *dryRunPtr {
 			klog.Info("********** START (DRY RUN) **********")
 		} else {
 			klog.Info("********** START **********")
@@ -383,12 +389,25 @@ func main() {
 	if !ok {
 		klog.Exitln("encountered errors during edge filtering")
 	}
-	err = sc.Promote(promotionEdges, mkProducer, nil)
-	if err != nil {
-		klog.Exitln(err)
+
+	if *severityThresholdPtr >= 0 {
+		err = sc.RunChecks([]reg.PreCheck{
+			reg.MKImageVulnCheck(sc, promotionEdges,
+				*severityThresholdPtr, nil),
+		})
+		if err != nil {
+			klog.Exitln(err)
+		}
+	} else {
+		err = sc.Promote(promotionEdges, mkProducer, nil)
+		if err != nil {
+			klog.Exitln(err)
+		}
 	}
 
-	if *dryRunPtr {
+	if *severityThresholdPtr >= 0 {
+		klog.Info("********** FINISHED (VULN CHECK) **********")
+	} else if *dryRunPtr {
 		klog.Info("********** FINISHED (DRY RUN) **********")
 	} else {
 		klog.Info("********** FINISHED **********")
--- a/lib/dockerregistry/checks.go
+++ b/lib/dockerregistry/checks.go
@@ -17,14 +17,22 @@ limitations under the License.
 package inventory
 
 import (
+	"context"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"os"
+	"path"
 	"sort"
 	"strings"
+	"sync"
 
+	containeranalysis "cloud.google.com/go/containeranalysis/apiv1"
+	"google.golang.org/api/iterator"
+	grafeaspb "google.golang.org/genproto/googleapis/grafeas/v1"
 	gogit "gopkg.in/src-d/go-git.v4"
 	"gopkg.in/src-d/go-git.v4/plumbing"
+	"sigs.k8s.io/k8s-container-image-promoter/lib/stream"
 )
 
 // MBToBytes converts a value from MiB to Bytes.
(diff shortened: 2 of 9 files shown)


lib/dockerregistry/checks_test.go
```diff
@@ -317,3 +320,60 @@ func TestImageSizeCheck(t *testing.T) {
 			fmt.Sprintf("checkError: test: %v (ImageSizeCheck)\n", test.name))
 	}
 }
+
+func TestImageVulnCheck(t *testing.T) {
+	mkPopulateRequestsFake := func(
+		vulnerabilities []*grafeaspb.VulnerabilityOccurrence,
```
listx, line 326: You should probably use a mapping instead of a flat list. That is, something like `map[Digest]*grafeaspb.VulnerabilityOccurrence`.

lib/dockerregistry/checks.go
```diff
@@ -245,3 +251,121 @@ func (check *ImageSizeCheck) Run() error {
...
+	return false
+}
+
+// GetVulnOccurrences gets all vulnerability Occurrences associated with the
+// image represented in the PromotionEdge.
+func GetVulnOccurrences(
+	edge PromotionEdge,
+) ([]*grafeaspb.Occurrence, error) {
+	// resourceURL := fmt.Sprintf("https://gcr.io/my-project/my-image")
+	resourceURL := string(edge.SrcRegistry.Name) +
```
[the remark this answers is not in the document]
yodahekinsew: done

lib/dockerregistry/checks.go
```diff
@@ -245,3 +251,121 @@ func (check *ImageSizeCheck) Run() error {
...
+	}
+	return false
+}
+
+// GetVulnOccurrences gets all vulnerability Occurrences associated with the
+// image represented in the PromotionEdge.
+func GetVulnOccurrences(
+	edge PromotionEdge,
+) ([]*grafeaspb.Occurrence, error) {
+	// resourceURL := fmt.Sprintf("https://gcr.io/my-project/my-image")
```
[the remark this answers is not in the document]
yodahekinsew: done

lib/dockerregistry/checks_test.go
yodahekinsew: I still think using Digest's wouldn't add anything to the tests. Each unit test is just running through the vulnerabilities for one image.

lib/dockerregistry/checks.go
```diff
@@ -245,3 +252,130 @@ func (check *ImageSizeCheck) Run() error {
...
+	newPullEdges map[PromotionEdge]interface{},
+	severityThreshold int,
+) *ImageVulnCheck {
+	var populateRequestsReal PopulateRequests = func(
+		sc *SyncContext,
+		reqs chan<- stream.ExternalRequest,
+		wg *sync.WaitGroup) {
+		for edge := range newPullEdges {
+			var req stream.ExternalRequest
+			req.RequestParams = edge
```
listx, line 269: This is missing the `req.StreamProducer = ...` pattern seen elsewhere in the codebase. For example, for ReadRegistries(), we use this pattern so that in TestReadRegistries(), we test the *same* ProcessRequest function in both the real and the fake version. This way, we can test more of the code in ReadRegistries().

The way you have it here, you test slightly different things in the real vs the fake, because your `processRequestReal` is not tested.

In other words, only the portion of the code that gets the occurrences (how occurrences are retrieved) should be faked. This is because in the real world, we have to retrieve info from another service (occurrences API). For the fake, we only fake this part (instead of talking to the API to get the occurrences, we just use the occurrences hardcoded in the test cases). Everything else should be the same, for both the real vs the fake.

yodahekinsew: yeah, that makes sense. I implemented this by just making a new ImageVulnProducer type

lib/dockerregistry/checks_test.go
```diff
@@ -317,3 +322,106 @@ func TestImageSizeCheck(t *testing.T) {
 			fmt.Sprintf("checkError: test: %v (ImageSizeCheck)\n", test.name))
 	}
 }
+
+func TestImageVulnCheck(t *testing.T) {
+	mkPopulateRequestsFake := func(
```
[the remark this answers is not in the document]
yodahekinsew: yup, comments have been added

lib/dockerregistry/checks.go
```diff
@@ -245,3 +252,130 @@ func (check *ImageSizeCheck) Run() error {
...
+
+// GetVulnOccurrences gets all vulnerability Occurrences associated with the
+// image represented in the PromotionEdge.
+func GetVulnOccurrences(
+	edge PromotionEdge,
+) ([]*grafeaspb.Occurrence, error) {
+	// resourceURL is of the form https://gcr.io/[projectID]/my-image
+	resourceURL := "https://" + path.Join(string(edge.SrcRegistry.Name),
+		string(edge.SrcImageTag.ImageName))
+	projectID := strings.Split(string(edge.SrcRegistry.Name), "/")[1]
```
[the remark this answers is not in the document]
yodahekinsew: done

cip.go
```diff
@@ -134,6 +134,14 @@ func main() {
 	if *maxImageSizePtr <= 0 {
 		*maxImageSizePtr = 2048
 	}
+	checkVulnPtr := flag.Bool(
+		"check-image-vuln",
+		false,
+		"Only perform the vulnerability check for the images to be promoted")
```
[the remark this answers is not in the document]
yodahekinsew: done

cip.go
```diff
@@ -134,6 +134,14 @@ func main() {
...
+	severityThresholdPtr := flag.Int(
+		"vuln-severity-threshold",
+		2,
+		"Found vulnerabilities at or above this threshold will result in the vulnerability check failing [severity levels beteen 0 and 5]")
```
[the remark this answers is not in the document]
yodahekinsew: done

cip.go
```diff
@@ -383,12 +389,24 @@ func main() {
 	if !ok {
 		klog.Exitln("encountered errors during edge filtering")
 	}
-	err = sc.Promote(promotionEdges, mkProducer, nil)
-	if err != nil {
-		klog.Exitln(err)
+
+	if *severityThresholdPtr != -1 {
```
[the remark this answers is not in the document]
yodahekinsew: Done

lib/dockerregistry/checks.go
```diff
@@ -245,3 +252,170 @@ func (check *ImageSizeCheck) Run() error {
...
+				SrcRegistry: edge.SrcRegistry,
+				SrcImageTag: edge.SrcImageTag,
+			}
+			// Only check the vulnerability for the source image if it
+			// hasn't been checked already since multiple promotion
+			// edges can contain the same source image
+			if _, found := check.PullEdges[srcImage]; found {
+				continue
+			}
+			check.PullEdges[srcImage] = nil
```
listx, line 292: You're mutating the check.PullEdges data structure while looping over it --- instead you should use a separate map.

lib/dockerregistry/checks.go
```diff
@@ -245,3 +252,170 @@ func (check *ImageSizeCheck) Run() error {
...
+		ctx := context.Background()
+
+		req := &grafeaspb.ListOccurrencesRequest{
+			Parent: fmt.Sprintf("projects/%s", projectID),
+			Filter: fmt.Sprintf("resourceUrl = %q kind = %q",
+				resourceURL, "VULNERABILITY"),
+		}
+
+		var occurrenceList []*grafeaspb.VulnerabilityOccurrence
+		it := client.GetGrafeasClient().ListOccurrences(ctx, req)
```
listx, line 407: You should probably run GetGrafeasClient() on the outside, in block 303 somewhere. No need to call GetGrafeasClient() for every request, I think.

lib/dockerregistry/checks_test.go
```diff
@@ -317,3 +318,113 @@ func TestImageSizeCheck(t *testing.T) {
...
 }
+
+// TestImageVulnCheck uses a fake populateRequests function and a fake
+// vulnerability producer. The fake vulnerability producer simply returns the
+// vulnerability occurrences that have been mapped to a given PromotionEdge in
+// order to simulate running the real check without having to make an api call
+// to the Container Analysis Service.
+func TestImageVulnCheck(t *testing.T) {
+	edge1 := reg.PromotionEdge{
+		Digest: "foo",
```
listx, line 329: Can you use `sha256:000` and `sha256:111` instead?

Please try to comply with existing convention in the codebase!

lib/dockerregistry/checks_test.go
```diff
@@ -317,3 +318,113 @@ func TestImageSizeCheck(t *testing.T) {
...
+						Severity: grafeaspb.Severity_HIGH,
+					},
+				},
+				"bar": {
+					{
+						Severity: grafeaspb.Severity_HIGH,
+					},
+				},
+			},
+			fmt.Errorf("Encountered an error while executing requests"),
```
listx, line 392: Is there any way to make these errors less vague?

I see that this error message is coming from ExecRequests... I think we can improve this without touching ExecRequests.

lib/dockerregistry/types.go
```diff
@@ -50,6 +51,20 @@ type ImageSizeError struct {
 	InvalidImages   map[string]int
 }
 
+// ImageVulnError contains ImageVulnCheck information on images that contain a
+// vulnerability with a severity level at or above the defined threshold.
+type ImageVulnError struct {
+	ImageName             ImageName
```
listx, line 57: This should also include the Digest, because a single ImageName is by definition ambiguous.

lib/dockerregistry/checks.go
yodahekinsew: yeah good call

lib/dockerregistry/checks.go
yodahekinsew: GetGrafeasClient() just gets a field variable that already exists in the client (similar to calling client.grafeasClient), no extra logic. Is there any need to create another variable to hold the field variable?

lib/dockerregistry/checks.go
```diff
@@ -245,3 +252,170 @@ func (check *ImageSizeCheck) Run() error {
...
 }
+
+// MKImageVulnCheck returns an instance of ImageVulnCheck which
+// checks against images that have known vulnerabilities.
+// nolint[funlen]
+func MKImageVulnCheck(
+	syncContext SyncContext,
+	newPullEdges map[PromotionEdge]interface{},
+	severityThreshold int,
+	customVulnProducer ImageVulnProducer,
```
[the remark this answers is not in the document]
yodahekinsew: what does this comment mean?

lib/dockerregistry/checks_test.go
yodahekinsew: Done

lib/dockerregistry/types.go
yodahekinsew: Done

lib/dockerregistry/checks_test.go
yodahekinsew: Done

yodahekinsew: Improving this error message actually makes it easier to test some edge cases. For example, making sure that if two promotion edges have the same src image, only one of those edges is actually tested for vulnerabilities

lib/dockerregistry/checks.go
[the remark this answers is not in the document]
listx: Please rename `customVulnProducer` with `imageVulnProducer`.

lib/dockerregistry/checks.go
listx: I was just thinking that you'd pass around not the client but the result of client.GetGrafeasClient(). No need to pass around 2 variables if you are only going to be using 1 of them.

But as we discussed offline, it's fine to keep this as-is, it's probably not a big deal. Up to you.

lib/dockerregistry/checks_test.go
```diff
@@ -317,3 +318,153 @@ func TestImageSizeCheck(t *testing.T) {
...
+				"sha256:111": {
+					{
+						Severity: grafeaspb.Severity_HIGH,
+					},
+					{
+						Severity: grafeaspb.Severity_CRITICAL,
+					},
+				},
+			},
+			fmt.Errorf("VulnerabilityCheck: The following vulnerable images were found: bar (2), foo (1)"),
```
listx, line 437: Can you make this error report each image and its digest (ideally, the FQIN)? And also show the vulnerability information as well (the entire thing we get from the API)?

In general, you never want to render a list of information on a single line. Imagine if there are 40 images with errors. You will force users to read a very long single line string. Not to mention, that would be a parsing nightmare if you wanted to use the output of this test to feed that into some automation downstream.

yodahekinsew: The ImageVulnError type should be verbose enough (it contains all of the vulnerability data) and is the error that will be written to the Prow job logs and what the users will ultimately see. The error that is returned from Run() is just information I'm using for testing. I updated it to have the FQIN but I don't see the benefit of putting the entire vulnerability json return from Container Analysis into the error if I'm just using it for testing. The FQIN on the other hand is actually important for testing purposes.

listx: That's fine, I just want to make sure that we are actually showing all the information that we are dealing with at *some* point in the process. If it doesn't surface here in this particular case, but does get surfaced elsewhere, that works too.

lib/dockerregistry/checks.go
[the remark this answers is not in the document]
listx: I think we went over this. Better to include the substring`fake` in there, right?

lib/dockerregistry/checks.go
```diff
@@ -245,3 +253,195 @@ func (check *ImageSizeCheck) Run() error {
...
+		reqs chan<- stream.ExternalRequest,
+		wg *sync.WaitGroup) {
+		srcImages := make(map[PromotionEdge]interface{})
+		for edge := range check.PullEdges {
+			srcImage := PromotionEdge{
+				SrcRegistry: edge.SrcRegistry,
+				SrcImageTag: edge.SrcImageTag,
+				Digest:      edge.Digest,
+			}
+			// Only check the vulnerability for the source image if it
```
listx, line 289: You should only be checking for the digest field. There can be multiple sources for the same digest (aka "redundant" promotions as discussed in your design doc IIRC).

lib/dockerregistry/checks.go
```diff
@@ -245,3 +253,195 @@ func (check *ImageSizeCheck) Run() error {
...
+								edge.Digest,
+								occ,
+							},
+						})
+						severeOccurrences++
+					}
+				}
+				if severeOccurrences > 0 {
+					vulnerableImages = append(vulnerableImages,
+						fmt.Sprintf("%v@%v [%v severe vulnerabilitie(s), %v total]",
```
listx, line 353: s/vulnerabilitie(s)/vulnerabilities

lib/dockerregistry/checks.go
yodahekinsew: done

lib/dockerregistry/checks.go
yodahekinsew: done

lib/dockerregistry/checks.go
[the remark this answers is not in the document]
yodahekinsew: Done

lib/dockerregistry/checks.go
```diff
@@ -245,3 +253,193 @@ func (check *ImageSizeCheck) Run() error {
...
+func (err ImageVulnError) Error() string {
+	vulnJSON, _ := json.MarshalIndent(err.Vulnerability, "", "  ")
+	return fmt.Sprintf("The image %v@%v contains the following "+
+		"vulnerability:\n%v",
+		err.ImageName, err.Digest, string(vulnJSON))
+}
+
+// CheckSeverity checks if a vulnerability is a high enough severity to fail
+// the ImageVulnCheck.
+func CheckSeverity(
```
listx, line 386: Because this function reduces down to a boolean check, I think a better name is`SevereOccurrence`. That way you can use language like `if SevereOccurrence(...) { ... }` in the codebase. WDYT?

lib/dockerregistry/checks.go
```diff
@@ -245,3 +253,193 @@ func (check *ImageSizeCheck) Run() error {
...
+			reqRes := RequestResult{Context: req}
+			errors := make(Errors, 0)
+			edge := req.RequestParams.(PromotionEdge)
+			occurrences, err := vulnProducer(edge)
+			if err != nil {
+				errors = append(errors, Error{
+					Context: "error getting vulnerabilities",
+					Error:   err})
+			}
+			if len(occurrences) > 0 {
```
listx, line 334: I don't think you need this boolean check, because the `for` loop will only run if occurrences > 0. Right?

lib/dockerregistry/checks.go
yodahekinsew: maybe IsSevereOccurrence so I can say if IsSevereOccurrence(...)

yodahekinsew: Done

lib/dockerregistry/checks.go
yodahekinsew: Done

listx:
Thanks!

/lgtm
/approve

Issue #237, opened 2020-07-13
drain WaitGroup in ExecRequests after processing the request 


yodahekinsew:
Currently, ExecRequests is meant to return an error if any of the requests it processes return an error. However, there is a bug where ExecRequests returns too early. ExecRequests returns after a WaitGroup has been drained (set to 0), but the WaitGroup is currently being drained before each request is fully finished being processed. This means that sometimes ExecRequests returns before finishing all requests. 

The error can simply be fixed by draining the WaitGroup whenever a request is fully finished being processed.

listx:
Fixed in #238.
/close

PR #238, opened 2020-07-13
moved WaitGroup drain to after request results processed

yodahekinsew:
In order to fix the current bug in ExecRequests (Issue #237), we can simply move the WaitGroup draining logic to after the request results are processed. So from now on, the WaitGroup is drained at the end of the ProcessRequest function if and only if the request results channel is not being filled. Also, for the logic to still work, the request results channel needs to be filled at the end of the function.

### DIFF
--- a/lib/dockerregistry/inventory.go
+++ b/lib/dockerregistry/inventory.go
@@ -1258,7 +1258,6 @@ func (sc *SyncContext) ReadRegistries(
 						Context: "getRegistryTagsWrapper",
 						Error:   err}}
 				requestResults <- reqRes
-				wg.Add(-1)
 
 				// Invalidate promotion conservatively for the subset of images
 				// that touch this network request. If we have trouble reading
@@ -1318,9 +1317,6 @@ func (sc *SyncContext) ReadRegistries(
 				mutex.Unlock()
 			}
 
-			reqRes.Errors = Errors{}
-			requestResults <- reqRes
-
 			// Process child repos.
 			if recurse {
 				for _, childRepoName := range tagsStruct.Children {
@@ -1353,7 +1349,8 @@ func (sc *SyncContext) ReadRegistries(
 			}
 			// When we're done processing this node (req), decrement the
 			// semaphore.
-			wg.Add(-1)
+			reqRes.Errors = Errors{}
+			requestResults <- reqRes
 		}
 	}
 	sc.ExecRequests(populateRequests, processRequest)
@@ -1433,8 +1430,6 @@ func (sc *SyncContext) ReadGCRManifestLists(
 						Context: "getGCRManifestListWrapper",
 						Error:   err}}
 				requestResults <- reqRes
-				wg.Add(-1)
-
 				continue
 			}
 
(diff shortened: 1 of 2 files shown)


listx:
Can you have a look at the test failures?

lib/dockerregistry/inventory.go
```diff
@@ -2503,18 +2495,11 @@ func (sc *SyncContext) ClearRepository(
 		for req := range reqs {
 			reqRes := RequestResult{Context: req}
 			jsons, errors := getJSONSFromProcess(req)
-			if len(errors) > 0 {
-				reqRes.Errors = errors
-				requestResults <- reqRes
-				// Don't skip over this request, because stderr here is ignorable.
-				// continue
-			}
```
yodahekinsew, line 2511: In this portion of code, we are filling the requestResults channel with the same error twice. Now that the WaitGroup drain logic (_wg.Add(-1)_) has been moved to the requestsResults processor, this means that the WaitGroup was getting drained twice for each request. So this code block had to be removed in order to fix issues with the e2e tests where the WaitGroup was at a negative state.

lib/dockerregistry/inventory.go
```diff
@@ -1353,7 +1349,8 @@ func (sc *SyncContext) ReadRegistries(
 			}
 			// When we're done processing this node (req), decrement the
 			// semaphore.
-			wg.Add(-1)
+			reqRes.Errors = Errors{}
```
listx, line 1352: Do we absolutely have to move this block from line 1321?

lib/dockerregistry/inventory.go
listx: I think I get what you're saying. Here's a rewording of it: This commit moves the semaphore reduction line to ExecRequests. If we do not remove this particular block of code, then it can be the case that we push 2x into requestResults, *for the same request*. This meant WaitGroup would subtract "1" 2x even though we only added "1" 1x (negative state). Did I get it right?

yodahekinsew: Yup, that's exactly right!

lib/dockerregistry/inventory.go
yodahekinsew: Yes, we do. This is because in that recurse code block (line 1321 in this pr) we are creating new requests and pushing them into the requests channel. However, you can imagine the scenario where we push a result into the requestsResults channel which results in the WaitGroup being drained to 0. As soon as the semaphore is drained, both the requests channel and the requestResults channel are closed. This causes a panic error because the recurse block is trying to push requests into a channel that has just been closed.

listx: Ah, that's right. I think the much simpler explanation is, because we moved the semaphore decrement to ExecRequests, we need to move the push into the requestResults channel to where the old semaphore decrement in *this* function *used to be*, which is the location here at line 1352 (clearly shown in the diff). Otherwise if we leave it as-is, then it's as if we are doing `wg.Add(1)` back up at line 1321, which is clearly wrong.

Thanks!

listx:
One final nit: Can you more or less append the description of this PR into your commit message body?

listx:
Thanks!

/lgtm
/approve

PR #239, opened 2020-07-13
Upgrade rules_docker so that it works with the latest bazel

justinsb:


listx:
Looks like the e2e test failed because it couldn't download the new rules?

```
WARNING: Download from https://github.com/bazelbuild/rules_pkg/releases/download/0.2.6/rules_pkg-0.2.6.tar.gz failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 404 Not Found
```

I can't get it either when I try to curl it. 🤷

justinsb:
Thanks for checking @listx ... looks like this is because of https://github.com/bazelbuild/rules_pkg/issues/189  ... I'll fix!

justinsb:
/retest

I'm wondering if the failure comes when I run tests concurrently...

justinsb:
I'm guessing rules_tar or rules_docker changed the shas of the golden images

listx:
@justinsb You are correct! Closing this in favor of #250. Can you LGTM there?

PR #241, opened 2020-07-13
Disable new golangci-lint linters

yodahekinsew:
golangci-lint v1.28.0 and further have added multiple new linters such as gofumpt that are extremely strict and have been causing linter issues for our pull requests. Also, only v1.27.0 is available on Mac. Let's disable these new linters until they are needed in the future.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -55,11 +55,15 @@ linters-settings:
 linters:
   enable-all: true
   disable:
+  - exhaustive
+  - exportloopref
   - gochecknoglobals
   - godox
   - goerr113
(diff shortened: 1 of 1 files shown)


.golangci.yml
```diff
@@ -63,7 +63,13 @@ linters:
   - stylecheck
   - whitespace
   - wsl
-
+  - Noctx
```
listx, line 66: Can you sort these lines?

yodahekinsew: Done

listx:
Thanks!

/lgtm
/approve

PR #242, opened 2020-07-15
MkRequestsCapturer: do not manually decrement semaphore

listx:
As the semaphore decrementing duty has been simplified to be strictly
within ExecRequests() [1], the decrementing logic here must be replaced with
an empty RequestResult{} being pushed into the requestResults channel.

Also, fix a typo and rename `errs` to `requestResults` to be in line
with the rest of the codebase.

[1]: https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/238

### DIFF
--- a/lib/dockerregistry/inventory.go
+++ b/lib/dockerregistry/inventory.go
@@ -2092,7 +2092,7 @@ func getRegistriesToRead(
 	return rcsFinal
 }
 
-// Promote perferms container image promotion by realizing the intent in the
+// Promote performs container image promotion by realizing the intent in the
 // Manifest.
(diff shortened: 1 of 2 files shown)


yodahekinsew:
Should we also get rid of the manual decrementing logic in inventory_test.go? Specifically in TestGarbageCollection and TestGarbageCollectionMulti

listx:
> Should we also get rid of the manual decrementing logic in inventory_test.go? Specifically in TestGarbageCollection and TestGarbageCollectionMulti

Good catch. Will add it to this PR.

PR #244, opened 2020-07-19
Allow file manifest export to filter by prefix

justinsb:
This is useful when we want to separate files by version.


listx:
Thanks Justin!

/lgtm
/approve

PR #245, opened 2020-07-27
modify the vulnerability to check against fixable vulnerabilities

yodahekinsew:
Modifying the vulnerability check so that it only rejects a pull request if it finds vulnerabilities for an image that are BOTH too severe and have a fix available. All vulnerabilities will still be logged (regardless if they have a fix or not).


listx:
Thanks!

/lgtm
/approve

PR #248, opened 2020-07-30
fixed issue where vulnerabilities were only getting logged when a fix…

yodahekinsew:
Fixing the issue where vulnerabilities were only getting logged when a fixable vulnerability was found. The current behavior after this fix is that the vulnerability check will fail if and only if a severe and fixable vulnerability is found; but the vulnerability check will always log all of the vulnerabilities it finds. Additionally, if a fix is not available for an issue, we now clear out some of the fields that would incorrectly imply to the user that there was a fix when logged.

### DIFF
--- a/cip.go
+++ b/cip.go
@@ -286,6 +286,13 @@ func main() {
 
 		if *severityThresholdPtr >= 0 {
 			klog.Info("********** START (VULN CHECK) **********")
+			klog.Info("DISCLAIMER: Vulnerabilities are found as issues with " +
+				"package binaries within image layers, not necessarily " +
+				"with the image layers themselves. So a \"fixable\" " +
+				"vulnerability may not necessarily be immediately" +
+				"actionable. For example, even though a fixed version " +
+				"of the binary is available, it doesn't necessarily mean " +
+				"that a new version of the image layer is available.")
 		} else if *dryRunPtr {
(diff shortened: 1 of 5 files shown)


cip.go
```diff
@@ -286,6 +286,10 @@ func main() {
 
 		if *severityThresholdPtr >= 0 {
 			klog.Info("********** START (VULN CHECK) **********")
+			klog.Info("DISCLAIMER: Vulnerabilities are found as issues with " +
+				"package binaries within image dependencies, not necessarily " +
+				"with the image dependencies themselves. So a \"fixable\" " +
```
listx, line 291: "For example, even though a fixed version of the binary is available, it does not necessarily mean that a new version of the image layer is available."

yodahekinsew: Done

cip.go
```diff
@@ -286,6 +286,10 @@ func main() {
 
 		if *severityThresholdPtr >= 0 {
 			klog.Info("********** START (VULN CHECK) **********")
+			klog.Info("DISCLAIMER: Vulnerabilities are found as issues with " +
+				"package binaries within image dependencies, not necessarily " +
```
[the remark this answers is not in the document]
yodahekinsew: Done

listx:
/lgtm
/approve

/hold

Remove hold when ready.

Issue #253, opened 2020-08-11
add a vulnerability dashboard that surfaces all production image vulnerabilities


yodahekinsew:
Continuing on the vulnerability check that was made, a vulnerability dashboard should be created that will surface all the image vulnerabilities for images that have been promoted. The vulnerability check is a good way of instantly surfacing vulnerabilities to users at the instant they promote an image, but it would be useful to have a platform that continuously surfaces these issues long after the promotion has finished.

A simple design for achieving this dashboard using a static page and GCS:
- [x] Setup a GCS bucket to host the vulnerability dashboard static html page
- [x] Setup a service account that has access to both Container Analysis Service information for production images and write access to the GCS bucket
- [x] Implement an adapter that can convert the vulnerability occurrence information stored in the Container Analysis Service into a parseable JSON
- [x] Upload an html page, json file containing vulnerability data, and js file that translates the json data into an html table to GCS
- [ ] Create a Prow job, that will run the adapter moving CAS information to the GCS json at the start of each day to update the static HTML page

listx:
Thanks for tracking this @yodahekinsew !

justaugustus:
Continued in https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/274.
/close

PR #256, opened 2020-08-18
add the adapter and files needed for the vulnerability dashboard

yodahekinsew:
This PR is meant to implement the vulnerability dashboard described in Issue #253.
I added the adapter which reads vulnerability information for the entire Kubernetes' production project ("k8s-artifacts-prod") and parses into a simple JSON containing all the necessary information. That JSON is then uploaded to the dashboard's bucket in GCS. 
I also added the HTML and JS files necessary for the dashboard, which are also uploaded to the dashboard's bucket in GCS.
This PR will eventually be run in a Prow job that acts as a cron job at the start of each day.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,7 @@ test-mac:
 		//pkg/api/files:go_default_test \
 		//lib/audit:go_default_test \
 		//lib/dockerregistry:go_default_test \
+		//dashboard/adapter:go_default_test \
 		//pkg/cmd:go_default_test
 test-ci: download
 	make build
--- a/cmd/promobot-files/BUILD.bazel
+++ b/cmd/promobot-files/BUILD.bazel
@@ -52,4 +52,3 @@ docker_push(
     name = "push-image-bundle",
     bundle = "image-bundle",
 )
-
--- a/dashboard/BUILD.bazel
+++ b/dashboard/BUILD.bazel
@@ -1,0 +1,18 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
+
+go_library(
+    name = "go_default_library",
+    srcs = ["dashboard.go"],
+    importpath = "sigs.k8s.io/k8s-container-image-promoter/dashboard",
+    visibility = ["//visibility:private"],
+    deps = [
+        "//dashboard/adapter:go_default_library",
+        "@io_k8s_klog//:go_default_library",
+    ],
+)
+
+go_binary(
+    name = "dashboard",
+    embed = [":go_default_library"],
+    visibility = ["//visibility:public"],
+)
--- a/dashboard/adapter/BUILD.bazel
+++ b/dashboard/adapter/BUILD.bazel
@@ -1,0 +1,27 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
+
+go_library(
+    name = "go_default_library",
+    srcs = [
+        "adapter.go",
+        "types.go",
+    ],
+    importpath = "sigs.k8s.io/k8s-container-image-promoter/dashboard/adapter",
+    visibility = ["//visibility:public"],
+    deps = [
+        "@com_google_cloud_go//containeranalysis/apiv1:go_default_library",
+        "@com_google_cloud_go_storage//:go_default_library",
+        "@go_googleapis//grafeas/v1:grafeas_go_proto",
+        "@org_golang_google_api//iterator:go_default_library",
+        "@org_golang_x_net//html:go_default_library",
+    ],
+)
+
+go_test(
+    name = "go_default_test",
+    srcs = ["adapter_test.go"],
+    embed = [":go_default_library"],
+    deps = [
+        "@go_googleapis//grafeas/v1:grafeas_go_proto",
+    ],
+)
--- a/dashboard/adapter/adapter.go
+++ b/dashboard/adapter/adapter.go
@@ -1,0 +1,200 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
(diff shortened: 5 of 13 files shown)


dims:
Love the idea!

lib/dashboard/adapter.go
```diff
@@ -0,0 +1,193 @@
...
+	const timeout = 60
+	ctx := context.Background()
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		return fmt.Errorf("storage.NewClient: %v", err)
+	}
+	defer client.Close()
+
+	// Open local file.
+	f, err := os.Open("/Users/yodalemu/google/myfork/k8s-container-image-promoter/" + object)
```
listx, line 49: Local file, please change.

lib/dashboard/dashboard.html
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.4.1/jquery.min.js"></script>
+    <script type="text/javascript" src="https://storage.googleapis.com/yodahe-vuln-dashboard-test/test-dashboard.js"></script>
```
listx, line 6: Replace `yodahe-...` with something more official.

lib/dashboard/dashboard.js
```diff
@@ -0,0 +1,68 @@
+const DASHBOARD_JSON = "https://storage.googleapis.com/yodahe-vuln-dashboard-test/test-dashboard.json";
```
listx, line 1: Replace `yodahe-...` with something more official.

lib/dashboard/adapter.go
```diff
@@ -0,0 +1,193 @@
...
+
+// GenerateDashboardJSON parses the a slice of vulnerability occurrences
+// into a JSON that only contains the necessary information for each
+// vulnerability. The JSON is also formatted to be easily processed into
+// an HTML table by the vulnerability dashboard.
+func GenerateDashboardJSON(
+	productionVulnerabilities []*grafeaspb.Occurrence,
+) ([]byte, error) {
+	var occByResource = make(map[string]ImageVulnOcc)
+	var resourceURIs []string
```
listx, line 120: Please explain what a resourceURI is, and give an actual example of what one might look like.

lib/dashboard/adapter.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+// GenerateDashboardJSON parses the a slice of vulnerability occurrences
+// into a JSON that only contains the necessary information for each
+// vulnerability. The JSON is also formatted to be easily processed into
+// an HTML table by the vulnerability dashboard.
+func GenerateDashboardJSON(
+	productionVulnerabilities []*grafeaspb.Occurrence,
+) ([]byte, error) {
+	var occByResource = make(map[string]ImageVulnOcc)
```
listx, line 119: What is `string` here?

lib/dashboard/adapter.go
```diff
@@ -0,0 +1,193 @@
...
+		if vuln.GetFixAvailable() {
+			imageVulnOcc.FixableVulnerabilities = append(
+				imageVulnOcc.FixableVulnerabilities,
+				vulnName,
+			)
+		}
+		occByResource[occ.ResourceUri] = imageVulnOcc
+	}
+
+	sort.Strings(resourceURIs)
```
listx, line 153: What does this section do?

listx: Maybe add a comment there that says it adds determinism?

listx: I think if you use yaml instead of json the sorting will happen automatically. Not sure, please check.

listx: https://stackoverflow.com/questions/18668652/how-to-produce-json-with-sorted-keys-in-go

lib/dashboard/dashboard.html
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
```
listx, line 1: Can you add a sanity check test here? To check that this HTML is valid.

yodahekinsew: added the sanity check in adapter.go, I basically just call html.Parse() and see if any errors get returned

lib/dashboard/adapter.go
yodahekinsew: Done

lib/dashboard/dashboard.html
yodahekinsew: Done

lib/dashboard/dashboard.js
yodahekinsew: Done

lib/dashboard/adapter.go
yodahekinsew: Done

lib/dashboard/adapter.go
yodahekinsew: yeah the JSON marshaling a map is keeping everything ordered. originally it wasn't working for me because I wanted to marshal a list for purposes of parsing in the JavaScript file; but I've since updated the JavaScript so marshaling a map should work well.

lib/dashboard/adapter.go
yodahekinsew: a resource URI. Do you think it would be better to make a separate type? For better code-readability?

listx: You could use a dedicated separate type, or just leave a comment in the code.

listx:
/lgtm
/approve
/hold

Thanks! Release hold when ready.

listx:
Thanks!

/hold cancel

Issue #266, opened 2020-10-16
[Umbrella] Repo improvements


justaugustus:
This a braindump of tasks I've done / plan on doing to improve the repo...

- [x] Code organization
  - [x] (https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/261) Init submodules, so libraries are importable
  - [x] (https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/260) Init v2 module (needs bazel cleanup first)
  - [x] (https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/262) Migrate `lib/` into `pkg/`
  - [x] (https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/263) Migrate vulnerability dashboard library into `pkg/`
- [x] OWNERS
  - [x] (https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/265) Establish Release Engineering ownership of repo
  - [x] (https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/265) Add active maintainers / prune inactive
  - [x] (https://github.com/kubernetes/k8s.io/pull/1341) Enable OWNERS to perform image promotion
  - [x] (https://github.com/kubernetes/org/pull/2272) Update SIG Release teams to include current CIP maintainers
- [x] Test Infra
  - [x] Contributing
    - [x] RelEng ownership of CI jobs
      - [x] OWNERS update in https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-release/cip/OWNERS: https://github.com/kubernetes/test-infra/pull/23598
      - [x] (https://github.com/kubernetes/test-infra/pull/19599) Inform Release Managers of CI failures
    - [x] (https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/267) Add issue/PR templates
    - [x] (https://github.com/kubernetes/test-infra/pull/19596) Add RelEng labels to repo
    - [x] (https://github.com/kubernetes/test-infra/pull/19597) Enable common prow plugins
    - [x] (https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/268, https://github.com/kubernetes/test-infra/pull/19599) Verify scripts
  - [x] Migrating jobs to K8s Infra - https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/269
[...]

PR #268, opened 2020-10-16
Add verify scripts and fix some header files

cpanato:
Add verify scripts and fix some header files

ref: https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/266

~~Will update the presubmit jobs~~
test-infra PR: https://github.com/kubernetes/test-infra/pull/19599

/cc @justaugustus @saschagrunert @listx


bin/verify_boilerplate.py
```diff
@@ -0,0 +1,261 @@
+#!/usr/bin/env python3
```
saschagrunert, line 1: We still use bazel/repo-infra inside this repository, so I think we can use their boilerplate verification.

cpanato: can we just not use bazel for the verify tasks?

wdty @justaugustus ?

cpanato: adding the bazel to make all the verify checks

cpanato: ok, now we need to update the job again to have the bazel installed, which image should I use for that?

justaugustus: I'm aiming at removing bazel in this repo, so I'd prefer the checks to be in bash/python, please. You can use the ones we have in k/release.

justaugustus:
I did some fixups and removed the bazel bits, because the presubmit is now blocking PRs.
Added the squash label as well.

/retest

justaugustus:
Oh, right:
/lgtm
/approve

bin/verify_boilerplate.py
cpanato: cool, thanks @justaugustus

cpanato: was more easy just to drop the bazel commits :)

PR #273, opened 2020-10-21
Migrate non-essential promotion CLI tools to k/release

justaugustus:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

Companion PR for https://github.com/kubernetes/release/pull/1652.
Part of https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/266.

/assign @listx @hasheddan 
cc: @justinsb @kubernetes-sigs/release-engineering 

#### Does this PR introduce a user-facing change?

```release-note
- Migrate non-essential promotion CLI tools to k/release
  The following tools have been removed from the repo:
  - `cip-mm`
  - `promobot-files`
  - `promobot-generate-manifest`
```

PR #285, opened 2020-12-01
cmd/cip: Default to non-production runs with `--confirm`

justaugustus:
#### What type of PR is this?

/kind bug regression
/priority critical-urgent

- test-e2e: Minor cleanup for cmp.Diff() output
- cmd/cip: Use `--confirm` instead of `--dry-run`

  Similar to our `--nomock` flags for other tools, we use a `--confirm`
  flag for `cip`, since the default nil value for booleans is `false`.
  
  Meaning: If `--dry-run` is not explicitly set to `true`, our tooling
  will initiate an image promotion.

- Flip confirm logic in cip/cip-auditor e2e MakeSyncContext() calls

/assign @hasheddan @xmudrii 

```release-note
- cmd/cip: Use `--confirm` instead of `--dry-run`

  Similar to our `--nomock` flags for other tools, we use a `--confirm`
  flag for `cip`, since the default nil value for booleans is `false`.
  
  Meaning: If `--dry-run` is not explicitly set to `true`, our tooling
  will initiate an image promotion.
```

### DIFF
--- a/cmd/cip/cmd/root.go
+++ b/cmd/cip/cmd/root.go
@@ -52,13 +52,6 @@ func init() {
 		"info",
 		fmt.Sprintf("the logging verbosity, either %s", log.LevelNames()),
 	)
-
-	rootCmd.PersistentFlags().BoolVar(
-		&rootOpts.DryRun,
-		"dry-run",
-		rootOpts.DryRun,
-		"test run promotion without modifying any registry",
-	)
 }
 
 func initLogging(*cobra.Command, []string) error {
--- a/cmd/cip/cmd/run.go
+++ b/cmd/cip/cmd/run.go
@@ -49,6 +49,13 @@ var runOpts = &cli.RunOptions{}
 // TODO: Function 'init' is too long (171 > 60) (funlen)
 // nolint: funlen
 func init() {
+	runCmd.PersistentFlags().BoolVar(
+		&runOpts.Confirm,
+		"confirm",
+		runOpts.Confirm,
+		"initiate a PRODUCTION image promotion",
+	)
+
 	// TODO: Move this into a default options function in pkg/promobot
 	runCmd.PersistentFlags().StringVar(
 		&runOpts.Manifest,
--- a/legacy/cli/root.go
+++ b/legacy/cli/root.go
@@ -24,7 +24,6 @@ import (
 
 type RootOptions struct {
 	LogLevel string
-	DryRun   bool
 }
 
 func printVersion() {
--- a/legacy/cli/run.go
+++ b/legacy/cli/run.go
@@ -40,7 +40,7 @@ type RunOptions struct {
 	Threads                 int
 	MaxImageSize            int
 	SeverityThreshold       int
-	DryRun                  bool
+	Confirm                 bool
 	JSONLogSummary          bool
 	ParseOnly               bool
 	MinimalSnapshot         bool
@@ -142,7 +142,7 @@ func RunPromoteCmd(opts *RunOptions) error {
 		sc, err = reg.MakeSyncContext(
 			mfests,
 			opts.Threads,
-			opts.DryRun,
+			opts.Confirm,
 			opts.UseServiceAcct,
 		)
 		if err != nil {
@@ -159,8 +159,9 @@ func RunPromoteCmd(opts *RunOptions) error {
 		sc, err = reg.MakeSyncContext(
 			mfests,
 			opts.Threads,
-			opts.DryRun,
-			opts.UseServiceAcct)
+			opts.Confirm,
+			opts.UseServiceAcct,
+		)
 		if err != nil {
 			logrus.Fatal(err)
 		}
@@ -210,10 +211,10 @@ So a 'fixable' vulnerability may not necessarily be immediately actionable. For
 example, even though a fixed version of the binary is available, it doesn't
 necessarily mean that a new version of the image layer is available.`,
 			)
-		} else if opts.DryRun {
-			logrus.Info("********** START (DRY RUN) **********")
-		} else {
+		} else if opts.Confirm {
 			logrus.Info("********** START **********")
+		} else {
+			logrus.Info("********** START (DRY RUN) **********")
 		}
 	}
 
@@ -249,7 +250,7 @@ necessarily mean that a new version of the image layer is available.`,
 			sc, err = reg.MakeSyncContext(
 				mfests,
 				opts.Threads,
-				opts.DryRun,
+				opts.Confirm,
 				opts.UseServiceAcct,
 			)
 			if err != nil {
@@ -302,7 +303,7 @@ necessarily mean that a new version of the image layer is available.`,
 	}
 
 	// Check the pull request
-	if opts.DryRun {
+	if !opts.Confirm {
 		err = sc.RunChecks([]reg.PreCheck{})
 		if err != nil {
 			return errors.Wrap(err, "running prechecks before promotion")
@@ -362,10 +363,10 @@ necessarily mean that a new version of the image layer is available.`,
 
 	if opts.SeverityThreshold >= 0 {
 		logrus.Info("********** FINISHED (VULN CHECK) **********")
-	} else if opts.DryRun {
-		logrus.Info("********** FINISHED (DRY RUN) **********")
-	} else {
+	} else if opts.Confirm {
 		logrus.Info("********** FINISHED **********")
+	} else {
+		logrus.Info("********** FINISHED (DRY RUN) **********")
 	}
 
 	return nil
--- a/legacy/dockerregistry/inventory.go
+++ b/legacy/dockerregistry/inventory.go
@@ -60,11 +60,11 @@ func GetSrcRegistry(rcs []RegistryContext) (*RegistryContext, error) {
 func MakeSyncContext(
 	mfests []Manifest,
 	threads int,
-	dryRun, useSvcAcc bool,
+	confirm, useSvcAcc bool,
 ) (SyncContext, error) {
 	sc := SyncContext{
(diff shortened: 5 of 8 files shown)


justaugustus:
(Needed to flip the flag in the e2e test as well. Fixed in https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/285/commits/6a5b04ae3e69f2bfaed0d2737547d49cdf8f2ed8.)

justaugustus:
```
 checking snapshots BEFORE promotion:
execing cmd bazel [run --workspace_status_command=/home/prow/go/src/sigs.k8s.io/k8s-container-image-promoter/workspace_status.sh //cmd/cip:cip -- run --snapshot=us.gcr.io/k8s-cip-test-prod/golden-bar]
time="2020-12-01T00:58:09Z" level=fatal msg="<<<<<<< got (type []inventory.Image)\n[{bar map[sha256:610b1ef6fec876146dee2b2846c890b566d26f235d7ea8982056a3e84bd35929:[1.0]]}]\n=======\n[]\n>>>>>>> expected (type []inventory.Image)" 
```

???

*grumbles*
/test pull-cip-e2e

justaugustus:
Local run:

```console
$ time bazel run --workspace_status_command=$(pwd)/workspace_status.sh //cmd/cip:cip -- run --snapshot=us.gcr.io/k8s-cip-test-prod/golden-bar
INFO: Analyzed target //cmd/cip:cip (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //cmd/cip:cip up-to-date:
  bazel-bin/cmd/cip/cip_/cip
INFO: Elapsed time: 2.650s, Critical Path: 2.49s
INFO: 13 processes: 13 linux-sandbox.
INFO: Build completed successfully, 14 total actions
INFO: Build completed successfully, 14 total actions
INFO Request {us.gcr.io/k8s-cip-test-prod/golden-bar   true}: OK 
INFO Request {us.gcr.io/k8s-cip-test-prod/golden-bar/bar   false}: OK 
- name: bar
  dmap:
    "sha256:610b1ef6fec876146dee2b2846c890b566d26f235d7ea8982056a3e84bd35929": ["1.0"]

real	0m4.111s
user	0m0.125s
sys	0m0.064s
```

justaugustus:
Oh hello, auditor flake?
/test pull-cip-auditor-e2e

hasheddan (approved):
/lgtm

saschagrunert (approved):

saschagrunert:
The CI issue seems reproducible to me. :thinking:

justaugustus:
Superseded by https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/291.
/close

justaugustus:
Reset this branch to be built on top of https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/291.
/hold until that merges and that content is rebased out.

justaugustus:
Bazel fetch flake for rules_python:
/test pull-cip-auditor-e2e

listx:
Looking at this again, can we keep the old `--dry-run` flag? Why can't we additionally add a `--no-dry-run` flag that simply aliases into the old `--dry-run` flag (that is, make `--no-dry-run` be the same thing as `--dry-run=false`)?

>  Meaning: If `--dry-run` is not explicitly set to `true`, our tooling will initiate an image promotion.

If we're by default trying to make the tool write (promote) into GCRs, I am against this approach. The default invocation (without setting any special flags) should be a dry run; this been the case for this tool for a long time (at least for the majority of the time that this tool has been in use) and I don't want to change semantics for folks who might not be as familiar with the tool.

In addition, I personally do not like the whole `--nomock` negative boolean semantics, and would not look to spread what I think was originally a design decision that was created in anago (the shell script). (Even if we want to follow `--nomock`-based semantics --- isn't that the same thing as what we currently do today, requiring users to provide `--dry-run=false` to actually perform a promotion rather than just dry-running by default?)

@justaugustus @saschagrunert @spiffxp  WDYT?

/hold

spiffxp:
We don't have a clearly consistent pattern across all of the tools used in this project.  While I don't like the double-negative of "no dry run false" I personally am a fan of what the flag is trying to accomplish:
- defaulting to a dry-run mode of operation for tools that mutate the state of the world via an API
- lack of explicit initialization implying false (this is how golang works for booleans, and feels closer to how bash would handle `[[ -n ${foo} ]]` or `${foo} == "true"`)

We've used `--confirm` for these semantics in a number of tools:
- https://github.com/kubernetes/k8s.io/blob/main/groups/reconcile.go#L131 - groups reconciler
- https://github.com/kubernetes/test-infra/blob/master/label_sync/main.go#L138 - label_sync
- https://github.com/kubernetes/test-infra/blob/master/prow/cmd/branchprotector/protect.go#L92 - branchprotector
- https://github.com/kubernetes/test-infra/blob/master/prow/cmd/peribolos/main.go#L88 - peribolos
- https://github.com/kubernetes/test-infra/blob/master/robots/commenter/main.go#L69 - commenter
- https://github.com/kubernetes/test-infra/blob/master/robots/pr-creator/main.go#L86 - pr-creator

There are, of course, just as many prow components that use `dry-run` and default it to `true` (hook, horologium, prow-controller-manager, sub, tide) because the confirm convention came about after a number of components had been written, and isn't consistently enforced.

The `--dry-run` and `--no-dry-run` convention feels more pythonic, but would be less easy to consistently implement across languages, and I'm not finding prior art for it in other tools on this project atm.

cs.k8s.io links for reference:
- [confirm](https://cs.k8s.io/?q=%22confirm%22&i=nope&files=&excludeFiles=&repos=)
- [dry-run](https://cs.k8s.io/?q=%22dry-run%22&i=nope&files=&excludeFiles=&repos=)
- [no-dry-run](https://cs.k8s.io/?q=%22no-dry-run%22&i=nope&files=&excludeFiles=&repos=)

tl;dr I would prefer `confirm` over `no-dry-run`, or reverting back to `dry-run (default: true)`

justaugustus:
> We don't have a clearly consistent pattern across all of the tools used in this project. While I don't like the double-negative of "no dry run false" I personally am a fan of what the flag is trying to accomplish:
> 
>     * defaulting to a dry-run mode of operation for tools that mutate the state of the world via an API
> 
>     * lack of explicit initialization implying false (this is how golang works for booleans, and feels closer to how bash would handle `[[ -n ${foo} ]]` or `${foo} == "true"`)

@spiffxp -- You nailed it here! The goal is ensure the tool defaults to dry-run and requires additional work to trigger a production promotion.

> tl;dr I would prefer `confirm` over `no-dry-run`, or reverting back to `dry-run (default: true)`

`--confirm` sounds good to me; I'll update the PR to that effect.

listx:
> > We don't have a clearly consistent pattern across all of the tools used in this project. While I don't like the double-negative of "no dry run false" I personally am a fan of what the flag is trying to accomplish:
> > ```
> > * defaulting to a dry-run mode of operation for tools that mutate the state of the world via an API
> > 
> > * lack of explicit initialization implying false (this is how golang works for booleans, and feels closer to how bash would handle `[[ -n ${foo} ]]` or `${foo} == "true"`)
> > ```
> 
> @spiffxp -- You nailed it here! The goal is ensure the tool defaults to dry-run and requires additional work to trigger a production promotion.

To be clear, the tool already defaults to `--dry-run=true`, even if you don't specify the flag.

spiffxp:
To @listx's point, I think this change on its own can come across as churn.  If there is a broader effort to standardize on this pattern across tools, then it seems like there's more value in doing this.

spiffxp:
I feel like maybe it _used_ default to true and then somewhere in the churn of https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/273 that got lost?

justaugustus:
Slack convo: https://kubernetes.slack.com/archives/CJH2GBF7Y/p1629138583178000

listx:
> I feel like maybe it _used_ default to true and then somewhere in the churn of #273 that got lost?

Yeah that seems likely. FTR every deployed instance of CIP in recent memory (including the current version v2.4.1 that's seeing presubmit/postsubmit usage) has always had `--dry-run=true` as the default.

justaugustus:
> Yeah that seems likely. FTR every deployed instance of CIP in recent memory (including the current version v2.4.1 that's seeing presubmit/postsubmit usage) has always had `--dry-run=true` as the default.

Right, we haven't changed the actual `cip run` invocations for CI jobs, but the default behavior/underlying implementation is what needs to be fixed.

justaugustus:
Just to capture some things from Slack:

This line creates a new set of options:
https://github.com/kubernetes-sigs/k8s-container-image-promoter/blob/97e351eb08d2af791f1a62ead66e5291398c4f0b/cmd/cip/cmd/root.go#L38

The options themselves: https://github.com/kubernetes-sigs/k8s-container-image-promoter/blob/97e351eb08d2af791f1a62ead66e5291398c4f0b/legacy/cli/root.go#L25-L28

If `--dry-run` is never set, it will be its `nil` value, which is `false` AKA unconditional promotion, which is what we don't want.
We'll proceed w/ `--confirm` as the flag name (to improve readability and make it a little more consistent with other tools across the community).

justaugustus:
(rebasing for handoff)
I've asked @puerco to take over this branch/work in https://kubernetes.slack.com/archives/CJH2GBF7Y/p1631539082071700.
/assign @puerco

justaugustus:
@puerco -- Nice work! 🎉 
/hold cancel

puerco:
Reviewing @justaugustus commits :)
/lgtm

justaugustus:
Symbolic LGTM for @puerco's fixes (even though the bot will yell at me):
/lgtm
/approve
/honk

PR #290, opened 2021-05-24
modify readme

tylerferrara:
This is only a test


listx:
FTR the pull-cip-e2e failure is expected and what we wanted to see. It proves that this test actually attempts to compile and run the code in this repo (specifically the e2e.go file).

PR #301, opened 2021-06-04
fix cip-postsubmit-push-to-staging error on tagged commits

listx:
Tagged commits result in the PROW_GIT_TAG resulting in the truncated
form vYYYYMMDD-tag instead of vYYYYMMDD-tag-n-ghash. This breaks the old
strict regex check and aborts the cip-postsubmit-push-to-staging Prow
job (which containerizes the latest master branch contents into a new
image and pushes it to gcr.io/k8s-staging-artifact-promoter).

Here we relax the strict regex check, and use fallback values for
git_commit and git_desc.

```release-note

```

### DIFF
--- a/workspace_status.sh
+++ b/workspace_status.sh
@@ -37,17 +37,30 @@ p_() {
 # mechanism, use it instead of asking `git` directly, because the repo we are in
 # might not actually be a Git repo.
 if [[ -n "${PROW_GIT_TAG:-}" ]]; then
-    # Ensure that the PROW_GIT_TAG fits a known pattern of vYYYYMMDD-tag-n-ghash.
-    regex="v[0-9]{8}-[^-]+-[0-9]+-g[0-9a-f]{7}"
-    if ! [[ "${PROW_GIT_TAG}" =~ $regex ]]; then
-        echo >&2 "could not extract git hash from PROW_GIT_TAG ${PROW_GIT_TAG}"
-        exit 1
+    # Ensure that the PROW_GIT_TAG fits a known pattern of
+    # vYYYYMMDD-tag-n-ghash. According to
+    # https://github.com/kubernetes/test-infra/blob/f7e21a3c18f4f4bbc7ee170675ed53e4544a0632/config/jobs/image-pushing/README.md#custom-substitutions,
+    # the PROW_GIT_TAG will be one of vYYYYMMDD-hash, vYYYYMMDD-tag, or
+    # vYYYYMMDD-tag-n-ghash.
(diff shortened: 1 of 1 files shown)


listx:
/hold

Waiting on https://github.com/kubernetes-sigs/k8s-container-image-promoter/commit/657fd4e94ccab3129063c3dc3610797a45e2df10#commitcomment-51709013

justaugustus (changes requested):
@listx -- What do you think of simplifying the code and using something close to one of the following examples:

- https://github.com/kubernetes/test-infra/blob/master/hack/print-workspace-status.sh
- https://github.com/kubernetes/kubernetes/blob/master/hack/print-workspace-status.sh

listx:
> @listx -- What do you think of simplifying the code and using something close to one of the following examples:
> 
> * https://github.com/kubernetes/test-infra/blob/master/hack/print-workspace-status.sh
> * https://github.com/kubernetes/kubernetes/blob/master/hack/print-workspace-status.sh

I don't think that's a priority right now --- I'm just trying to fix the immediate problem I witnessed in https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/cip-postsubmit-push-to-staging/1399306188939595776#1:build-log.txt%3A8917

Later we can refactor the script, but that should be in a separate PR/issue.

listx:
@justaugustus pasting your slack comments:

```
justaugustus (he/him)  9 days ago
@listx -- actually, for yours, I think I can resolve mobile...
For the workspace script, the reason I mentioned the k/k and k/test-infra examples is because they're simplified versions of what we have today and live in battle-tested repos

justaugustus (he/him)  9 days ago
We should model after that and not have multiple code paths for interpreting tag version

justaugustus (he/him)  9 days ago
For the unit test script, I'm fine with taking a few things as follow-up (assuming a issue is filed for it), but CI should always report on as many failures as possible to limit the iteration cycles locally

justaugustus (he/him)  9 days ago
Overall, the unit tests run pretty fast, so -failfast is a relatively small optimization
```

listx:
> For the workspace script, the reason I mentioned the k/k and k/test-infra examples is because they're simplified versions of what we have today and live in battle-tested repos

That's fine but I'm more interested in fixing the immediate edge case of Prow breaking if we push up a tagged commit.

> justaugustus (he/him)  9 days ago
> We should model after that and not have multiple code paths for interpreting tag version

Can we do this sort of cleanup effort in a separate PR? I believe and prefer smaller PRs that fix one thing only if possible.

justaugustus:
I'm not sure I understand fixing this in two steps, especially given there's no immediate plan for a new tag, but I can issue a follow-up once this merges.

/lgtm
/approve
/release-note-none

PR #302, opened 2021-06-08
test-go.sh: exit with an error upon test failure

listx:
Also, we break early if any one of the unit tests fail. Previously even
if a unit test failed, we would still exit with successfully.

The `-failfast' option makes go stop executing more Test* functions in
the package if a failure is encountered.

#### What type of PR is this?

/kind bug
/kind regression

NONE

### DIFF
--- a/hack/test-go.sh
+++ b/hack/test-go.sh
@@ -46,6 +46,7 @@ mkdir -p "${ARTIFACTS}"
 
 go_test_flags=(
     -v
+    -failfast
     -count=1
     -timeout="${TEST_TIMEOUT}s"
     -cover -coverprofile "${ARTIFACTS}/coverage.out"
@@ -54,6 +55,11 @@ go_test_flags=(
 packages=()
 mapfile -t packages < <(go list ./... | grep -v 'sigs.k8s.io/k8s-container-image-promoter/cmd\|test-e2e')
 
-GO111MODULE=on go test "${go_test_flags[@]}" "${packages[@]}"
+for p in "${packages[@]}"; do
+    if ! GO111MODULE=on go test "${go_test_flags[@]}" "${p}"; then
+        # Exit early.
+        exit 1
+    fi
+done
 
 go tool cover -html "${ARTIFACTS}/coverage.out" -o "${ARTIFACTS}/coverage.html"


hack/test-go.sh
```diff
@@ -54,6 +55,11 @@ go_test_flags=(
 packages=()
 mapfile -t packages < <(go list ./... | grep -v 'sigs.k8s.io/k8s-container-image-promoter/cmd\|test-e2e')
 
-GO111MODULE=on go test "${go_test_flags[@]}" "${packages[@]}"
+for p in "${packages[@]}"; do
+    if ! GO111MODULE=on go test "${go_test_flags[@]}" "${p}"; then
+        # Exit early.
+        exit 1
```
justaugustus, line 61: As this reads right now, if any suite of package tests fails, the entire script fails.
I think I'd prefer to see this encapsulated in a function, with `return 1` instead.

More below 👇🏾

hack/test-go.sh
```diff
@@ -54,6 +55,11 @@ go_test_flags=(
...
+    fi
+done
 
 go tool cover -html "${ARTIFACTS}/coverage.out" -o "${ARTIFACTS}/coverage.html"
```
justaugustus, line 65: Then maybe a counter of sorts to increment on the exit statuses of each function run, _after_ `go tool cover`.
If the number of `return 1`s is greater than zero, then `exit 1`.

justaugustus (changes requested):
@listx -- Left some comments.

One thing to consider...
If someone doesn't run the unit tests locally, that means they'd potentially only push changes with a subset of the errors fixed.

Is that a experience we want, given each push kicks off new e2e tests?

justaugustus:
> that means they'd potentially only push changes with a subset of the errors fixed

That is to say, from a personal preference, I'd prefer to fix as many/all errors, if I can, before submitting tests again.
cc: @kubernetes-sigs/release-engineering to get some takes

listx:
> @listx -- Left some comments.
> 
> One thing to consider...
> If someone doesn't run the unit tests locally, that means they'd potentially only push changes with a subset of the errors fixed.

Isn't that what CI is for? `pull-cip-unit-tests`

> 
> Is that a experience we want, given each push kicks off new e2e tests?

We could do it in 2 ways: (1) for plain `make test`, we exit on the first error, as this is meant to be run locally (2) for CI, we exit after all errors are shown. WDYT?

listx:
Also, I would prefer to make the CI vs. local invocation changes in a separate PR.

listx:
This fixes #309

listx:
Filed issue #310. @justaugustus I'd like to incorporate the changes you suggested in a separate PR to keep PR sizes down.

Does this work for you?

tylerferrara:
Is there a way to replicate this behavior?
When attempting to break the version_test.go in #321 I still get a failure.

hack/test-go.sh
justaugustus: Note that `go_test_flags` includes `-cover -coverprofile "${ARTIFACTS}/coverage.out"`.
By putting this in loop, we'll mangle the results of `go tool cover`. Now it will only include the coverage report from the last package that gets tested in the loop.

justaugustus (changes requested):
@listx -- You raised this [here on Slack](https://kubernetes.slack.com/archives/C2C40FMNF/p1624896727383600), but the initial reviews have not been addressed.

listx:
OK, so, it appears that this change is just noise. Sorry about that.

Long story short: I was fiddling with the `TestParseRegistryManifest` test in order to see how `test-go.sh` works. I modified the `expectedErr` in there to be some other string (these strings have to be unique, otherwise the test is supposed to fail). However no matter how I modified it, `test-go.sh` would still pass.

It turns out that these test semantics were changed in https://github.com/kubernetes/release/pull/1772, which explains the above scenario. I will instead open up a separate PR to restore the old (much stricter) test semantics.

Issue #304, opened 2021-06-10
Removing Bazel from e2e tests


tylerferrara:
Sub-Issue of #300 

Outlining the steps of implementing [Tarball Image Loading](https://docs.google.com/document/d/1tQTb8dqKQtsL1a5jmCDbRg2MEDMkAnT07vrrxyZxLfk/edit?usp=sharing) within CIP.
This involves accomplishing the following:
- [x]  Generate tarball archives for golden-tests #305 #307
- [x]  Add archive loading to test-e2e-cip #306  
- [x]  Replace all Bazel invocations in e2e.go with `go run` #308 
- [x]  Add archive loading to test-e2e-cip-auditor #315 #318 
- [x]  Replace all Bazel invocations in cip-auditor-e2e.go with `go run` #316 
- [x]  Cleanup tests, golden image data, and archives #320 #319 #317 

/sig release
/area release-eng

/assign @tylerferrara 
cc: @listx @amwat @justaugustus @kubernetes-sigs/release-engineering

tylerferrara:
Completed, closing.

PR #306, opened 2021-06-14
Integrate archive loading into e2e tests

tylerferrara:
#### What type of PR is this?
/kind cleanup
/kind deprecation

This change replaces the process in which golden-images are generated inside `test-e2e-cip`. As apart of #304 to remove Bazel from the e2e tests, the `bazel run` command, is being replaced with the `push-golden.sh` script. This functional replacement generates the images from archives, located in `golden-archives`, and pushes them to `gcr.io/k8s-staging-cip-test`.
None

Using the golden-archives as test images require all testing fixtures (manifest files & tests.yaml) to include these new sha256 digests. The old Bazel generated images, when loaded and pushed as archives, could not re-produce the original digests specified in the current manifest files. Due to a mismatch of architectures when pulled, the Bazel images' layers are modified automatically upon push. Therefore, we rely on these new images introduced in #307 which all have consistent digest when pulling and pushing to the staging location.

None

```release-note
Adding push-golden.sh script to load and push golden images for test-e2e-cip.
Replacing old image digests within e2e test fixtures.
```

cc: @listx @justaugustus @amwat


tylerferrara:
`pull-cip-e2e` is failing because docker is not running in the Prow image:
```Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?```
Therefore, we need to change the configuration of [this prow job](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-release/cip/container-image-promoter.yaml#L81).

Blocked by PR: [#22575](https://github.com/kubernetes/test-infra/pull/22575)

**Comment by tylerferrara** [comment]
I'm not sure the archives can produce the exact sha256 digests defined in the e2e tttest fixtures.
Running this script, I attempt to pull one image, re-tag it & push to a different repo.
```sh
docker pull gcr.io/k8s-gcr-audit-test-prod/golden-bar/bar@sha256:610b1ef6fec876146dee2b2846c890b566d26f235d7ea8982056a3e84bd35929

docker tag gcr.io/k8s-gcr-audit-test-prod/golden-bar/bar@sha256:610b1ef6fec876146dee2b2846c890b566d26f235d7ea8982056a3e84bd35929 \
    gcr.io/k8s-staging-cip-test/golden-bar/bar:1.0

docker push gcr.io/k8s-staging-cip-test/golden-bar/bar:1.0
```
Unfortunately, the new image digest will not produce the same digest it was pulled as. However, we can rely on this digest to be consistent and reproducible, it just wont match our existing sha256 test definitions. Therefore, we must also update all e2e test fixtures aswell.

tylerferrara:
This PR uses new archives from #307. Waiting on that PR merge before removing WIP.

listx:
For posterity, the existing golden e2e images that we already use (the stuff inside the test-e2e folder) can indeed be saved back out from Bazel into docker, like this:

```
bazel run --workspace_status_command=${PWD}/workspace_status.sh //test-e2e/cip:golden-images-loadable
```

which runs a Shell script (!) (named `golden-images-loadable` in `bazel-bin/`) that loads the individual Docker image layers and eventually calls `docker load -i ...`. However, these images get different digests than the ones recorded in the test fixtures. So the only way to continue to use the old images (to avoid having to search-replace all of the test promoter manifest files to match the digests of the newly-added *.tar images in #305) is to pull them back down from an existing e2e test run (which left test images around in the test prod repo after the test finished), but this seems like overkill at this point.

justaugustus:
@tylerferrara -- general note...

Prefer rebases over branch merges, so you can escape the bot warnings

```console
git remote update
git rebase origin/master
```

test-e2e/cip/push-golden.sh
```diff
@@ -0,0 +1,70 @@
...
+stagingRepo="gcr.io/k8s-staging-cip-test"
+
+# Load archives.
+docker load -i "${archivePath}/bar/1.0.tar"
+docker load -i "${archivePath}/foo/1.0-linux_amd64.tar"
+docker load -i "${archivePath}/foo/1.0-linux_s390x.tar"
+docker load -i "${archivePath}/foo/NOTAG-0.tar"
+
+# Push to k8s-staging-cip-test.
+docker push ${stagingRepo}/golden-bar/bar:1.0
```
listx, line 45: For consistency, you should quote this variable too like you did for `archivePath`.

test-e2e/cip/e2e.go
```diff
@@ -140,114 +140,27 @@ func testSetup(repoRoot string, t E2ETest) error {
...
-	for _, line := range strings.Split(strings.TrimSuffix(stdout, "\n"), "\n") {
-		if strings.Contains(line, o) {
-			words := strings.Split(line, " ")
-			// nolint[gomnd]
-			if len(words) == 2 {
-				return words[1]
-			}
-		}
-	}
+	fmt.Println(std.Output())
```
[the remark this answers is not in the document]
tylerferrara: Good point.

test-e2e/cip/push-golden.sh
```diff
@@ -0,0 +1,70 @@
...
+archivePath="${repoRoot}/test-e2e/cip/golden-archives"
+stagingRepo="gcr.io/k8s-staging-cip-test"
+
+# Load archives.
+docker load -i "${archivePath}/bar/1.0.tar"
+docker load -i "${archivePath}/foo/1.0-linux_amd64.tar"
+docker load -i "${archivePath}/foo/1.0-linux_s390x.tar"
+docker load -i "${archivePath}/foo/NOTAG-0.tar"
+
+# Push to k8s-staging-cip-test.
```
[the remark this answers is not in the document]
tylerferrara: Added a new workspace_info.sh for easy parsing.

test-e2e/cip/push-golden.sh
```diff
@@ -0,0 +1,70 @@
...
+set -o nounset
+set -o pipefail
+
+if [[ $# -ne 1 ]]
+then
+    printf "Error: Invalid number of arguments!\n\tUsage: publish-golden.sh repo-root\n"
+    exit 1
+fi
+
+repoRoot=$1
```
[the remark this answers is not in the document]
listx: For your script the invocation would be

```
$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)
```

(notice the extra `/..`) because the script is 2 directory levels below the root (unlike test-go.sh, which is only 1 directory level below the root). Does this invocation also not work?

test-e2e/cip/e2e.go
```diff
@@ -140,114 +140,27 @@ func testSetup(repoRoot string, t E2ETest) error {
...
+	cmd := command.NewWithWorkDir(
+		repoRoot,
+		"/bin/bash",
+		goldenPush,
+		repoRoot,
+	)
 
-	return nil
-}
+	fmt.Printf("executing %s\n", cmd.String())
```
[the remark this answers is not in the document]
listx: I think it's fine to leave the STDOUT and STDERR as-is without klog, because it's nice to see the output of the actual command that was executed as-is (without wrapping it inside klog).

test-e2e/cip/push-golden.sh
```diff
@@ -0,0 +1,70 @@
...
+#
+# Usage:
+#   ./push-golden.sh repo-root
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+if (( $# != 1 )); then
+    echo -e "Error: Invalid number of arguments!
```
listx, line 29: The `-e` is spurious here.

Also, use `echo >&2 ...` instead because it's generally good practice to always print to STDERR any error message.

test-e2e/cip/push-golden.sh
```diff
@@ -0,0 +1,70 @@
...
+    Usage: publish-golden.sh repo-root"
```
listx, line 30: You can do `$0` instead of hardcoding the name of the script.

test-e2e/cip/push-golden.sh
```diff
@@ -0,0 +1,70 @@
...
+    exit 1
+fi
+
+repoRoot=$1
+archivePath="${repoRoot}/test-e2e/cip/golden-archives"
+eval "$(~/k8s-container-image-promoter/workspace_info.sh)"
```
listx, line 36: You should use `source`or `.` instead of `eval`.

workspace_info.sh
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 The Kubernetes Authors.
```
listx, line 3: Update year to 2021

test-e2e/cip/push-golden.sh
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 The Kubernetes Authors.
```
listx, line 3: Update year to 2021

workspace_info.sh
```diff
@@ -0,0 +1,62 @@
...
+else
+    git_commit="$(git rev-parse HEAD)"
+    git_desc_raw="$(git describe --always --dirty --long)"
+    git_desc="${git_desc_raw//\//-}"
+    timestamp_utc_rfc3339=$(date -u +"%Y-%m-%d %H:%M:%S%z")
+    timestamp_utc_date_dashes="${timestamp_utc_rfc3339% *}"
+    timestamp_utc_date_no_dashes="${timestamp_utc_date_dashes//-/}"
+fi
+
+cat << EOF
```
listx, line 48: You can remove the cat command. Just assign the variables here as-is. That way you don't have to `eval` from the other script.

listx: > You can remove the cat command. Just assign the variables here as-is. That way you don't have to `eval` from the other script.

Actually, seeing as how k/k and k/test-infra do the same thing, this is fine as-is.

test-e2e/cip/push-golden.sh
listx: > You should use `source`or `.` instead of `eval`.

On second thought, for consistency sake let's just use `eval`. See my comment below about using/not-using `cat`.

listx: Thinking about this again, probably `source <(${repoRoot}/workspace_info.sh)` is the best way. The `<(..)` syntax is called process substitution.

test-e2e/cip/push-golden.sh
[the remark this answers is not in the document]
tylerferrara: Yes! This works when invoking push-golden.sh from any directory.
Thank you

test-e2e/cip/push-golden.sh
tylerferrara: Removed runtime argument, so this error message is removed aswell.

test-e2e/cip/push-golden.sh
tylerferrara: Acknowledged for future reference.

workspace_info.sh
tylerferrara: Changing, though this may cause a failure in boilerplate validation.

test-e2e/cip/push-golden.sh
tylerferrara: Changing.

test-e2e/cip/push-golden.sh
tylerferrara: Changing. Out of curiosity, why might this be the _better_ way to load workspace variables?

listx: See https://stackoverflow.com/questions/17529220/why-should-eval-be-avoided-in-bash-and-what-should-i-use-instead

listx:
/lgtm
/approve

@justaugustus PTAL

test-e2e/cip/push-golden.sh
```diff
@@ -0,0 +1,64 @@
...
+#
+# Usage:
+#   ./push-golden.sh repo-root
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+repoRoot=$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)
+archivePath="${repoRoot}/test-e2e/cip/golden-archives"
```
justaugustus, line 29: Style nit: prefer snakecase (`repo_root`, `archive_path`) for local variables in shell scripts.

test-e2e/cip/push-golden.sh
```diff
@@ -0,0 +1,64 @@
...
+source <(${repoRoot}/workspace_info.sh)
```
justaugustus, line 30:
```suggestion
source <(${repoRoot}/workspace_status.sh)
```

workspace_info.sh
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env bash
```
justaugustus, line 1: Can you instead restore this to `workspace_status.sh` and work on top of the existing script?
Recent changes were made in https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/301 and https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/311 and we should maintain the history.

workspace_info.sh
```diff
@@ -0,0 +1,62 @@
...
+
+# Copyright 2021 The Kubernetes Authors.
```
justaugustus, line 3:
```suggestion
# Copyright 2020 The Kubernetes Authors.
```

Related to the previous review comment: this is a modification of an existing file, so let's preserve history and the copyright date.

justaugustus (changes requested):
@tylerferrara -- Great work; just a few comments!

workspace_info.sh
tylerferrara: Removing workspace_info.sh

workspace_info.sh
tylerferrara: Sure thing! In order to keep workspace_status.sh working with Bazel builds, I added an optional parameter to allow for easier variable injection.

Usage: `workspace_status.sh [inject]`

This removes the need to have two separate workspace scripts.

test-e2e/cip/push-golden.sh
tylerferrara: Ack

Issue #312, opened 2021-06-22
`images` check results in GCB run failures despite successful image push


justaugustus:
#### What happened:

https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/311 added a check for the existence of the pushed images (`cip`, `cip-auditor`) before marking the GCB job successful.

Here are some examples:
- https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-cip-push-image-cip/1406887836669448192
- https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-cip-push-image-cip/1407220804646604800
- https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-cip-push-image-cip/1407381867107192832

From some of the output, we can see that the images are actually getting built:

```console
INFO: Build completed successfully, 37 total actions
2021/06/22 17:00:33 Destination {STABLE_IMG_REGISTRY}/{STABLE_IMG_REPOSITORY}/{STABLE_IMG_NAME}-auditor:latest was resolved to gcr.io/k8s-staging-artifact-promoter/cip-auditor:latest after stamping.
2021/06/22 17:00:33 Destination {STABLE_IMG_REGISTRY}/{STABLE_IMG_REPOSITORY}/{STABLE_IMG_NAME}-auditor:{STABLE_IMG_TAG} was resolved to gcr.io/k8s-staging-artifact-promoter/cip-auditor:v20210622-v1.337.0-16-gfdaf8db after stamping.
2021/06/22 17:00:33 Destination {STABLE_IMG_REGISTRY}/{STABLE_IMG_REPOSITORY}/{STABLE_IMG_NAME}:{STABLE_IMG_TAG} was resolved to gcr.io/k8s-staging-artifact-promoter/cip:v20210622-v1.337.0-16-gfdaf8db after stamping.
2021/06/22 17:00:33 Destination {STABLE_IMG_REGISTRY}/{STABLE_IMG_REPOSITORY}/{STABLE_IMG_NAME}:latest was resolved to gcr.io/k8s-staging-artifact-promoter/cip:latest after stamping.
2021/06/22 17:00:34 Successfully pushed Docker image to gcr.io/k8s-staging-artifact-promoter/cip-auditor:latest
2021/06/22 17:00:34 Successfully pushed Docker image to gcr.io/k8s-staging-artifact-promoter/cip:latest
[...]

PR #313, opened 2021-06-22
gcb: Temporarily disable incorrectly failing `images` check

justaugustus:
#### What type of PR is this?

/kind failing-test

ref: https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/312
cc: @tylerferrara @listx @kubernetes-sigs/release-engineering 

/hold I'm going to try and fix the optional presubmit here too
/test ls

```release-note
NONE
```

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -25,8 +25,10 @@ tags:
 - ${_GIT_TAG}
 - ${_PULL_BASE_REF}
 
-images:
-  - 'gcr.io/$PROJECT_ID/cip:$_GIT_TAG'
-  - 'gcr.io/$PROJECT_ID/cip:latest'
-  - 'gcr.io/$PROJECT_ID/cip-auditor:$_GIT_TAG'
-  - 'gcr.io/$PROJECT_ID/cip-auditor:latest'
+# TODO: Temporarily disabling this check, which is currently failing in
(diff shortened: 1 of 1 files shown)


justaugustus:
I started boiling the ocean trying to support the presubmit, but let's fix that in a follow-up:
/hold cancel
/override pull-cip-image-cip

cloudbuild.yaml
```diff
@@ -25,8 +25,10 @@ tags:
 - ${_GIT_TAG}
 - ${_PULL_BASE_REF}
 
-images:
-  - 'gcr.io/$PROJECT_ID/cip:$_GIT_TAG'
-  - 'gcr.io/$PROJECT_ID/cip:latest'
-  - 'gcr.io/$PROJECT_ID/cip-auditor:$_GIT_TAG'
-  - 'gcr.io/$PROJECT_ID/cip-auditor:latest'
+# TODO: Temporarily disabling this check, which is currently failing in
```
listx, line 28: Once we remove bazel (which is doing the image building (and pushing) *without* tagging them locally, which breaks this stanza), we should be able to reinstate this.

justaugustus: @listx -- Thanks for the context! I had a feeling it was a bazel oddity at work here.

PR #314, opened 2021-06-22
Adapt push-golden.sh for audits

tylerferrara:
#### What type of PR is this?
/kind cleanup
/kind deprecation

Modify the `push-golden.sh` location and function to accept the `--audit` flag. This changes the staging location, allowing the auditor to use the same script for pushing golden images (#315).

Partially satisfies #304 

None

This new flag is optional.
`Usage: ./test-e2e/push-golden.sh [--audit | -a]`

NONE

```release-note
Adding optional flag [--audit | -a] to push-golden.sh
```


test-e2e/push-golden.sh
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env bash
```
justaugustus, line 1: @tylerferrara -- To preserve the history of this file, let's do the following:
- One commit to `git mv` the file up a directory
- A second commit that adds the logic for the auditor pushes

test-e2e/cip-auditor/cip-auditor-e2e.go
```diff
@@ -307,90 +307,114 @@ func testSetup(
...
-		fmt.Println(std.Output())
-		fmt.Println(std.Error())
-	}
+	return err
 
-	return nil
 }
 
+// // nolint[funlen]
+// func oldPopulateGoldenImages(repoRoot, pushRepo string) error {
```
justaugustus, line 334: Debug tip: it's a little simpler to use a multi-line comment (`/* */`) to disable a function

```go
/*
  func disabledStuff() {
    // stuff
  }
*/
```

instead of single-line comments:

Debug tip: it's a little simpler to use a multi-line comment to disable a function

```go
//  func disabledStuff() {
//    // stuff
//  }
```

justaugustus: Also, no need to rename the function, since it's commented out.

justaugustus (commented):
@tylerferrara -- Quick tips!

test-e2e/push-golden.sh
tylerferrara: Absolutely! This was the original plan, though I'm still testing the script to make sure it works with the auditor. This PR will only be modifying `push-golden.sh` to keep the diff small.

test-e2e/cip-auditor/cip-auditor-e2e.go
tylerferrara: Ack. Still just testing.

test-e2e/push-golden.sh
justaugustus: @tylerferrara -- this still needs to be done as separate commits using `git mv` to preserve the history and make the changes to content clear in the diff

tylerferrara: After changing more than (the default similarity), git interprets this as a (creation and deletion) rather than a mv.

justaugustus: @tylerferrara -- This is why it needs to be done as two separate commits. It'll look like a creation/deletion in GitHub's UI, but git's tracking of the file should remain intact.

tylerferrara: Even separating the `git mv` to another commit still shows the same creation/deletion.

listx: I spoke offline with @tylerferrara. There will be 1 commit doing a pure `git mv` only and subsequent commits actually making changes to the business logic.

listx: As it stands now it appears this has been done already. Thanks @tylerferrara !

justaugustus: Looks good now. Thanks @tylerferrara!
(FYI, you'll be able to see the difference if you step through commit-by-commit; the GitHub UI does not handle moves well in the aggregated diff view.)

justaugustus (approved):
Thanks!

PR #316, opened 2021-06-23
Remove Bazel from building and executing auditor within Makefile

tylerferrara:
#### What type of PR is this?
/kind cleanup
/kind deprecation

#### What this PR does / why we need it:
This change replaces all invocations of `bazel build` with `./go_with_version.sh [build | run]` for the auditor. Bazel is now not required for building or running `cip-auditor-e2e.go`. Additionally, the auditor adopts the `internal/version` for standardized logging.

Partially satisfies #304

#### Which issue(s) this PR fixes:
None

#### Special notes for your reviewer:
The function [getBazelOption](https://github.com/kubernetes-sigs/k8s-container-image-promoter/blob/2638c8a3dea81fe8da2a85369008efe5c1791eeb/test-e2e/cip-auditor/cip-auditor-e2e.go#L327) was replaced with `getWorkspaceStatus`. This retrieves all variables within workspace_status.sh by returning a `map[string]string`. Such a change avoids printing the following output 4 times when triggering `test-e2e-cip-auditor`:
```
STABLE_GIT_COMMIT v1.337.0-23-g2638c8a-dirty
STABLE_IMG_REGISTRY gcr.io
STABLE_IMG_REPOSITORY k8s-staging-artifact-promoter
STABLE_IMG_NAME cip
STABLE_IMG_TAG v20210623-v1.337.0-23-g2638c8a-dirty
STABLE_TEST_AUDIT_PROD_IMG_REPOSITORY us.gcr.io/k8s-gcr-audit-test-prod
STABLE_TEST_AUDIT_STAGING_IMG_REPOSITORY gcr.io/k8s-gcr-audit-test-prod
STABLE_TEST_AUDIT_PROJECT_ID k8s-gcr-audit-test-prod
STABLE_TEST_AUDIT_PROJECT_NUMBER 375340694213
STABLE_TEST_AUDIT_INVOKER_SERVICE_ACCOUNT k8s-infra-gcr-promoter@k8s-gcr-audit-test-prod.iam.gserviceaccount.com
STABLE_TEST_STAGING_IMG_REPOSITORY gcr.io/k8s-staging-cip-test
```

#### Does this PR introduce a user-facing change?
NONE

```release-note
Simplify parsing of workspace_status.sh within cip-auditor-e2e.go
Remove Bazel invocations of auditor in favor of go_with_version.sh
```
cc: @listx @amwat @justaugustus @kubernetes-sigs/release-engineering

PR #319, opened 2021-06-24
Add more detailed logging statement in auditor

tylerferrara:
#### What type of PR is this?
/kind cleanup

This change fixes an unaddressed concern in #316. When finding a malformed key value pair, while parsing `workspace_status.sh`, the error will log the content of the malformed line.

None

If the auditor finds a malformed key-value pair, this would be an example of the log output:
```
F0625 00:06:56.732779    2663 cip-auditor-e2e.go:325] Error: Unexpected key value pair when parsing workspace_status.sh.
Line: "SOME_MALFORMED_KEY foo bar"
```

NONE

```release-note
Printing contents of malformed line in getWorkspaceStatus. 
```
cc: @listx @amwat @justaugustus

### DIFF
--- a/test-e2e/cip-auditor/cip-auditor-e2e.go
+++ b/test-e2e/cip-auditor/cip-auditor-e2e.go
@@ -322,7 +322,7 @@ func getWorkspaceStatus(repoRoot string) map[string]string {
 	for idx, line := range strings.Split(strings.TrimSuffix(stdout, "\n"), "\n") {
 		words := strings.Split(line, " ")
 		if len(words) != 2 {
-			klog.Fatalf("Unexpected key value pair in line: %d!\n", idx)
+			klog.Fatalf("ERROR: Unexpected key value pair when parsing workspace_status.sh. Line %d: %q", idx, line)
 		}
 		status[words[0]] = words[1]
 	}


test-e2e/cip-auditor/cip-auditor-e2e.go
```diff
@@ -319,10 +319,10 @@ func getWorkspaceStatus(repoRoot string) map[string]string {
 	}
 
 	stdout := std.Output()
-	for idx, line := range strings.Split(strings.TrimSuffix(stdout, "\n"), "\n") {
+	for _, line := range strings.Split(strings.TrimSuffix(stdout, "\n"), "\n") {
 		words := strings.Split(line, " ")
 		if len(words) != 2 {
-			klog.Fatalf("Unexpected key value pair in line: %d!\n", idx)
+			klog.Fatalf("Error: Unexpected key value pair when parsing workspace_status.sh.\nLine: \"%s\"", line)
```
listx, line 325: Not sure if the leading "Error: " prefix is required (unless we already do this elsewhere, can you check?).

Can you also include the line number (`idx`) in the log output as well?

Also, you can drop the embedded newlines. Also you should use `%q` instead. See https://golang.org/pkg/fmt/

tylerferrara: Fatal errors are not consistent. However, [here](https://github.com/kubernetes-sigs/k8s-container-image-promoter/blob/master/test-e2e/cip-auditor/cip-auditor-e2e.go#L127-#L148) in cip-auditor-e2e.go there's a pattern of using the `error` keyword at the beginning of the log. This makes it easier to find errors (with ctrl-f search) inside huge logfiles.

tylerferrara (commented):
There's not a clear convention for Fatal errors in CIP.

test-e2e/cip-auditor/cip-auditor-e2e.go
amwat: FYI It might be worth going with `ERROR:` because prow highlights these in the log
https://github.com/kubernetes/test-infra/blob/bb8a234a43a9824d8f7518b51321d826245926ff/config/prow/config.yaml#L62

amwat (commented):
/lgtm

listx:
Thanks!

/lgtm
/approve

Issue #310, opened 2021-06-19
test-go.sh: when run in CI (Prow), do not exit immediately upon the first test failure


listx:
#### What would you like to be added:

#302 fixes a longstanding bug (AFAICS this has been around for months) around test-go.sh just printing statements and not actually exiting with an error if a unit test fails. However it should sometimes exit late rather than early (that is, show all failing tests instead of exiting on the first error encountered).

#### Why is this needed:

For Prow/CI jobs that use this script, we want contributors to see all of their errors before they submit their next revision to CI (assuming that they didn't even run the tests locally).

listx:
As explained in https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/302#issuecomment-869847962, this was a false alarm. Sorry for the noise; closing.

PR #327, opened 2021-06-29
support serial unit test execution

listx:
This makes it easier to iterate against failing tests (no need to swim
across verbose outputs of passing tests to find the failing one).

#### What type of PR is this?

/kind feature

Allow devs to clearly see which test is failing without having to look at the parallelized test output of the numerous passing tests.

The motivations behind this PR is driven in part by the already-closed issue #310. The problem with that issue was that it thought that the `test-go.sh` script was broken (that `test-go.sh` would exit with a success even given a test failure). However upon further investigation this turned out to be a false alarm. The `test-go.sh` script works correctly. The remaining issue though, is that even if a test fails, the script by default runs tests in parallel across multiple packages and those tests will finish regardless of a test failure in one particular package. This makes debugging hard because developers have to search for the test failure output.

This adds a `--failfast` flag to `test-go.sh`. Developers who want to run tests serially (to make the tests stop upon the first failure) can now do so with this flag.

### DIFF
--- a/hack/test-go.sh
+++ b/hack/test-go.sh
@@ -37,6 +37,10 @@ do
         TEST_TIMEOUT="$2"
         shift
         shift
+        ;;
+        --failfast)
+        TEST_FAILFAST="-failfast"
+        shift
     esac
 done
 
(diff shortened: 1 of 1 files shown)


justaugustus (approved):
/release-note-none

Issue #328, opened 2021-06-29
Verify building with Docker


tylerferrara:
#### What would you like to be added:
Sub-issue of #339 

The process of building container images should be guarded by tests. Currently, its possible to change the Dockerfile to use any base image and include any extra code. 

When building with `cip-image.sh` we can verify:

- Base images of each container
- Size constraints
- Number of image layers (this may change based on relied upon base images)
- Running processes (`docker top <container>`)

#### Why is this needed:
The make targets `image-push-cip-auditor-e2e` and `image-push` both rely on `push-cip.sh` to effectively build images correctly. Currently, if this script is modified, or broken, pre-submits will not warn of such a failure. Adding a test for the `push-cip.sh` will ensure the reliability of both make targets. Additionally, testing the contents of the images will strengthen trust of provenance in our container builds.

/assign
cc: @listx @amwat @justaugustus @kubernetes-sigs/release-engineering

justaugustus:
Closed via https://github.com/kubernetes/test-infra/pull/22619
/close

PR #329, opened 2021-06-29
[WIP] Add test-cip-image-build make target

tylerferrara:
#### What type of PR is this?
/kind feature

This make target introduces a new script, `test-cip-image.sh`, which validates the build process for [cip-image.sh](https://github.com/kubernetes-sigs/k8s-container-image-promoter/blob/master/hack/cip-image.sh). More specifically, it ensures the expected images are built after triggering make targets: [image-build and image-build-cip-auditor-e2e](https://github.com/kubernetes-sigs/k8s-container-image-promoter/blob/master/Makefile#L35-L41). After merging, this test will be triggered for every PR by a new Prow pre-submit [#22736](https://github.com/kubernetes/test-infra/pull/22736). Until then, the new target: `test-cip-image-build` will be triggered within the `pull-cip-e2e` pre-submit.

Partially satisfies #328 
Must be merged:
- [x] #323
- [x] #331 

None

This PR must first be merged in order to create the new pre-submit.

NONE

```release-note
Create new test-push-cip make target.
```

### DIFF
--- a/hack/test-cip-image.sh
+++ b/hack/test-cip-image.sh
@@ -1,0 +1,79 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# About: This script tests the build process of the cip-image.sh script for both
+# variants (cip and auditor).
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+repo_root=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)
+
+# Ensures the provided images are built locally.
+validateImages() {
+    local images="${*}"
+    local img_id
+    # Look for every image locally.
+    for img in $images; do
+        # Obtain the image ID.
+        img_id=$(docker image ls "$img" --format '{{.ID}}')
+        # Ensure the image exists.
+        if [[ -z "$img_id" ]]; then
+            >&2 echo "ERROR: Image \"$img\" was not found locally."
+            exit 1
+        fi
+    done
+}
+
+# Validate the container build process for the given variant.
+testVariant() {
+    # Build based on the variant.
+    if [[ "$1" == "auditor" ]]; then
+        make image-build-cip-auditor-e2e
+    else
+        make image-build
+    fi
+
+    # Check required images are built locally.
+    shift
+    validateImages "${*}"
+}
+
+# Program entrypoint.
+main() {
+    # Inject workspace variables.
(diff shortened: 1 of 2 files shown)


justaugustus:
/hold
See https://github.com/kubernetes/test-infra/pull/22619.
/test ls

tylerferrara:
> /hold
> See [kubernetes/test-infra#22619](https://github.com/kubernetes/test-infra/pull/22619).
> /test ls

The prow pre-submit `pull-cip-image-cip` runs GCB with the cloudbuild.yaml configuration. This in turn calls, and tests, `make image-push`. After the merging of #323, this target triggers the cip/prod variant to be built and pushed. The auditor/test variant is triggered by the auditor e2e test [here](https://github.com/kubernetes-sigs/k8s-container-image-promoter/blob/master/test-e2e/cip-auditor/cip-auditor-e2e.go#L604).

Realistically, that's complete code coverage for the new `push-cip.sh` script. However, I would still like to go through with directly testing this script for these reasons:

1. pull-cip-image-cip is not automatically triggered for every PR. Therefore, if I completely deleted [these lines](https://github.com/kubernetes-sigs/k8s-container-image-promoter/blob/master/hack/push-cip.sh#L108-L110), and don't push the cip variant at all, pre-submits would not catch this. Although post submits would catch this, it would first require merging to catch it.
2. The auditor e2e only tests a portion of `push-cip.sh`. The `pull-cip-auditor-e2e` prow job only notifies us if we can no longer push auditor test images. Therefore, there's no single test that verifies every portion of `push-cip.sh` to be working.
3. Narrowing failures will be much easier if we have overlapping tests. For example, if the `pull-cip-auditor-e2e` job fails, that could mean a bug in `golden-push.sh`, `cip-auditor-e2e.go`, or `push-cip.sh` to name a few. But if we get this same job failure, with a passing test for `push-cip.sh`, that's one less executable to worry about.

For these reasons, I believe it's necessary to add a new prow pre-submit, in accompaniment with `pull-cip-image-cip`.

hack/test-cip-image.sh
```diff
@@ -0,0 +1,79 @@
...
+# Validate the container build process for the given variant.
+testVariant() {
+    # Build based on the variant.
+    if [[ "$1" == "auditor" ]]; then
+        make image-build-cip-auditor-e2e
+    else
+        make image-build
+    fi
+
+    # Check required images are built locally.
```
listx, line 51: Because we have `set -o errexit` on, if the `make` targets above execute successfully, then by definition the images have been built by the time we're here in line 51. So the `validateImages()` isn't providing a lot of value.

It's sort of like saying, "test the compilation of the cip binary by calling `go build` on it and then listing all the local go binaries in the system to see if the binary is there". It's not adding much value.

Anyway, is this script necessary? Why not just make the `test-cip-image-build` target invoke the `image-build` and `image-build-cip-auditor-e2e` targets directly? Isn't that the same thing? You can have multiple things performed in a single make target, like this:

```
some-target:
    doSomething
    doAnotherThing
```

tylerferrara: This script ensures the required images are built. If we somehow modify or extend cip-image.sh to where the `gcr/k8s-staging-artifact-promoter/cip:[stable-image-tag]` was no longer present, we would catch it with this script. We are essentially guarding the image bundles for cip/prod and auditor/test variants.

tylerferrara:
After offline discussion with @listx, improvements to provenance (guarding code) should be prioritized in a new issue.

tylerferrara:
This testing effort will be tracked by the issue: #339
There, each area of interest will be addressed separately.

PR #331, opened 2021-07-02
Modify push-cip.sh to handle builds and pushes separately

tylerferrara:
#### What type of PR is this?
/kind cleanup

This change separates the build and push processes by modifying the `push-cip.sh` script - now called `cip-image.sh`. Such a separation allows for simplified testing of the build process only. If we were to test the original `push-cip.sh` script, a `docker push` would be unavoidable, requiring proper authentication. Now, this script can specify whether to build OR push images.

```
Usage: ./cip-image.sh <command> [--audit]
Commands:
  build     docker build images from the project's Dockerfile
  push      docker push images to their tagged location

Optional Flag:
  --audit   handle images for e2e test auditor
  ```

This simplifies the testing of #329
Partially satisfies #328 

None

NONE

```release-note
Replace cip-push.sh with cip-image.sh.
Add new make targets: image-build and image-build-cip-auditor-e2e.
```

### DIFF
--- a/hack/cip-image.sh
+++ b/hack/cip-image.sh
@@ -14,27 +14,61 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# About: This script builds and pushes the cip and auditor images to a public repository,
-# defined in workspace_status.sh.
+# About: This script can either build or push the cip and auditor container images. Providing the '--audit' flag
+# handles test images used in the e2e auditor.
+
+# Usage: cip-image.sh <command> [--audit]
+# Commands:
+#   build     docker build images from the project's Dockerfile
+#   push      docker push images to their tagged location
+#
+# Optional Flag:
+#   --audit   handle images for e2e test auditor
 
 set -o errexit
 set -o nounset
 set -o pipefail
 
 repo_root=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)
-# Detect which container variant to build.
-build_variant="cip"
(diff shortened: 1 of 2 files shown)


hack/cip-image.sh
```diff
@@ -14,27 +14,60 @@
...
 }
 
-if (( $# > 1 )) || { (( $# > 0 )) && [[ "$1" != --audit ]];}; then
-    >&2 echo "ERROR: Invalid arguments."
+# Parse and validate runtime arguments.
+if (( $# < 1 )) || (( $# > 2 )); then
+    >&2 echo "ERROR: Invalid number of arguments."
+    printUsage
+    exit 1
+elif { [[ "$1" != "build" ]] && [[ "$1" != "push" ]]; }; then
```
listx, line 55: I would close the above if-statement (lines 51-54) here with `fi`.

Then I would capture the operation with line 60 (incremental parsing).

Then I would `shift` to "consume" the argument (to really hit home the fact that we're parsing incrementally for the reader, because we're done with processing the "verb" of `build` or `push`).

Then I would validate the operation to check whether it was `build` or `push` (it would be its own `if` --- `fi` block).

Lastly I would check if there's another optional argument for `--audit`. At this point the `(( $# == 2 ))` check would be unnecessary. You can also error out if the flag is anything other than `--audit`, such as a typo'd `--auditt` or some other.

listx:
Thanks!

/lgtm
/approve

Issue #300, opened 2021-05-30
Remove bazel from repo


justaugustus:
- [x] Update to latest / roughly parity with k/k: https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/264
- [ ] ~Swap to k/repo-infra~
- [ ] Non-bazel bash/go testing
  - [x] unit tests: https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/291
  - [ ] (@tylerferrara) e2e tests: https://github.com/kubernetes-sigs/k8s-container-image-promoter/issues/304
- [ ] Non-bazel container + artifact generation
- [ ] Swap presubmits to non-bazel
- [ ] Remove bazel

/sig release
/area release-eng

/assign @tylerferrara @justaugustus
cc: @listx @kubernetes-sigs/release-engineering

justaugustus:
Design doc from @tylerferrara: https://docs.google.com/document/d/1tQTb8dqKQtsL1a5jmCDbRg2MEDMkAnT07vrrxyZxLfk/edit?usp=sharing

cc: @kubernetes-sigs/release-engineering

tylerferrara:
Removing Bazel from e2e tests will take several PRs. To organize this, I created a sub-issue outlining each step.
CIP e2e tests: #304

tylerferrara:
- [x] e2e tests and completed

#323 is looking to tackle `Non-bazel container + artifact generation`

tylerferrara:
@justaugustus Bazel is now deprecated in CIP.
Suggest closing.

PR #324, opened 2021-06-28
Building cip with go CLI

tylerferrara:
#### What type of PR is this?
/kind cleanup
/kind deprecation

#### What this PR does / why we need it:
This replaces Bazel, for building go code, with the go CLI. Since Bazel is no longer building any binaries within this project, Gazell's dependency tracking can also be removed.

Partially satisfies #300 
Part of the effort to [reduce build maintenance](https://docs.google.com/document/d/1tQTb8dqKQtsL1a5jmCDbRg2MEDMkAnT07vrrxyZxLfk/edit?usp=sharing).

#### Which issue(s) this PR fixes:
None

#### Does this PR introduce a user-facing change?
NONE

```release-note
Remove Bazel build and installation, in favor of the go CLI.
```
cc: @listx @amwat @justaugustus @kubernetes-sigs/release-engineering

PR #333, opened 2021-07-02
Remove all Bazel artifacts

tylerferrara:
#### What type of PR is this?
/kind cleanup
/kind deprecation

Since Bazel is no longer relied upon within CIP, this change deletes all Bazel-related files. Additionally, all documentation reflects this complete deprecation, removing all mention of Bazel. 

Part of the effort to [Reduce Build Maintenance](https://docs.google.com/document/d/1tQTb8dqKQtsL1a5jmCDbRg2MEDMkAnT07vrrxyZxLfk/edit?usp=sharing).
#300 

```release-note
Remove Bazel-related files from the repository.
Update Docs to reflect Bazel deprecation.
```


tylerferrara:
NOTE: `dependencies.yaml` must be kept, as it's used in the [verify-dependencies](https://github.com/kubernetes-sigs/k8s-container-image-promoter/blob/master/Makefile#L111) make target, currently triggered by the pre-submit: `pull-cip-verify`. However, if this is no longer needed, please indicate as such, and I can make an accompanying PR.

justaugustus:
@tylerferrara -- Please leave `dependencies.yaml` in place as it's used to track non-bazel dependencies.

tylerferrara:
Additionally, should `legacy/` remain untouched? This function is being referenced in many other legacy programs:
```golang
func bazelTestPath(testName string, paths ...string) string {
	prefix := []string{
		os.Getenv("PWD"),
		"inventory_test",
		testName,
	}
	return filepath.Join(append(prefix, paths...)...)
}
```

listx:
> Additionally, should `legacy/` remain untouched? This function is being referenced in many other legacy programs:
> 
> ```go
> // legacy/dockerregistry/inventory_test.go
> func bazelTestPath(testName string, paths ...string) string {
> 	prefix := []string{
> 		os.Getenv("PWD"),
> 		"inventory_test",
> 		testName,
> 	}
> 	return filepath.Join(append(prefix, paths...)...)
> }
> ```
> 
> Even just changing the name may help to reduce confusion.
> What do you think? @listx @justaugustus @amwat

Yeah we should change the name. You should do this in a separate commit (but within this PR), because renames (that don't change any behavior) are special and should be isolated from other changes.

VERSION
```diff
@@ -1 +0,0 @@
-1.337.0
```
listx, line 1: Is this file necessarily tied to Bazel?

justaugustus (approved):

hack/verify-archives.sh
```diff
@@ -16,7 +16,7 @@
 
 # About:
 # This script extracts the golden image tarballs and compares their contents against the
-# already-tracked golden folder (which is used by Bazel to generate deterministic images
+# already-tracked golden folder (which is used by Docker to generate deterministic images
```
listx, line 19: Docker does not generate deterministic images.

That said, this parenthetical is worth deleting entirely.

VERSION
justaugustus: Good catch. Let's restore this.

Makefile
```diff
@@ -22,7 +22,7 @@ all: test
 ##@ Build
 .PHONY: build
 build: ## Go build
-	go build -o cip ${REPO_ROOT}/cmd/cip/...
+	${REPO_ROOT}/go_with_version.sh build ${REPO_ROOT}/cmd/cip
```
tylerferrara, line 25: This should have been merged with #324 
When building, we currently have an incomplete version stamping.
```bash
$ cip version
GitVersion:    
GitCommit:     
GitTreeState:  
BuildDate:     
GoVersion:     go1.16.5
Compiler:      gc
Platform:      linux/amd64
```

With this change, we get all the information we want.
```bash
$ cip version
GitVersion:    v3.0.0
GitCommit:     999f2538d2d4a63a1b5c389d7b53411e9ad24226
GitTreeState:  clean
BuildDate:     2021-07-07T21:06:08Z
GoVersion:     go1.16.5
Compiler:      gc
Platform:      linux/amd64
```

cmd/cip/cmd/root.go
```diff
@@ -26,17 +26,6 @@ import (
 	"sigs.k8s.io/release-utils/log"
 )
 
-var (
```
tylerferrara, line 29: This is part of the stamping fix, as mentioned above.

hack/verify-archives.sh
tylerferrara: ack.

justaugustus (approved):
@tylerferrara -- Great work!
@listx -- the review comments look to all be addressed, but if there's anything outstanding, let's take that as a follow-up.

/hold cancel

Issue #345, opened 2021-07-12
Version flag not printing version


tylerferrara:
#### What happened:
Using the `--version` flag does not print version.

#### What you expected to happen:
```bash
# from the CIP project root
make build && ./cip audit --version
```
output should be the same as the `version` command:
```bash
GitVersion:    v3.0.0
GitCommit:     b3e6408e6c99329eeea451c1ca92fd69d4b128e2
GitTreeState:  clean
BuildDate:     2021-07-12T18:59:36Z
GoVersion:     go1.16.5
Compiler:      gc
Platform:      linux/amd64
```

#### How to reproduce it (as minimally and precisely as possible):
```bash
# from the CIP project root
make build && ./cip audit --version
```
output:
```bash
INFO Setting image auditor options...             
INFO Starting auditor in Regular Mode (bd1f3917-8f88-4150-a7c3-820f28a7d59c) 
INFO Image auditor options: [GCP project: , repo URL: , repo branch: , path: , UUID: bd1f3917-8f88-4150-a7c3-820f28a7d59c]
```

#### Anything else we need to know?:
Since the `version` should command accomplishes the same thing, we should remove the `--version` flag altogether.

#### Environment:

- Cloud provider or hardware configuration:
Home Desktop Workstation
- OS (e.g: `cat /etc/os-release`):
NAME="Ubuntu"
VERSION="21.04 (Hirsute Hippo)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 21.04"
VERSION_ID="21.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=hirsute
UBUNTU_CODENAME=hirsute

- Kernel (e.g. `uname -a`):
Linux tyler-ubuntu 5.11.0-22-generic #23-Ubuntu SMP Thu Jun 17 00:34:23 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

- Others:
cc: @listx @amwat @justaugustus @kubernetes-sigs/release-engineering

tylerferrara:
Resolved by the merging of #346.
Closing.

PR #346, opened 2021-07-12
Remove --version flag from cip command.

tylerferrara:
#### What type of PR is this?
/kind bug
/kind cleanup
/kind api-change

This change removes the `--version` flag in favor of the existing `version` command. Currently the flag does not actually change any behavior in CIP, nor does it actually print the version. Since we already have a command responsible for producing version information, we no longer need this flag.

Such a change WILL effect our users, as it's an API change of the `cip` CLI. Therefore, this should be pointed out within the release notes of the next CIP major version.

#345 

**This API change should be documented in the next release.**

YES

The flag `--version` no longer exists. 

```release-note
Remove the version flag from the cobra cli.
```

### DIFF
--- a/cmd/cip/cmd/root.go
+++ b/cmd/cip/cmd/root.go
@@ -59,13 +59,6 @@ func init() {
 		rootOpts.DryRun,
 		"test run promotion without modifying any registry",
 	)
-
-	rootCmd.PersistentFlags().BoolVar(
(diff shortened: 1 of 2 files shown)


tylerferrara:
How can we keep track of this API change for release documentation?
@justaugustus

listx:
Thanks!

/lgtm
/approve

Issue #259, opened 2020-09-29
make auditor less network-heavy on each pub/sub request


listx:
Currently the auditor scans the entire staging repo "responsible" for a promotion to prod for every single incoming pub/sub event. This can get very heavy if the staging repo is large and/or the staging repo has tons of images. This has resulted in hitting the network API limits (quota limits) against GCR, where the staging repos reside.

There are two optimizations to do:

1. don't scan the entire staging repo, but instead only the subdirectory containing the image-that-was-promoted.
2. make the auditor process pub/sub messages in batch mode --- that is, only read staging repos every 20 seconds or some such, regardless of the number of pub/sub messages received in-between

The first item should be good enough as a temporary stopgap, and should be straightforward to implement. The 2nd item is more difficult as it changes how the auditor behaves (and requires introducing new concurrency designs).

listx:
Oops

/remove-lifecycle stale

PR #351, opened 2021-07-14
`kpromo cip`: Initial parent image verification work

tylerferrara:
#### What type of PR is this?
/kind feature
/kind api-change

In order to explore the possible resolution of #259 with simpler queries to GCR, we must first validate that all parent images have children that are located in the same project location. If guaranteed, we no longer have to query the entire GCR repository to find verify a child image, since we know it's location. This improves the speed complexity of the auditor from O(n) to O(1), where n represents the number of images within GCR.

To use this feature, we must first create a snapshot of GCR and save it to file:

```console
kpromo cip run --snapshot us.gcr.io/k8s-artifacts-prod > ./snap.yaml
```

Now, we can check this manifest snapshot:
```console
kpromo cip run --repository us.gcr.io/k8s-artifacts-prod --check-manifest-lists-from-snapshot ./snap.yaml
```

Notice the required `--repository` flag, which indicates where these images can be queried. Once the verification starts, all snapshot images are queried for their manifests. If an image is of `"mediaType": "application/vnd.docker.distribution.manifest.list.v2+json"`, then we know it's a `parent` image, holding pointers to other `child` images. Each child image is then verified to be in the same image location as the parent.

Image location: `us.gcr.io/some-project/my-image-name`

NONE

This feature is **experimental**, and was only created to automate a validation process. In order to be considered for integration with the `cip` tool, it should first be cleaned up, re-written, and tested.

```release-note
kpromo cip: Initial parent image verification work
```


listx:
For posterity, this PR documents work @tylerferrara did to check whether there are any manifest lists in the production k8s-artifacts-prod GCR that have child images that reside in a location other than the same location as the parent. The answer so far to this question is "no".

Answering this question with a "no" is important because it opens up the door to doing selective, partial reads of GCR whenever the cip auditor receives a Pub/Sub message of a state change in GCR.

Tyler can clean this up to reuse more of the existing code (for one thing, we don't need to parse the snapshot text file, and instead can augment the existing snapshotting functionality to perform this parent/child analysis directly in one step).

justaugustus (approved):

PR #357, opened 2021-07-21
Fix unit tests

### LINKED ISSUE #322: Fix silently broken tests
#### What happened:
In https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/302#issuecomment-869847962, we explained that some of the unit tests have changed in behavior in https://github.com/kubernetes/release/pull/1772 (when the test code lived for a time in k/release).

The strictness of a number of tests were relaxed (before, we checked that certain particular known cases of bugs would trigger particular codepaths with different error messages).

If a test's expectedError field is changed, the test should fail.

Modify the error message in the `expectedError` field of `TestParseRegistryManifest()` in inventory_test.go. 

#### Anything else we need to know?

We can either drop the `expectedErr` field from the anonymous structs, or restore the old test behavior where we actually checked the expected error message. I'm in favor of the latter.

listx:
#### What type of PR is this?

/kind bug
/kind regression

Fixes #322 

To be more precise, we used `require.Regexp()` and `require.Equal()` wrong in https://github.com/kubernetes/release/pull/1772 because we did not change our argument order from `got/actual, expected` to `expected, got/actual` as is required by the `require` package. So this resulted in various tests actually doing the wrong thing, even though things were passing (this was especially pronounced in the tests involving `require.Regexp()`).

Fixes #322

NONE

### DIFF
--- a/legacy/audit/auditor_test.go
+++ b/legacy/audit/auditor_test.go
@@ -22,10 +22,9 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"reflect"
 	"regexp"
 	"testing"
-
-	"github.com/stretchr/testify/require"
 
 	"sigs.k8s.io/k8s-container-image-promoter/legacy/audit"
 	reg "sigs.k8s.io/k8s-container-image-promoter/legacy/dockerregistry"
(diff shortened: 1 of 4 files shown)


listx:
I've created reverts by mostly copy/pasting the code in k/release-infra just prior to the merge of https://github.com/kubernetes/release/pull/1772, which introduced the regressions as described in #322.

puerco:
Thanks !!
/lgtm

Issue #358, opened 2021-07-21
Audit Request Counter


tylerferrara:
#### What would you like to be added:
Enable a simple request counter when the user passes the `--verbose` flag. This counter should increment each time the auditor makes an HTTP request to GCR. Every 10min period, the number of queries should printed to logs and reset the counter back to zero. This effectively measures the network performance of the auditor in the same way the Container Registry measures it's [quota](https://cloud.google.com/container-registry/quotas): 50,000 HTTP requests every 10 minutes.

**Requirements:**

- [x] Create a simple request counter #359
- [x] Concurrently log the counter at 10min intervals #363 
- [x] Create multiple counters that log at different offsets #364 #383
- [x] Allow for non-resettable request counters #382 
- [x] Create warning if queries exceed the GCR quota #384
- [x] Test the Log function with fake time #370 #371 #376 #377 
- [x] Enable this feature for all audit builds #367 #381 
- [ ] Allow for gracefully termination
- [ ] Create a quota structure that uses a sliding window

#### Why is this needed:
The auditor makes many requests to GCR in order to validate children of manifest.lists. This is currently handled by requesting all images within the project:`k8s-artifacts-prod` which is a known bottleneck #259.

Additionally, the way in which we request GCR is via simple HTTP requests which are not tracked by the Container Registry API  #353. Therefore, gaining visibility into the number of requests the auditor is making while running can help diagnose throttling. This behavior often causes a backup of unacknowledged Pub/Sub messages, resulting in unwanted alerts [#2364](https://github.com/kubernetes/k8s.io/issues/2364).

cc: @listx @amwat @justaugustus @kubernetes-sigs/release-engineering

PR #359, opened 2021-07-21
Introduce a request counter to the auditor

tylerferrara:
#### What type of PR is this?
/kind feature

This change creates a simple request counter, able to concurrently keep track of all HTTP requests  made to GCR. The counter is only enabled when the user passes the `--verbose` flag. In the future, this counter will be logged to gain visibility in the network performance of the auditor.

Supersedes #356
Partially satisfies #358 

The auditor does not yet call `Increment()` to add to the counter. This would perpetually increase the value of the counter if enabled. Once the concurrent logger is implemented (part of #358), which resets the counter every 10min, will the auditor call `Increment()`.

Less importantly, this package is made specifically for the auditor, it can just as easily be used for other commands, like `cip run --snapshot`. 

NONE

```release-note
Triggering request counter for the auditor when --verbose flag is present.
```

### DIFF
--- a/legacy/cli/audit.go
+++ b/legacy/cli/audit.go
@@ -24,6 +24,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"sigs.k8s.io/k8s-container-image-promoter/legacy/audit"
+	"sigs.k8s.io/k8s-container-image-promoter/legacy/reqcounter"
 )
 
 type AuditOptions struct {
@@ -52,6 +53,11 @@ func RunAuditCmd(opts *AuditOptions) error {
 	)
 	if err != nil {
 		return errors.Wrap(err, "creating auditor context")
+	}
+
+	if opts.Verbose {
+		// Initialize global counter to track the number of HTTP requests made to GCR.
+		reqcounter.Init()
 	}
 
 	auditorContext.RunAuditor()
--- a/legacy/reqcounter/reqcounter.go
+++ b/legacy/reqcounter/reqcounter.go
@@ -1,0 +1,53 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package reqcounter
+
+import "sync"
+
+// RequestCounter records the number of HTTP requests to GCR.
+type RequestCounter struct {
+	mutex    sync.Mutex
+	requests uint64
+}
+
+var (
+	// enableCounting will only become true if the Init function is called. This allows
+	// requests to be counted and logged.
+	enableCounting = false
+	// counter will continuously be modified by the Increment function to count all
+	// HTTP requests to GCR.
+	counter = RequestCounter{}
+)
+
+// Init allows request counting to begin.
+func Init() {
+	enableCounting = true
+	counter = RequestCounter{
+		mutex:    sync.Mutex{},
+		requests: 0,
+	}
+}
+
+// Increment increases the request counter by 1, signifying an HTTP
+// request to GCR has been made.
+func Increment() {
+	if enableCounting {
+		counter.mutex.Lock()
+		counter.requests++
+		counter.mutex.Unlock()
+	}
+}


legacy/querycounter/querycounter.go
```diff
@@ -0,0 +1,54 @@
...
+type networkMonitor struct {
+	mutex   sync.Mutex
+	queries uint64
+}
+
+var (
+	// enableLogging will only become true if the Init function is called. This signifies
+	// the query counter to begin monitoring network performance.
+	enableLogging = false
+	// counter will continiuously be modified by the Increment function to count all
```
listx: continiuously -> continuously

legacy/cli/audit.go
```diff
@@ -54,6 +55,11 @@ func RunAuditCmd(opts *AuditOptions) error {
 		return errors.Wrap(err, "creating auditor context")
 	}
 
+	// Allow query counting when the --verbose flag is passed.
```
listx, line 58: Please move this comment to be inside the `if opts.Verbose` block, because it could be that in the future the Verbose flag does other things as well.

Please reword the comment to read:

```
// Initialize global counter to track the number of HTTP requests made to GCR.
```

legacy/cli/audit.go
```diff
@@ -24,6 +24,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"sigs.k8s.io/k8s-container-image-promoter/legacy/audit"
+	"sigs.k8s.io/k8s-container-image-promoter/legacy/querycounter"
```
listx, line 27: I think a better name would be `requestcounter` or `reqcounter` because we'll use it to track HTTP *requests* made to GCR (in industry parlance the phrase "HTTP request" is the standard phrase, rather than "HTTP query"). Please change.

legacy/querycounter/querycounter.go
```diff
@@ -0,0 +1,54 @@
...
+*/
+
+package querycounter
+
+import "sync"
+
+// networkMonitor records the number of HTTP requests (queries) to GCR.
+type networkMonitor struct {
+	mutex   sync.Mutex
+	queries uint64
```
listx, line 24: Please rename this variable to `requests`.

legacy/querycounter/querycounter.go
```diff
@@ -0,0 +1,54 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package querycounter
+
+import "sync"
+
+// networkMonitor records the number of HTTP requests (queries) to GCR.
+type networkMonitor struct {
```
listx, line 22: Can this struct be renamed to just `counter` (I forget if Golang allows variables and structs to have the same name)?

legacy/querycounter/querycounter.go
```diff
@@ -0,0 +1,54 @@
...
+	// the query counter to begin monitoring network performance.
+	enableLogging = false
+	// counter will continiuously be modified by the Increment function to count all
+	// queries made to GCR.
+	counter = networkMonitor{}
+)
+
+// Init allows query counting to begin.
+func Init() {
+	enableLogging = true
```
listx, line 38: A better name would be `enableCounting` because counting (incrementing) is what we're doing in this code inside `Increment()`. Please rename.

legacy/querycounter/querycounter.go
```diff
@@ -0,0 +1,54 @@
...
+	counter = networkMonitor{
+		mutex:   sync.Mutex{},
+		queries: 0,
+	}
+}
+
+// Increment increases the query counter by 1, signifying an HTTP
+// request to GCR has been made.
+func Increment() {
+	// Logging must be enabled to count queries.
```
listx, line 48: Please prefer the word `requests` or phrase `HTTP requests` over the word `queries` throughout the code to make it easier to grep for folks new to the codebase. Consistency.

listx (changes requested):

legacy/querycounter/querycounter.go
tylerferrara: Fixed.

legacy/cli/audit.go
tylerferrara: Ack.

legacy/cli/audit.go
tylerferrara: Understood.

legacy/querycounter/querycounter.go
tylerferrara: ack.

legacy/querycounter/querycounter.go
tylerferrara: Changing to `RequestCounter` to avoid confusion with the local variable.

legacy/querycounter/querycounter.go
tylerferrara: ack.

legacy/querycounter/querycounter.go
tylerferrara: Got it.

tylerferrara (commented):
Replaced all mention of "query" counting with "request" counting.

legacy/reqcounter/reqcounter.go
```diff
@@ -0,0 +1,54 @@
...
+	counter = RequestCounter{
+		mutex:    sync.Mutex{},
+		requests: 0,
+	}
+}
+
+// Increment increases the request counter by 1, signifying an HTTP
+// request to GCR has been made.
+func Increment() {
+	// Counting must be enabled to actually count requests.
```
listx, line 48: Please delete this comment line, as the code is obvious enough as it stands.

tylerferrara: Done

listx:
Thanks!

/lgtm
/approve

PR #360, opened 2021-07-21
Log request counter in intervals

tylerferrara:
#### What type of PR is this?
/kind feature

This change logs the number of HTTP requests to GCR every 10min if the auditor is passed the `--verbose` flag. If enabled, the request counter will continuously produce logs that look like this:
```
Since 2006-01-02 15:04:05 there have been 3402 requests to GCR.
```

Since the actual Container Registry defines it's [quota](https://cloud.google.com/container-registry/quotas) as 50,000 HTTP requests every 10 minutes, this feature will give us insight into how close the auditor is getting to this upper-limit.

**NOTE:** If the auditor surpasses this GCR quota, the responses will be throttled, resulting in slower image verification. Depending on the number of Pub/Sub messages it must respond to, this can lead to unacknowledged messages. 
Partially satisfies #358 

The `--verbose` flag will not be added to auditor builds (the Dockerfile) until tests validate all behavior.

NONE

```release-note
Continuously log request counter every 10min. 
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,8 @@ WORKSPACE
 
 # downloaded and built binaries
 bin
+# tmp files
+tmp/
 
 # vscode settings
 .vscode


legacy/reqcounter/reqcounter.go
```diff
@@ -16,30 +16,73 @@ limitations under the License.
...
+	rc.requests = 0
+	rc.since = time.Now()
 }
 
+const (
+	// measurementWindow specifies the length of time to batch requests to GCR. Since Google's
+	// Container Registry specifies a quota of 50,000 HTTP requests per 10 min, the window
+	// for recording requests is set to 10 min.
+	// Source: https://cloud.google.com/container-registry/quotas
+	measurementWindow = time.Minute * 10
```
listx, line 59: 1. How do we know that the 10 minutes that we measure is the the same 10 minute block measured by GCP? Does this matter?

Trying to figure out this question myself, I looked at https://cloud.google.com/container-registry/quotas and then clicked on the link to https://cloud.google.com/docs/quota. There is a lot of info there and it appears that quotas are already automatically monitored by GCP (see https://cloud.google.com/docs/quota#viewing_quota).

BUT according to this heading https://cloud.google.com/docs/quota#checking_metric_support it's unclear if GCR quotas (which seem to be tied to Google Cloud Storage quotas) are automatically monitored. FWIW I have seen the `429 TOO MANY REQUESTS` error (see https://cloud.google.com/docs/quota#quota_errors) before from the auditor, so we have definitely hit the quota limit before.

2. Can you just double-check that HTTP requests to GCR are *not* visible even with the `Quota` tool in GCP, under the top-level `IAM & Admin` tool in GCP?

legacy/reqcounter/reqcounter.go
```diff
@@ -16,30 +16,73 @@ limitations under the License.
...
+	"github.com/sirupsen/logrus"
+)
 
 // RequestCounter records the number of HTTP requests to GCR.
 type RequestCounter struct {
-	mutex    sync.Mutex
-	requests uint64
+	mutex    sync.Mutex // Lock to prevent race-conditions with concurrent processes.
+	requests uint64     // Number of HTTP requests since recording started.
+	quota    uint64     // Maximum requests allowed before GCR throttling occurs.
```
listx, line 31: This field shouldn't be here, because it's already a global, read-only constant in the module.

legacy/reqcounter/reqcounter.go
```diff
@@ -16,30 +16,73 @@ limitations under the License.
...
+// log records the number of HTTP requests found and resets the counter.
+func (rc *RequestCounter) log() {
+	// Grab the lock to prevent race-conditions.
+	rc.mutex.Lock()
+	defer rc.mutex.Unlock()
+	// Log the number of requests found.
+	msg := fmt.Sprintf("Since %s there have been %d requests to GCR.", counter.since.Format("01-02-2006 15:04:05"), counter.requests)
+	logrus.Debug(msg)
+	// Reset the counter.
+	rc.reset()
```
listx, line 44: To make it easier to search our logs from the GCP console, can you have 2 counters?

1. a global counter that always increments and is never reset
2. a counter that is reset each measurement window (you already have this)

And, can you log the 1st counter as well whenever we log the 2nd counter? Just seeing a running total will be useful.

legacy/dockerregistry/inventory.go
```diff
@@ -1593,6 +1594,9 @@ func MkReadRepositoryCmdReal(
 ) stream.Producer {
 	var sh stream.HTTP
 
+	// Log the HTTP request to GCR.
+	reqcounter.Increment()
```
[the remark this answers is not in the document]
tylerferrara: Good Point!

legacy/reqcounter/reqcounter.go
tylerferrara: I included this in the RequestCounter for easier testing that relies solely on the object, not global constants. But I can get around this.

legacy/reqcounter/reqcounter.go
```diff
@@ -16,30 +16,73 @@ limitations under the License.
...
+	requests uint64     // Number of HTTP requests since recording started.
+	quota    uint64     // Maximum requests allowed before GCR throttling occurs.
+	since    time.Time  // When the current counter began recording requests.
+}
+
+// log records the number of HTTP requests found and resets the counter.
+func (rc *RequestCounter) log() {
+	// Grab the lock to prevent race-conditions.
+	rc.mutex.Lock()
+	defer rc.mutex.Unlock()
```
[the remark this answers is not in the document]
tylerferrara: I created `log` and `reset` as methods for RequestCounter because they are unique to the counter itself. Therefore, it has built in functions to modify itself. Having these also simplifies the `requestLogger` function.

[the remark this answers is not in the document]
tylerferrara: Ohhhh! I see your point, because I was still reading `counter.requests`. Fixing!

legacy/reqcounter/reqcounter.go
```diff
@@ -16,30 +16,73 @@ limitations under the License.
...
+	// Log the number of requests found.
+	msg := fmt.Sprintf("Since %s there have been %d requests to GCR.", counter.since.Format("01-02-2006 15:04:05"), counter.requests)
```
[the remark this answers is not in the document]
tylerferrara: Ack.

legacy/reqcounter/reqcounter.go
```diff
@@ -16,30 +16,73 @@ limitations under the License.
...
+	logrus.Debug(msg)
+	// Reset the counter.
+	rc.reset()
+}
+
+// reset clears the number of requests and stamps the object with the current time of reset.
+// WARNING: To avoid race conditions, only use this method when holding the counter's mutex.
```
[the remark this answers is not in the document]
tylerferrara: It's not safe to release the lock and expect to immediately grab it again. Here is an example of what could happen:

0. log grabs the lock
1. log collects the requests
2. log lets go of the lock
3. Increment grabs the lock
4. Increment adds 1 to the already logged counter
5. Increment releases the lock
6. reset grabs the lock
7. reset wipes the counter
8. reset lets go of the lock

In this scenario, we reset the lock with an extra request that was not logged. To avoid this, the logic must remain how it is, or we add the `reset` logic inside of the `log` function.

legacy/reqcounter/reqcounter.go
```diff
@@ -16,30 +16,73 @@ limitations under the License.
...
+func (rc *RequestCounter) reset() {
+	rc.requests = 0
+	rc.since = time.Now()
 }
 
+const (
+	// measurementWindow specifies the length of time to batch requests to GCR. Since Google's
```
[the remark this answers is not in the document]
tylerferrara: Ack.

legacy/reqcounter/reqcounter.go
tylerferrara: 1. I am not going to assume the 10min window we are logging at will line up with GCP's quota. This feature is merely a benchmark to aid in the understanding of how many times the auditor is requesting GCP. Therefore, it is possible for our logs to reflect just under the quota limit, when in actuality we surpassed the quota and are being throttled. But this is okay, as these logs will at least tell us that we got VERY close to the quota limit during that time period.

tylerferrara: 2. After offline discussion, we both discovered we can't view these HTTP requests to GCR in any GCP service. Therefore, this is the only way to gain insight into how much we are querying GCR.

tylerferrara (commented):
The mutex must be held the entire time the RequestCounter object is being modified. Therefore, I can either:
1. Leave the logic the way it is (with some sort of warning about race conditions)
2. Add the reset logic to the log function

legacy/reqcounter/reqcounter.go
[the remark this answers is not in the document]
listx: Maybe this is more complicated than it needs to be? What if we only lock during writes to the counter, not reads?

That is, we could remove step 0 and 2 in your example (i.e., make `log` read the requests counter without using a lock); this loses accuracy down to 1 (maybe 2?) requests, but I think it's a fine because (1) this off-by-1 or 2 error does not accumulate over time (that is, it will never become an off-by-1000 or 2000 error), and (2) we don't even know when the GCP reference 10-minute interval starts/stops (even if we did, we won't be able to synchronize our clock to match GCP's exactly anyway), so our own 10-minute interval is a best-guess effort.

legacy/reqcounter/reqcounter.go
listx: Thinking about this a bit more, it would be helpful to expand counters (if necessary) to log how many requests were made in the last:

1. 10 minutes (we already do this in this PR)
2. 1 hour (not yet implemented)
3. 24-hour period (not yet implemented)

The request counter information for the previous 1-hour and 24-hour periods would both be a rolling on a 10-minute basis. That is, every 10 minutes the log would say:

```
There were X HTTP requests to GCR made in the last 10 minutes.
There were Y HTTP requests to GCR made in the last hour.
There were Z HTTP requests to GCR made in the last 24 hours.
```

legacy/reqcounter/reqcounter.go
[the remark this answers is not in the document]
tylerferrara: This still doesn't seem to be the way to handle race-conditions. The simplest approach, that still feels most correct, is to just combine the reset logic and let go of the lock when reading and writing has ended. Race conditions are eliminated and there is one less function to handle.

legacy/reqcounter/reqcounter.go
tylerferrara: Could this be added in a following PR?

legacy/dockerregistry/inventory.go
```diff
@@ -1607,6 +1608,9 @@ func MkReadRepositoryCmdReal(
 			repoPath)
 	}
 
+	// Log the HTTP request to GCR.
+	reqcounter.Increment()
```
listx, line 1612: Looking at this code again, we should instead just call reqcounter.Increment() once, inside the `ExecRequests()` function, inside the `for reqRes := range requestResults {` loop. This is both simpler (it's only 1 place we have to increment instead of 2 places), and also reuses the RequestResult channel which already acts as a global "funnel" where all requests, *after they've been processed*, comes in. Another reason we should do it there is that it will work for both real HTTP requests, and also fake ones we generate during unit tests (so it appears this was bound to come up anyway).

I should have taken a closer look the first time around, so this is my bad. But please do address it! :)

tylerferrara: Thank you! I'm sorry I didn't ask this myself. I was actually unsure where to put this counter in the first place. This is very helpful.

legacy/reqcounter/reqcounter.go
tylerferrara: Added an extra TODO within #358  "Create multiple counters that log at different offsets"

listx:
Thanks!

/lgtm
/approve

PR #361, opened 2021-07-21
[WIP] Log when GCR requests exceed quota

tylerferrara:
#### What type of PR is this?
/kind feature

This change logs a warning when the auditor is observed to make more HTTP requests than the [GCR quota](https://cloud.google.com/container-registry/quotas): 50,000 requests per 10min. In this scenario, the auditor is expected to be throttled by GCR and experience longer response times from each request. This behavior should be avoided at all costs, since throttling will also slow the verification process of each GCR change. In the worst case, this slowdown may leave multiple Pub/Sub messages unacknowledged causing a backup of verifications and cause alerts to fire. This change looks to gain visibility into when, if at all, this behavior occurs.

Partially satisfies #358

This warning will only occur if the request counting is enabled in the first place. Therefore, the auditor must be run with the `--verbose` flag to gain such visibility.

NONE

```release-note
Log a warning when the number of GCR requests exceeds quota.
```

### DIFF
--- a/legacy/reqcounter/reqcounter.go
+++ b/legacy/reqcounter/reqcounter.go
@@ -40,8 +40,18 @@ func (rc *RequestCounter) log() {
 	// Log the number of requests found.
 	msg := fmt.Sprintf("Since %s there have been %d requests to GCR.", counter.since.Format("01-02-2006 15:04:05"), counter.requests)
 	logrus.Debug(msg)
+	// Warn of GCR scaling when requests exceed quota.
+	rc.warn()
 	// Reset the counter.
 	rc.reset()
+}
+
+// warn only logs a warning if the number of requests has exceeded the quota.
+func (rc *RequestCounter) warn() {
+	if rc.requests > rc.quota {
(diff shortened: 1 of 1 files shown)


tylerferrara:
Add "error code" to make filtering logs easier

legacy/reqcounter/reqcounter.go
```diff
@@ -16,30 +16,83 @@ limitations under the License.
...
+	// Warn of GCR scaling when requests exceed quota.
+	rc.warn()
+	// Reset the counter.
+	rc.reset()
+}
+
+// warn only logs a warning if the number of requests has exceeded the quota.
+func (rc *RequestCounter) warn() {
+	if rc.requests > rc.quota {
+		msg := fmt.Sprintf("The GCR quota of \"%d requests per %d min\" has been exceeded.", quotaLimit, measurementWindow/10)
```
listx, line 52: Create a unique "error-code"-like string to make it easier to search after-the-fact?

QUOTA_EXCEEDED ?

tylerferrara:
Closing in favor of #384

PR #364, opened 2021-07-22
Support multiple request counters

tylerferrara:
#### What type of PR is this?
/kind feature

This change modifies the `reqcounter` package to easily extend support for multiple counters. Each request counter can now independently log itself at self defined intervals. This allows us to capture HTTP traffic in a variety of intervals: 10min, 30min, 24hrs. 

Why is this helpful?
According to the Container Registry [quota docs](https://cloud.google.com/container-registry/quotas), there are two measurements for inspecting traffic volume.
- 50,000 HTTP requests every 10 minutes
- 1,000,000 HTTP requests per day

This feature allows us to track both metrics for HTTP requests to GCR.

Partially satisfies #358

NONE

```release-note
Extend support for multiple request counters to log at individual intervals.
```

### DIFF
--- a/legacy/reqcounter/reqcounter.go
+++ b/legacy/reqcounter/reqcounter.go
@@ -25,70 +25,127 @@ import (
 )
 
 // RequestCounter records the number of HTTP requests to GCR.
+// TODO: @tylerferrara in the future, add a field 'persistent bool' to determine if
+// the request counter should ever reset.
 type RequestCounter struct {
-	mutex    sync.Mutex // Lock to prevent race-conditions with concurrent processes.
-	requests uint64     // Number of HTTP requests since recording started.
-	since    time.Time  // When the current counter began recording requests.
+	Mutex    sync.Mutex    // Lock to prevent race-conditions with concurrent processes.
+	Requests uint64        // Number of HTTP requests since recording started.
+	Since    time.Time     // When the current request counter began recording requests.
+	Interval time.Duration // The duration of time between each log.
 }
 
-// log records the number of HTTP requests found and resets the counter.
-func (rc *RequestCounter) log() {
-	// Hold onto the lock when reading & writing the counter.
-	rc.mutex.Lock()
-	defer rc.mutex.Unlock()
+// increment adds 1 to the request counter, signifying another call to GCR.
+func (rc *RequestCounter) Increment() {
+	rc.Mutex.Lock()
+	rc.Requests++
+	rc.Mutex.Unlock()
+}
+
+// Flush records the number of HTTP requests found and resets the request counter.
+func (rc *RequestCounter) Flush() {
+	// Hold onto the lock when reading & writing the request counter.
+	rc.Mutex.Lock()
+	defer rc.Mutex.Unlock()
 
 	// Log the number of requests within this measurement window.
-	msg := fmt.Sprintf("Since %s there have been %d requests to GCR.", counter.since.Format("2006-01-02 15:04:05"), rc.requests)
-	logrus.Debug(msg)
+	msg := fmt.Sprintf("From %s to %s [%d min] there have been %d requests to GCR.", rc.Since.Format(timestampFormat), time.Now().Format(timestampFormat), rc.Interval/time.Minute, rc.Requests)
+	Debug(msg)
 
-	// Reset the counter.
-	rc.requests = 0
-	rc.since = time.Now()
+	// Reset the request counter.
+	rc.reset()
+}
+
+// reset clears the request counter and stamps the current time of reset.
+// TODO: @tylerferrara in the future, use the request counter field 'persistent' to check
+// if it must be reset.
+func (rc *RequestCounter) reset() {
+	rc.Requests = 0
+	rc.Since = time.Now()
+}
+
+// watch continuously logs the request counter at the specified intervals.
+func (rc *RequestCounter) watch() {
+	go func() {
+		for {
+			Sleep(rc.Interval)
+			rc.Flush()
+		}
+	}()
+}
+
+// RequestCounters holds multiple request counters.
+type RequestCounters []*RequestCounter
+
+// NetworkMonitor is the primary means of monitoring network traffic between CIP and GCR.
+type NetworkMonitor struct {
+	RequestCounters RequestCounters
+}
+
+// increment adds 1 to each request counter, signifying a new request has been made to GCR.
+func (nm *NetworkMonitor) increment() {
+	for _, rc := range nm.RequestCounters {
+		rc.Increment()
+	}
+}
+
+// Log begins logging each request counter at their specified intervals.
+func (nm *NetworkMonitor) Log() {
+	for _, rc := range nm.RequestCounters {
+		rc.watch()
+	}
 }
 
 const (
-	// measurementWindow specifies the length of time to wait before logging the RequestCounter. Since Google's
+	// MeasurementWindow specifies the length of time to wait before logging the request counters. Since Google's
 	// Container Registry specifies a quota of 50,000 HTTP requests per 10 min, the window
 	// for recording requests is set to 10 min.
+	// NOTE: This metric is only a rough approximation of the actual GCR quota. The specific 10min measurement
+	// is ambiguous, as the start and end time are not specified in the docs. Therefore, it's impossible for our
+	// requests counters to perfectly line up with the actual GCR quota.
 	// Source: https://cloud.google.com/container-registry/quotas
-	measurementWindow = time.Minute * 10
+	MeasurementWindow time.Duration = time.Minute * 10
+	// timestampFormat specifies the syntax for logging time stamps of request counters.
+	timestampFormat string = "2006-01-02 15:04:05"
 )
 
 var (
-	// enableCounting will only become true if the Init function is called. This allows
+	// EnableCounting will only become true if the Init function is called. This allows
 	// requests to be counted and logged.
-	enableCounting = false
-	// counter will continuously be modified by the Increment function to count all
-	// HTTP requests to GCR.
-	counter = &RequestCounter{}
+	EnableCounting bool
+	// NetMonitor holds all request counters for recording HTTP requests to GCR.
+	NetMonitor *NetworkMonitor
+	// Debug is defined to simplify testing of logrus.Debug calls.
+	Debug func(args ...interface{}) = logrus.Debug
+	// Sleep is defined to allow mocking of the time.Sleep function in tests.
+	Sleep func(d time.Duration) = time.Sleep
 )
 
 // Init allows request counting to begin.
 func Init() {
-	enableCounting = true
-	counter = &RequestCounter{
-		mutex:    sync.Mutex{},
-		requests: 0,
-		since:    time.Now(),
+	EnableCounting = true
+
+	// Create a request counter for logging traffic every 10mins. This aims to mimic the actual
+	// GCR quota, but acts as a rough estimation of this quota, indicating when throttling may occur.
+	requestCounter := &RequestCounter{
+		Mutex:    sync.Mutex{},
+		Requests: 0,
+		Since:    time.Now(),
+		Interval: MeasurementWindow,
 	}
-	// Trigger the logger to run in the background.
-	go requestLogger(measurementWindow)
+
+	// Create a new network monitor.
+	NetMonitor = &NetworkMonitor{
+		RequestCounters: RequestCounters{requestCounter},
+	}
+
+	// Begin logging network traffic.
+	NetMonitor.Log()
 }
 
-// Increment increases the request counter by 1, signifying an HTTP
+// Increment increases the all request counters by 1, signifying an HTTP
 // request to GCR has been made.
 func Increment() {
-	if enableCounting {
-		counter.mutex.Lock()
-		counter.requests++
-		counter.mutex.Unlock()
-	}
-}
-
-// requestLogger continuously logs the number of recorded HTTP requests every interval.
-func requestLogger(interval time.Duration) {
-	for {
-		time.Sleep(interval)
-		counter.log()
+	if EnableCounting {
+		NetMonitor.increment()
 	}
 }
--- a/legacy/reqcounter/reqcounter_test.go
+++ b/legacy/reqcounter/reqcounter_test.go
@@ -1,0 +1,126 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package reqcounter_test
+
+import (
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	rc "sigs.k8s.io/k8s-container-image-promoter/legacy/reqcounter"
+)
+
+// defaultTime should be used as a timestamp for all request counters.
+// The actual time represents: September 22, 2002 at 05:03:16.
+var defaultTime, _ = time.Parse("020106 150405", "020106 150405")
+
+// NewRequestCounter returns a new request counter with the given number of requests.
+// All other object fields are set to default values.
+func NewRequestCounter(requests uint64) rc.RequestCounter {
+	return rc.RequestCounter{
+		Mutex:    sync.Mutex{},
+		Requests: requests,
+		Since:    defaultTime,
+		Interval: time.Second,
+	}
+}
+
+func TestInit(t *testing.T) {
+	rc.Init()
+	// Ensure request counting is enabled.
+	require.True(t, rc.EnableCounting, "Init did not enable counting.")
+	// Ensure request counters are created.
+	require.Greater(t, len(rc.NetMonitor.RequestCounters), 0, "Init did not create any request counters within the global Monitor.")
+	// Ensure at least one request counter uses the MeasurementWindow.
+	found := false
+	for _, requestCounter := range rc.NetMonitor.RequestCounters {
+		if requestCounter.Interval == rc.MeasurementWindow {
+			found = true
+			break
+		}
+	}
+	require.True(t, found, "No request counters are using the Interval: MeasurementWindow.")
+}
+
+func TestIncrement(t *testing.T) {
+	// Create request counters which use these request counts and
+	// populate the Monitor global variable.
+	requestCounters := []rc.RequestCounter{
+		NewRequestCounter(0),
+		NewRequestCounter(9),
+		NewRequestCounter(2839),
+	}
+	netMonitor := &rc.NetworkMonitor{
+		RequestCounters: rc.RequestCounters{
+			&requestCounters[0],
+			&requestCounters[1],
+			&requestCounters[2],
+		},
+	}
+	// Create the request counters we expect to get after calling rc.Increment.
+	expectedCounters := []rc.RequestCounter{
+		NewRequestCounter(1),
+		NewRequestCounter(10),
+		NewRequestCounter(2840),
+	}
+	expectedNetMonitor := &rc.NetworkMonitor{
+		RequestCounters: rc.RequestCounters{
+			&expectedCounters[0],
+			&expectedCounters[1],
+			&expectedCounters[2],
+		},
+	}
+	// Set the global network monitor.
+	rc.NetMonitor = netMonitor
+	// Ensure request counter modification can only occur when counting is enabled. Therefore,
+	// the global network monitor should not be mutated with this call to Increment.
+	rc.EnableCounting = false
+	rc.Increment()
+	require.EqualValues(t, netMonitor, rc.NetMonitor, "Request counters were modified while counting was disabled.")
+	// Ensure the Increment function actually increments each request counter's requests field.
+	rc.EnableCounting = true
+	rc.Increment()
+	require.EqualValues(t, expectedNetMonitor, rc.NetMonitor, "Request counters were not incremented correctly.")
+}
+
+func TestFlush(t *testing.T) {
+	// Create a local invocation of time.
+	requestCounter := NewRequestCounter(33)
+	// Mock the logrus.Debug function.
+	debugCalls := 0
+	rc.Debug = func(args ...interface{}) {
+		debugCalls++
+	}
+	requestCounter.Flush()
+	// Ensure logrus.Debug was called.
+	require.Equal(t, 1, debugCalls, "Flush() failed to trigger a debug statement.")
+	// Ensure the request counter is reset, where time advances and the requests are zeroed.
+	require.Equal(t, uint64(0), requestCounter.Requests, "Calling Flush() did not reset the request counter to 0.")
+	require.True(t, defaultTime.Before(requestCounter.Since), "Calling Flush() did not reset the request counter timestamp.")
+}
+
+func TestRequestCounterIncrement(t *testing.T) {
+	// Create a simple request counter.
+	requestCounter := NewRequestCounter(36)
+	// Create a request counter expected after Increment is called.
+	expected := NewRequestCounter(37)
+	// Increment the counter.
+	requestCounter.Increment()
+	// Ensure the request counter was incremented.
+	require.EqualValues(t, &expected, &requestCounter, "The request counter failed to increment its request field.")
+}


legacy/reqcounter/reqcounter.go
```diff
@@ -24,28 +24,71 @@ import (
...
+
+// increment adds 1 to each request counter, signifying a new request has been made to GCR.
+func (nm *networkMonitor) increment() {
+	// Hold onto the lock when writing to each counter.
+	nm.mutex.Lock()
+	defer nm.mutex.Unlock()
+
+	// Increment each counter.
+	for _, rc := range nm.counters {
+		rc.requests++
```
amwat, line 81: consider using https://gobyexample.com/atomic-counters instead 

also adding to the above comment, let's move the incrementing logic to the individual requestCounters to avoid directly trying to modify struct variables in a multi-threaded environment.

legacy/reqcounter/reqcounter.go
```diff
@@ -24,28 +24,71 @@ import (
...
+type networkMonitor struct {
+	counters []*requestCounter
+	mutex    sync.Mutex
+}
+
+// increment adds 1 to each request counter, signifying a new request has been made to GCR.
+func (nm *networkMonitor) increment() {
+	// Hold onto the lock when writing to each counter.
+	nm.mutex.Lock()
+	defer nm.mutex.Unlock()
```
amwat, line 77: do we expect multithreaded calls to `networkMonitor.increment()`? if not we can drop the mutexes here (unlike the individual request counters which do seem to need them during log() )

legacy/reqcounter/reqcounter.go
```diff
@@ -24,28 +24,71 @@ import (
...
+type requestCounter struct {
+	mutex    sync.Mutex    // Lock to prevent race-conditions with concurrent processes.
+	requests uint64        // Number of HTTP requests since recording started.
+	since    time.Time     // When the current counter began recording requests.
+	interval time.Duration // The duration of time between each log.
 }
 
 // log records the number of HTTP requests found and resets the counter.
-func (rc *RequestCounter) log() {
+func (rc *requestCounter) log() {
```
amwat, line 38: nit: consider naming this as `flush()` (which is used typically for such use cases)

legacy/reqcounter/reqcounter.go
```diff
@@ -24,28 +24,71 @@ import (
...
 	// Hold onto the lock when reading & writing the counter.
 	rc.mutex.Lock()
 	defer rc.mutex.Unlock()
 
 	// Log the number of requests within this measurement window.
-	msg := fmt.Sprintf("Since %s there have been %d requests to GCR.", counter.since.Format("2006-01-02 15:04:05"), rc.requests)
+	msg := fmt.Sprintf("Since %s there have been %d requests to GCR.", rc.since.Format("2006-01-02 15:04:05"), rc.requests)
```
listx, line 44: It would help to log the `interval` field here, so that the user doesn't have to do time math to figure out what the temporal delta is between `rc.since` and the time that this message was logged.

legacy/reqcounter/reqcounter.go
```diff
@@ -24,28 +24,71 @@ import (
...
+	for {
+		time.Sleep(rc.interval)
+		rc.log()
+	}
+}
+
+// networkMonitor is the primary means of monitoring network traffic between CIP and GCR.
+type networkMonitor struct {
+	counters []*requestCounter
+	mutex    sync.Mutex
```
listx, line 70: If we have a global mutex like this, do we still need the `mutex` field in the `requestCounter` struct?

amwat (commented):
Let's also add a few unit tests to go along with the changes?

legacy/reqcounter/reqcounter.go
```diff
@@ -24,28 +24,71 @@ import (
...
+	nm.mutex.Lock()
+	defer nm.mutex.Unlock()
+
+	// Increment each counter.
+	for _, rc := range nm.counters {
+		rc.requests++
+	}
+}
+
+// scan begins logging each counter at their specified intervals.
```
listx, line 85: I think `log()` is a better name than `scan()`, because your comment says that it logs the counters. `logCounters()`  would also work.

legacy/reqcounter/reqcounter.go
tylerferrara: In this case, adding `atomic.AddUint64(&rc.requests, 1)` will not get rid of our mutex. Since we require no modification of the counter within the `log()` function, both `increment()` and `log()` reach for the mutex when attempting any read or write to the counter.

Thank being said, I can most certainly add this, though it won't provide any benefits. Is this still worth pursuing?

legacy/reqcounter/reqcounter.go
tylerferrara: Nice catch! You are correct, this is not needed.

legacy/reqcounter/reqcounter.go
tylerferrara: Sure thing!

legacy/reqcounter/reqcounter.go
tylerferrara: On it!

legacy/reqcounter/reqcounter.go
tylerferrara: This is a mistake on my part. That mutex should not be present. Removing.

legacy/reqcounter/reqcounter.go
tylerferrara: Ack.

legacy/reqcounter/reqcounter.go
amwat: I think it's fine to use it anyway (which I see in the current version). It would avoid needing a mutex in increment() atleast. 

And also if flush weren't accessing multiple variables, we could've gotten away with atomic.LoadInt and avoid mutexes altogether.

tylerferrara: I can use the atomic.AddUnit64, but again, I still must keep the mutex in increment(). This is because we could still increment the request when flush holds onto the mutex. In order for race conditions to be avoided, all consumers must be reliant on first grabbing the mutex.

legacy/reqcounter/reqcounter.go
```diff
@@ -19,33 +19,79 @@ package reqcounter
...
+	interval time.Duration // The duration of time between each log.
 }
 
-// log records the number of HTTP requests found and resets the counter.
-func (rc *RequestCounter) log() {
+// increment adds 1 to the request counter, signifying another call to GCR.
+func (rc *requestCounter) increment() {
+
+	atomic.AddUint64(&rc.requests, 1)
+	rc.mutex.Unlock()
```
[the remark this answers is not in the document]
tylerferrara: Actually forgot to Lock. If we didn't hold the mutex, this could happen:

1. flush grabs the mutex,
2. flush reads the counter
3. increment adds +1 to the counter
4. flush resets the counter

In this scenario, because increment did not grab the counter, we lost count of an HTTP request.

legacy/reqcounter/reqcounter.go
```diff
@@ -19,33 +19,79 @@ package reqcounter
...
+	// Increment each counter.
+	for _, rc := range nm.counters {
+		rc.increment()
+	}
+}
+
+// log begins logging each counter at their specified intervals.
+func (nm *networkMonitor) log() {
+	for _, rc := range nm.counters {
+		go rc.watch()
```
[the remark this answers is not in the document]
tylerferrara: Good point!

tylerferrara:
Added unit tests to `reqcounter`. In order to test as much as possible, I exported several methods and explicitly defined the `logrus.Debug` function locally for easier mocking.

legacy/reqcounter/reqcounter.go
```diff
@@ -19,76 +19,124 @@ package reqcounter
...
 	// Hold onto the lock when reading & writing the counter.
-	rc.mutex.Lock()
-	defer rc.mutex.Unlock()
+	rc.Mutex.Lock()
+	defer rc.Mutex.Unlock()
 
 	// Log the number of requests within this measurement window.
-	msg := fmt.Sprintf("Since %s there have been %d requests to GCR.", counter.since.Format("2006-01-02 15:04:05"), rc.requests)
-	logrus.Debug(msg)
+	msg := fmt.Sprintf("As of %s, within %s min, there have been %d requests to GCR.", rc.Interval, rc.Since.Format("2006-01-02 15:04:05"), rc.Requests)
```
listx, line 52: I think you have rc.Interval and rc.Since reversed. But also, can you make this more explicit? Something like this:

"From _start_timestamp to _end_timestamp [_interval], there have been %d requests to GCR."

You'd have to do the _end_timestamp calculation but that's simple because it's just `time.Now()`. I think it's easier to read.

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+	rc "sigs.k8s.io/k8s-container-image-promoter/legacy/reqcounter"
+)
+
+func TestInit(t *testing.T) {
+	rc.Init()
+	// Ensure request counting is enabled.
+	if rc.EnableCounting == false {
+		fmt.Println("Init did not enable counting")
+		t.Fail()
```
listx, line 34: You should use the `testify/require` package as we do in `inventory_test.go`.

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,191 @@
...
+	"testing"
+	"time"
+
+	rc "sigs.k8s.io/k8s-container-image-promoter/legacy/reqcounter"
+)
+
+func TestInit(t *testing.T) {
+	rc.Init()
+	// Ensure request counting is enabled.
+	if rc.EnableCounting == false {
```
listx, line 32: Looks like you're doing a number of `if-statement` tests here to check various properties. The more succinct way to do it is to create an `expectedRc` variable of type `RequestCounter` and to just use `require.Equal(t, expected, actual)`. That way (1) you don't have to write custom error messages for every single possible failure and (2) the testing code can be much shorter and simpler, because all of the properties can be captured in an "expected" RequestCounter (or some other struct) value.

Keep the above in mind for future PRs.

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,191 @@
...
+	// Ensure at least one counter uses the MeasurementWindow
+	found := false
+	for _, counter := range rc.Monitor.Counters {
+		if counter.Interval == rc.MeasurementWindow {
+			found = true
+			break
+		}
+	}
+	if !found {
+		fmt.Println("No request counters are using the 10min Interval: MeasurementWindow")
```
listx, line 50: Hardcoding the 10min string here is not a good practice because if the quota changes you'd have to update this string and also the actual quota constant. Not DRY.

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+}
+
+func TestIncrement(t *testing.T) {
+	// Create several unique request counts.
+	requests := []uint64{0, 9, 2839}
+	// Create counters which use these request counts and
+	// populate the Monitor global variable.
+	rc.Monitor = &rc.NetworkMonitor{
+		Counters: []*rc.RequestCounter{
```
listx, line 61: It would be cleaner if there was a NewRequestCounter() function that returned a RequestCounter with the fields populated.

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,191 @@
...
+				Interval: time.Second,
+			}},
+	}
+	// Ensure that modifying counters can only occur when counting is enabled.
+	rc.EnableCounting = false
+	rc.Increment()
+	for i, counter := range rc.Monitor.Counters {
+		if counter.Requests != requests[i] {
+			fmt.Println("Counters were modified while counting was disabled")
+			t.Fail()
```
listx, line 87: Again, can you use the `testify/require` package instead of calling t.Fail()?

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+	}
+	// Ensure the Increment function actually increments each counter's requests field.
```
listx, line 90: Replace `the Increment function` with `Increment()`

legacy/reqcounter/reqcounter.go
```diff
@@ -19,76 +19,124 @@ package reqcounter
...
 	// requests to be counted and logged.
-	enableCounting = false
-	// counter will continuously be modified by the Increment function to count all
-	// HTTP requests to GCR.
-	counter = &RequestCounter{}
+	EnableCounting = false
+	// Monitor holds all counters for recording HTTP requests to GCR.
+	Monitor = &NetworkMonitor{}
+	// Debug is defined to simplify testing of logrus.Debug calls.
+	Debug = logrus.Debug
```
listx, line 112: It would be nice to have the type of `Debug` written down somewhere, so that I don't have to look up the type of `logrus.Debug()`.

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,191 @@
...
+	rc.EnableCounting = true
+	rc.Increment()
+	for i, counter := range rc.Monitor.Counters {
+		if counter.Requests != requests[i]+1 {
+			fmt.Printf("Counters were not incremented correctly. Expected: %d Got: %d\n", requests[i]+1, counter.Requests)
```
listx, line 95: If you use `testify/require`, the "expected vs got" language is unnecessary because the library already does it for you.

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,191 @@
...
+	// Flush the request counter.
+	counter.Flush()
+	// Ensure logrus.Debug was called.
+	if debugCalls != 1 {
+		fmt.Println("Flush failed to trigger a debug statement.")
+		t.Fail()
+	}
+	// Ensure the counter is reset, where time advances and the requests are zeroed.
+	counter.Flush()
+	if counter.Requests != 0 || !currentTime.Before(counter.Since) {
```
listx, line 124: You should break these two conditions into two separate if-statements, so that it's clear exactly what went wrong.

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,191 @@
...
+		fmt.Println("Calling Flush did not reset the counter.")
+		t.Fail()
+	}
+}
+
+func TestLog(t *testing.T) {
+	// Create two counters with different logging intervals. During testing, only one
+	// of the counters should fire logrus.Debug statements. To determine which counter
+	// was logged, we can use the Requests number to identify them.
+	unexpected := uint64(80808)
```
listx, line 134: Instead of unexpected and expected, use `shouldBeInvalid` and `shouldBeValid`.

I think there should be a NewRequestCounter method because it's hard to grok the values inside Counters.

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,191 @@
...
+			fmt.Println("The counter with 'Interval: 1hr' was prematurely logged after waiting only 3 seconds.")
+			t.Fail()
+		} else if strings.Contains(s, fmt.Sprint(expected)) {
+			sawExpectedLog = true
+		}
+	}
+	// Begin logging each counter.
+	netMonitor.Log()
+	// Allow enough time to ensure ONLY the expected counter can log itself.
+	time.Sleep(time.Second * 3)
```
listx, line 168: Currently the RequestCounters call time.Sleep() to pause between each interval. Instead you can give it either a real sleep function (time.Sleep()) or a fake sleep function (make one up). For production you can pass in the real sleep function, but for testing you can pass in a fake sleep function that returns instantaneously. Then we don't have to sleep 3 seconds for this single unit test (quite expensive).

In general any sort of reliance on actual real-world events (in this case, sleeping for a number of seconds) in testing code is an anti-pattern.

listx: > In general any sort of reliance on actual real-world events (in this case, sleeping for a number of seconds) in testing code is an anti-pattern.

I should clarify that this is the case for unit tests. For larger tests (e.g., integration, e2e) your scope increasingly includes the real world, so the above statement loses significance.

legacy/reqcounter/reqcounter.go
tylerferrara: Absolutely! That reads much nicer.

tylerferrara: For additional readability, intervals will be displayed in "minutes".

legacy/reqcounter/reqcounter_test.go
tylerferrara: Ack.

legacy/reqcounter/reqcounter_test.go
tylerferrara: Noted. I have reformatted some of the tests to follow this pattern.

legacy/reqcounter/reqcounter_test.go
tylerferrara: Ack.

legacy/reqcounter/reqcounter_test.go
tylerferrara: I agree. Adding this.

legacy/reqcounter/reqcounter.go
tylerferrara: Agreed. I will also explicitly define the types for all other global vars.

legacy/reqcounter/reqcounter_test.go
tylerferrara: Ack.

legacy/reqcounter/reqcounter_test.go
tylerferrara: I agree.

legacy/reqcounter/reqcounter_test.go
tylerferrara: This test has since been completely refactored. Resolving.

legacy/reqcounter/reqcounter_test.go
tylerferrara: I shortened the time to sleep, which should still be more than enough time to reliably log all counters. If this is still a concern, please bring this to my attention. Resolving.

legacy/reqcounter/reqcounter.go
```diff
@@ -19,76 +19,127 @@ package reqcounter
...
+	Since    time.Time     // When the current counter began recording requests.
+	Interval time.Duration // The duration of time between each log.
 }
 
-// log records the number of HTTP requests found and resets the counter.
-func (rc *RequestCounter) log() {
+// increment adds 1 to the request counter, signifying another call to GCR.
+func (rc *RequestCounter) Increment() {
+	rc.Mutex.Lock()
+	atomic.AddUint64(&rc.Requests, 1)
```
listx, line 41: https://stackoverflow.com/questions/47445344/is-there-a-difference-in-go-between-a-counter-using-atomic-operations-and-one-us

Using both the lock and the atomic addition seems redundant. See the post above for more info.

legacy/reqcounter/reqcounter.go
```diff
@@ -19,76 +19,127 @@ package reqcounter
...
+}
+
+// NetworkMonitor is the primary means of monitoring network traffic between CIP and GCR.
+type NetworkMonitor struct {
+	Counters []*RequestCounter
+}
+
+// increment adds 1 to each request counter, signifying a new request has been made to GCR.
+func (nm *NetworkMonitor) increment() {
+	// Increment each counter.
```
listx, line 84: Redundant comment

legacy/reqcounter/reqcounter.go
```diff
@@ -19,76 +19,127 @@ package reqcounter
...
+// Log begins logging each counter at their specified intervals.
+func (nm *NetworkMonitor) Log() {
+	for _, rc := range nm.Counters {
+		rc.watch()
+	}
 }
 
 const (
-	// measurementWindow specifies the length of time to wait before logging the RequestCounter. Since Google's
+	// MeasurementWindow specifies the length of time to wait before logging the RequestCounter. Since Google's
```
listx, line 98: Please add a clarification here that we are not measuring the same 10 minute window that GCR may be measuring. It's a rough approximation.

It may be that GCR measures a sliding 10 minute window every minute. The docs at https://cloud.google.com/container-registry/quotas don't say either way, so we have to live with this ambiguity. Because of this ambiguity, we have to call it out in our own code.

You could also reach out to the GCR team to clear this up (they work at Google ;)

legacy/reqcounter/reqcounter.go
```diff
@@ -19,76 +19,127 @@ package reqcounter
...
 	// Container Registry specifies a quota of 50,000 HTTP requests per 10 min, the window
 	// for recording requests is set to 10 min.
 	// Source: https://cloud.google.com/container-registry/quotas
-	measurementWindow = time.Minute * 10
+	MeasurementWindow = time.Minute * 10
+	// loggingFormat specifies the syntax for logging time stamps of request counters.
+	loggingFormat = "2006-01-02 15:04:05"
```
listx, line 104: A more precise name would be `timestampFormat`.

legacy/reqcounter/reqcounter.go
```diff
@@ -19,76 +19,127 @@ package reqcounter
...
 	// requests to be counted and logged.
-	enableCounting = false
-	// counter will continuously be modified by the Increment function to count all
-	// HTTP requests to GCR.
-	counter = &RequestCounter{}
+	EnableCounting bool
+	// Monitor holds all counters for recording HTTP requests to GCR.
+	Monitor *NetworkMonitor
+	// Debug is defined to simplify testing of logrus.Debug calls.
+	// type: func(args ...interface{})
```
listx, line 114: Please instead instead define the type with the `type` keyword, not as a comment.

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,191 @@
...
+		// Ensure race conditions are avoided when manipulating the collection.
+		mutex.Lock()
+		defer mutex.Unlock()
+		// Add the interval to represent the counter to the collection.
+		collection = append(collection, interval)
+	}
+	// Begin logging each counter.
+	netMonitor.Log()
+	// Allow enough time to ensure ONLY the expected counters can log themselves.
+	time.Sleep(time.Second * 2)
```
listx, line 177: Again, please try to use a fake Sleep function. This is like dependency injection.

legacy/reqcounter/reqcounter.go
tylerferrara: I forgot that was even possible! Fixing.

legacy/reqcounter/reqcounter.go
tylerferrara: Ack.

legacy/reqcounter/reqcounter.go
tylerferrara: Ack.

legacy/reqcounter/reqcounter.go
tylerferrara: Ack.

legacy/reqcounter/reqcounter.go
tylerferrara: Taking a second look, this does not appear to be a use case for atomic operations. This is because all manipulations (reads or writes) of RequestCounter must first hold the object's mutex.

We can't use atomic operations here because of the Flush function. It attempts to:
- Read requests
- Set requests = 0
- Set since = current time

The Flush function must do all of these things at once, ensuring no function can interject and (read or write) the object during execution. That is why holding a single mutex is the way to go. We should not conduct all of these operations atomically, because they MUST be grouped together.

legacy/reqcounter/reqcounter_test.go
listx: Maybe having a fake Sleep can be part of a later design. It might complicate this area too much for this initial PR. Can you record this thought somewhere?

tylerferrara: Added "Test the Log function with fake sleep function" to issue: #358
This will be explored in another PR.

legacy/reqcounter/reqcounter.go
```diff
@@ -25,70 +25,123 @@ import (
...
+		for {
+			Sleep(rc.Interval)
+			rc.Flush()
+		}
+	}()
+}
+
+// NetworkMonitor is the primary means of monitoring network traffic between CIP and GCR.
+type NetworkMonitor struct {
+	Counters []*RequestCounter
```
listx, line 78: When naming fields inside structs, I usually try to preserve the underlying type name. So here I think nothing stops you from using `RequestCounters` as the field name instead of `Counters`.

legacy/reqcounter/reqcounter.go
```diff
@@ -25,70 +25,123 @@ import (
...
-	// enableCounting will only become true if the Init function is called. This allows
+	// EnableCounting will only become true if the Init function is called. This allows
 	// requests to be counted and logged.
-	enableCounting = false
-	// counter will continuously be modified by the Increment function to count all
-	// HTTP requests to GCR.
-	counter = &RequestCounter{}
+	EnableCounting bool
+	// Monitor holds all counters for recording HTTP requests to GCR.
+	Monitor *NetworkMonitor
```
listx, line 112: Can you name this variable NetworkMonitor? For the same reason to rename `Counters` -> `RequestCounters` in my earlier comment.

legacy/reqcounter/reqcounter.go
```diff
@@ -25,70 +25,123 @@ import (
...
+	}
 }
 
 const (
-	// measurementWindow specifies the length of time to wait before logging the RequestCounter. Since Google's
+	// MeasurementWindow specifies the length of time to wait before logging the RequestCounter. Since Google's
 	// Container Registry specifies a quota of 50,000 HTTP requests per 10 min, the window
 	// for recording requests is set to 10 min.
+	// NOTE: This metric is only a rough approximation of the actual GCR quota. The specific 10min measurement
+	// is ambiguous, and therefore difficult to mimic correctly.
```
listx, line 100: Please explain why it's ambiguous, so that the reader does not have to guess for themselves.

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,127 @@
...
+			break
+		}
+	}
+	require.True(t, found, "No request counters are using the Interval: MeasurementWindow.")
+}
+
+func TestIncrement(t *testing.T) {
+	// Create counters which use these request counts and
+	// populate the Monitor global variable.
+	counters := []rc.RequestCounter{
```
listx, line 63: If you do

```
type RequestCounters []RequestCounter
```

in the reqcounter package, it would make this line a little easier to read, because it would become

```
counters := rc.RequestCounters{
```
```

it would make this a

File: legacy/reqcounter/reqcounter.go [hunk 31]
```diff
@@ -25,70 +25,123 @@ import (
...
-	enableCounting = true
-	counter = &RequestCounter{
-		mutex:    sync.Mutex{},
-		requests: 0,
-		since:    time.Now(),
+	EnableCounting = true
+
+	// Create a request counter for logging traffic every 10mins. This aims to mimic the actual
+	// GCR quota, but acts as a rough estimation of this quota, indicating when throttling may occur.
+	counter := &RequestCounter{
```
listx, line 125: Rename `counter` -> `requestCounter`

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+func TestFlush(t *testing.T) {
+	// Create a local invocation of time.
+	counter := NewRequestCounter(33)
+	// Mock the logrus.Debug function.
+	debugCalls := 0
+	rc.Debug = func(args ...interface{}) {
+		debugCalls++
+	}
+	// Flush the request counter.
```
listx, line 109: This comment doesn't really add value; please remove.

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,127 @@
...
+	require.EqualValues(t, netMonitor, rc.Monitor, "Counters were modified while counting was disabled.")
+	// Ensure the Increment function actually increments each counter's requests field.
+	rc.EnableCounting = true
+	rc.Increment()
+	require.EqualValues(t, expectedNetMonitor, rc.Monitor, "All counters were not incremented correctly.")
+}
+
+func TestFlush(t *testing.T) {
+	// Create a local invocation of time.
+	counter := NewRequestCounter(33)
```
listx, line 103: Rename `counter` to `requestCounter`

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,127 @@
...
+	// Ensure logrus.Debug was called.
+	require.Equal(t, 1, debugCalls, "Flush failed to trigger a debug statement.")
+	// Ensure the counter is reset, where time advances and the requests are zeroed.
+	require.Equal(t, uint64(0), counter.Requests, "Calling Flush did not reset the request counter to 0.")
+	require.True(t, defaultTime.Before(counter.Since), "Calling Flush did not reset the counter timestamp.")
+}
+
+func TestRequestCounterIncrement(t *testing.T) {
+	// Create a simple request counter.
+	counter := NewRequestCounter(36)
```
listx, line 120: Rename `counter` to `requestCounter`

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,127 @@
...
+	// Create a request counter expected after Increment is called.
+	expects := NewRequestCounter(37)
```
listx, line 122: Rename `expects` to `expected`

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,127 @@
...
+	// Increment the counter.
+	counter.Increment()
+	// Ensure the counter was incremented.
+	require.EqualValues(t, &expects, &counter, "The request counter failed to increment it's request field.")
```
listx, line 126: typo `it's` -> `its`

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,127 @@
...
+	counter := NewRequestCounter(33)
+	// Mock the logrus.Debug function.
+	debugCalls := 0
+	rc.Debug = func(args ...interface{}) {
+		debugCalls++
+	}
+	// Flush the request counter.
+	counter.Flush()
+	// Ensure logrus.Debug was called.
+	require.Equal(t, 1, debugCalls, "Flush failed to trigger a debug statement.")
```
listx, line 112: `Flush` -> `Flush()`

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,127 @@
...
+	rc.Monitor = netMonitor
+	// Ensure counter modification can only occur when counting is enabled. Therefore,
+	// the global network monitor should not be mutated with this call to Increment.
+	rc.EnableCounting = false
+	rc.Increment()
+	require.EqualValues(t, netMonitor, rc.Monitor, "Counters were modified while counting was disabled.")
+	// Ensure the Increment function actually increments each counter's requests field.
+	rc.EnableCounting = true
+	rc.Increment()
+	require.EqualValues(t, expectedNetMonitor, rc.Monitor, "All counters were not incremented correctly.")
```
listx, line 98: Reword `All counters` -> `Request counters` because if the Increment() function misbehaves, it may be that some counters were incremented, just not all of them.

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,127 @@
...
+			&expectedCounters[2],
+		},
+	}
+	// Set the global network monitor.
+	rc.Monitor = netMonitor
+	// Ensure counter modification can only occur when counting is enabled. Therefore,
+	// the global network monitor should not be mutated with this call to Increment.
+	rc.EnableCounting = false
+	rc.Increment()
+	require.EqualValues(t, netMonitor, rc.Monitor, "Counters were modified while counting was disabled.")
```
listx, line 94: Reword `Counters` -> `Request counters`

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,127 @@
...
+	rc.Debug = func(args ...interface{}) {
+		debugCalls++
+	}
+	// Flush the request counter.
+	counter.Flush()
+	// Ensure logrus.Debug was called.
+	require.Equal(t, 1, debugCalls, "Flush failed to trigger a debug statement.")
+	// Ensure the counter is reset, where time advances and the requests are zeroed.
+	require.Equal(t, uint64(0), counter.Requests, "Calling Flush did not reset the request counter to 0.")
+	require.True(t, defaultTime.Before(counter.Since), "Calling Flush did not reset the counter timestamp.")
```
listx, line 115: Reword `counters` -> `request counters`

legacy/reqcounter/reqcounter_test.go
```diff
@@ -0,0 +1,127 @@
...
+		NewRequestCounter(2839),
+	}
+	netMonitor := &rc.NetworkMonitor{
+		Counters: []*rc.RequestCounter{
+			&counters[0],
+			&counters[1],
+			&counters[2],
+		},
+	}
+	// Create the monitor counter we expect to get after calling rc.Increment.
```
listx, line 75: Reword `monitor counter` to something else --- it's a bit confusing because `monitor counter` is not a phrase used anywhere else I think.

listx (commented):
In the code, we should use the word `RequestCounter` or `requestCounter` consistently, everywhere.

In the code's comments, we should use the phrase `request counter` consistenly, everywhere.

legacy/reqcounter/reqcounter.go
tylerferrara: I can't actually do this, that would be considered a re-definition. I also must keep it public (first letter capital). So I could do NetMonitor??

legacy/reqcounter/reqcounter.go
tylerferrara: ack.

legacy/reqcounter/reqcounter.go
tylerferrara: ack.

legacy/reqcounter/reqcounter_test.go
tylerferrara: I still can't remove the local definition of each counter. This is because the counters must be passed as pointers. Regardless, I will still make this change.

listx:
Thanks!

/lgtm
/approve

PR #369, opened 2021-07-28
[WIP] Creating simple faketime package for time.Sleep mocking

tylerferrara:

```release-note

```

### DIFF
--- a/legacy/reqcounter/reqcounter_test.go
+++ b/legacy/reqcounter/reqcounter_test.go
@@ -17,11 +17,13 @@ limitations under the License.
 package reqcounter_test
 
 import (
+	"fmt"
 	"sync"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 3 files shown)


legacy/reqcounter/reqcounter_test.go
```diff
@@ -124,3 +126,48 @@ func TestRequestCounterIncrement(t *testing.T) {
...
+		},
+	}
+	// Mock the Debug function, capturing each statement.
+	logged := []string{}
+	rc.Debug = func(args ...interface{}) {
+		logged = append(logged, fmt.Sprint(args[0]))
+	}
+	// Mock the global clock with fake time.
+	fakeTime := &ft.FakeTime{}
+	timeStep := time.Second * 2
```
listx, line 155: Just to prove that this fake function actually works like we expect it to, can you change this to be a 1 hour interval instead of 2 seconds?

tylerferrara: This PR is not ready for review. It was only pushed for proof of concept.

PR #370, opened 2021-07-28
Verify the trigger and format of request counter logs

tylerferrara:
#### What type of PR is this?
/kind feature

This change uses the newly added `timewrapper` package for creation of a new test: `TestLog()`. This test simply ensures all request counters within a Network Monitor will log correctly. It does NOT test the frequency or order of the logs, as doing so requires much more complicated synchronizations with goroutines. In the future, request counters can individually verify their logging frequency by explicitly testing the `watch()` function.
Part of #358

NONE

```release-note
Creating TestLog() to verify the triggering and format of request counter logs.
```

### DIFF
--- a/legacy/reqcounter/reqcounter_test.go
+++ b/legacy/reqcounter/reqcounter_test.go
@@ -17,12 +17,14 @@ limitations under the License.
 package reqcounter_test
 
 import (
+	"fmt"
 	"sync"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/require"
 	rc "sigs.k8s.io/k8s-container-image-promoter/legacy/reqcounter"
+	tw "sigs.k8s.io/k8s-container-image-promoter/legacy/timewrapper"
 )
 
 // defaultTime should be used as a timestamp for all request counters.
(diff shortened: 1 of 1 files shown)


legacy/reqcounter/reqcounter_test.go
```diff
@@ -124,3 +126,56 @@ func TestRequestCounterIncrement(t *testing.T) {
...
+		NewRequestCounter(2839),
+	}
+	netMonitor := &rc.NetworkMonitor{
+		RequestCounters: rc.RequestCounters{
+			&requestCounters[0],
+			&requestCounters[1],
+			&requestCounters[2],
+		},
+	}
+	// Mock time with a fake global time that remains constent. This avoids testing the order in which
```
listx, line 145: typo: constent

legacy/reqcounter/reqcounter_test.go
```diff
@@ -124,3 +126,56 @@ func TestRequestCounterIncrement(t *testing.T) {
...
+	// request counters are logged. For simplicity, request counters can individually verify
+	// self logging occurs at the expected interval. This circumvents the complicated task of
```
listx, line 147: `self-logging` ?

legacy/reqcounter/reqcounter_test.go
```diff
@@ -124,3 +126,56 @@ func TestRequestCounterIncrement(t *testing.T) {
...
+		fmt.Sprintf("From %s to %s [%d min] there have been %d requests to GCR.", stableTime, stableTime, requestCounters[0].Interval/time.Minute, requestCounters[0].Requests): true,
+		fmt.Sprintf("From %s to %s [%d min] there have been %d requests to GCR.", stableTime, stableTime, requestCounters[1].Interval/time.Minute, requestCounters[1].Requests): true,
+		fmt.Sprintf("From %s to %s [%d min] there have been %d requests to GCR.", stableTime, stableTime, requestCounters[2].Interval/time.Minute, requestCounters[2].Requests): true,
+	}
+	// Log all request counters.
+	netMonitor.Log()
+	// Wait for all goroutines to finish their Debug calls.
+	wg.Wait()
+	// Ensure all request counters were logged and formed correctly.
+	require.EqualValues(t, expected, logged, "All request counters did not log correctly.")
```
listx, line 180: `All` could be a lie if just one of the counters doesn't match what is expected. Just `Request counter(s) did not log correctly.` is sufficient.

legacy/reqcounter/reqcounter_test.go
tylerferrara: Fixed.

legacy/reqcounter/reqcounter_test.go
tylerferrara: Re-worded.

legacy/reqcounter/reqcounter_test.go
tylerferrara: Ack.

legacy/reqcounter/reqcounter_test.go
```diff
@@ -124,3 +126,57 @@ func TestRequestCounterIncrement(t *testing.T) {
...
+		if resetLog == msg {
+			// Block the current goroutine.
+			select {}
+		}
+		logged[msg] = true
+		wg.Done()
+	}
+	// Define the expected logs.
+	expected := map[string]bool{
+		fmt.Sprintf("From %s to %s [%d min] there have been %d requests to GCR.", stableTime, stableTime, requestCounters[0].Interval/time.Minute, requestCounters[0].Requests): true,
```
amwat, line 172: here and elsewhere all "expected" strings should be literals.

don't use requestCounters[0].Requests use the literal value that you expect to see.

legacy/reqcounter/reqcounter_test.go
```diff
@@ -124,3 +126,57 @@ func TestRequestCounterIncrement(t *testing.T) {
...
+			&requestCounters[0],
+			&requestCounters[1],
+			&requestCounters[2],
+		},
+	}
+	// Mock time with a fake global time that remains constant. This avoids testing the order in which
+	// request counters are logged. For simplicity, additional tests may check whether a single request
+	// counter calls Debug() the expected number of times as fake time advances forward. This circumvents
+	// the task of concurrently communicating time-steps with multiple goroutines, demanding a
+	// complicated sleep and time-step synchronization.
```
amwat, line 149: I think we still want to mock sleep otherwise we aren't testing the intended behavior.

I forget why modifying sleep to just do an infinite loop checking on currentFakeTime - rc.Since was running into issues?

legacy/reqcounter/reqcounter_test.go
```diff
@@ -124,3 +126,57 @@ func TestRequestCounterIncrement(t *testing.T) {
 	// Ensure the request counter was incremented.
 	require.EqualValues(t, &expected, &requestCounter, "The request counter failed to increment its request field.")
 }
+
+func TestLog(t *testing.T) {
```
amwat, line 130: As we discussed, this would more bucket under an integration test so it's important to test the various things we want to happen

1. Initialize the requestCounters
2. "Advance time"
3. monitor.Increment a bunch of times
4. repeat 2,3 a bunch of times
5. Check the log statements of each counter (length, order and specific values)
(we don't care about relative ordering across counters but we do want to check the order in which logs are generated by a single counter)

amwat (commented):
Added some comments on how the test should look like overall to add more clarity.
we should't need any wait groups as part of the test. That should propbably be done as part of the actual code by having a monitor.StopWatching().

For that let's modify how we are doing watch(). instead of a raw infinite loop it should have a channel that let's us stop watching to avoid potential memory leaks.

https://dave.cheney.net/2016/12/22/never-start-a-goroutine-without-knowing-how-it-will-stop

legacy/reqcounter/reqcounter_test.go
tylerferrara: The issue with this is that all infinite loops do not synchronize together, as they are independent goroutines. This means, we can't rely on each counter to respond to time changes. The only way to do this is to use wait groups to stall the main reqcounter_test.go thread until all goroutines have "responded" to their new times.

legacy/reqcounter/reqcounter_test.go
amwat: Orthogonally, let's also reconsider if we are actually getting the required signal from this test if we are mocking/synchronizing a bunch of unexpected behavior just to make the integration test work.

It's okay if we can cover the individual functionality by way of focused unit tests.

legacy/reqcounter/reqcounter_test.go
amwat: ACK

tylerferrara:
After discussion with @amwat, it has been decided to abandon directly testing NetMonitor.Log() in favor of directly testing a resource counter's watch function. This will simplify the testing strategy to avoid independent goroutine synchronizations.
/close

listx:
> After discussion with @amwat, it has been decided to abandon directly testing NetMonitor.Log() in favor of directly testing a resource counter's watch function. This will simplify the testing strategy to avoid independent goroutine synchronizations.
> /close

Makes sense. +1

PR #372, opened 2021-07-28
[WIP] Added TestWatch to ensure request counters log at regular intervals.

tylerferrara:
#### What type of PR is this?
/kind feature

This change creates a new test for `reqcounter`, ensuring every request counter will sleep and log at regular intervals. Testing this behavior requires the modified `FakeTime` wrapper to advance the global fake time in a series of time-steps. A simple request counter  is used to log new requests upon every time advancement. Since `FakeTime` makes time advancements in a series of intervals, the request counter can wake, log, and sleep several times in one `fakeTime.Advance()` call. This simulates real world behavior, ensuring the correct logs are produced on every advancement.

Partially satisfies #358 

Blocked by the merging of #373 
Supersedes #372
NONE

```release-note
Added TestWatch to ensure request counters log at regular intervals.
```

### DIFF
--- a/legacy/timewrapper/timewrapper.go
+++ b/legacy/timewrapper/timewrapper.go
@@ -35,13 +35,81 @@ func (RealTime) Now() time.Time { return time.Now() }
 // Sleep simply calls time.Sleep(d), using the given duration.
 func (RealTime) Sleep(d time.Duration) { time.Sleep(d) }
 
-// FakeTime holds the global fake time.
+// FakeTime is an advancing clock, supporting only a single sleeping goroutine.
 type FakeTime struct {
-	Time time.Time
+	Time      time.Time
+	Interval  time.Duration
+	CheckTime chan bool
+	Updated   chan bool
+}
+
+// NewFakeTime creates a new fake time with the provided start time.
+func NewFakeTime(startTime time.Time) FakeTime {
+	return FakeTime{
+		Time:      startTime,
+		CheckTime: make(chan bool),
+		Updated:   make(chan bool),
+	}
+}
+
+// Advance adds the duration to the global fake time by breaking the
+// duration up into a series of time-steps. Each time-step is either less than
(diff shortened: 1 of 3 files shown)


legacy/timewrapper/timewrapper.go
```diff
@@ -37,11 +38,75 @@ func (RealTime) Sleep(d time.Duration) { time.Sleep(d) }
...
+
+func NewFakeTime(startTime time.Time) FakeTime {
+	return FakeTime{
+		Time:      startTime,
+		CheckTime: make(chan bool),
+		Updated:   make(chan bool),
+	}
+}
+
+// Advance adds the duration to the global fake time by braking the
```
listx, line 55: braking -> breaking

legacy/timewrapper/timewrapper.go
```diff
@@ -37,11 +38,75 @@ func (RealTime) Sleep(d time.Duration) { time.Sleep(d) }
...
+// duration up into a series of time steps. Each time step is either a full
+// Interval or remainder of the
```
listx, line 57: Unfinished sentence

legacy/timewrapper/timewrapper.go
listx: But also, why does fake time have to be global?

listx (commented):
I don't understand why we need to block goroutines and such.

If fake time's sleep() returns instantly, then somewhere inside the sleep() (before just before we instantly return) we have to send in some fake HTTP requests to simulate them happening, and then afterwards just check that we did log the right number of requests (our RequestCounter should be accurate).

legacy/timewrapper/timewrapper.go
tylerferrara: Resolved.

tylerferrara:
This PR is no longer needed, as #376 covers this behavior in a much simpler manner.

PR #368, opened 2021-07-27
[ARCHIVE] Added a faketime package to mock time.Sleep()

tylerferrara:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
This change addresses the difficulty with testing timed events within the `reqcounter` package. More specifically, this PR looks to mock the `time.Sleep` function as requested by [this comment](https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/364#discussion_r675293201).

The newly created `faketime` package, enables such deterministic testing of timed events. Independent goroutines can call `faketime.Sleep(d time.Duration)` in place of `time.Sleep(d time.Duration)` blocking their execution for a fixed duration. At the same time, reqcounter_test.go can concurrently advance time by a specific duration, unblocking the goroutines that have expired their sleep timers.

#### Which issue(s) this PR fixes:
Related to #358 

#### Special notes for your reviewer:
An example of implementing `fake_time` into reqcounter_test.go would look something like this:
```go
// Mock the sleep function
reqcounter.Sleep = faketime.Sleep
faketime.Init()
// Mock the debug function
debugCalls := 0
expectedDebugCalls
reqcounter.Debug = func(args ...interface{}) {
  debugCalls++
}
// Create a few request counters.
...
// Start logging.
netMonitor.Log()
// Advance time by 30sec.
faketime.Advance(time.Second * 30)
// Ensure the correct number of counters were fired.
request.Equal(t, expectedDebugCalls, debugCalls, "The incorrect number of Debug() calls were made.")
// Stop all blocking goroutines.
faketime.Close()
```

#### Does this PR introduce a user-facing change?
NONE

```release-note
Added a fake_time package to mock time.Sleep
```
cc: @listx @amwat @justaugustus @kubernetes-sigs/release-engineering

PR #376, opened 2021-07-29
Abstract a request counter's cycle logic from watch()

tylerferrara:
#### What type of PR is this?
/kind feature
/kind design

This change separates sleep/log cycles from the `watch()` function within `reqcounter`. Abstracting this repeated logic from watch's anonymous goroutine allows for simple deterministic testing of sleep/log behavior. Now there is no need to `TestWatch` (#372), as it simply executes Cycle() indefinitely. 

Partially satisfies #358 

This approach is much more favorable than previous attempts to test the logging of request counters. Since request counters spawn goroutines in `watch()`, advancing time forward involves complicated synchronization patterns to ensure a sleep/log cycle has finished. Instead, if we only test sleep/log behavior, we avoid unwanted complexity while still resulting in equivalent code coverage. 

Supersedes: #372 #368 #369
NONE

```release-note
Abstract sleep/log logic from a request counter's watch() function.
```

### DIFF
--- a/legacy/reqcounter/reqcounter_test.go
+++ b/legacy/reqcounter/reqcounter_test.go
@@ -17,12 +17,14 @@ limitations under the License.
 package reqcounter_test
 
 import (
+	"fmt"
 	"sync"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/require"
 	rc "sigs.k8s.io/k8s-container-image-promoter/legacy/reqcounter"
+	tw "sigs.k8s.io/k8s-container-image-promoter/legacy/timewrapper"
 )
 
 // defaultTime should be used as a timestamp for all request counters.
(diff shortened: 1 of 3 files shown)


legacy/reqcounter/reqcounter_test.go
```diff
@@ -124,3 +126,35 @@ func TestRequestCounterIncrement(t *testing.T) {
...
+	// Determine the expected logs.
+	expected := []string{
+		"From 2006-01-02 15:04:05 to 2006-01-02 15:04:08 [0 min] there have been 82 requests to GCR.",
+		"From 2006-01-02 15:04:08 to 2006-01-02 15:04:11 [0 min] there have been 0 requests to GCR.",
+		"From 2006-01-02 15:04:11 to 2006-01-02 15:04:14 [0 min] there have been 0 requests to GCR.",
+		"From 2006-01-02 15:04:14 to 2006-01-02 15:04:17 [0 min] there have been 0 requests to GCR.",
+		"From 2006-01-02 15:04:17 to 2006-01-02 15:04:20 [0 min] there have been 0 requests to GCR.",
+	}
+	// Repeatedly run sleep/log cycles.
+	numCycles := 5
```
listx, line 154: This should be `numCycles := len(expected)`, right?

listx: You could even simplify this to generate the expected log values programmatically, given just the number of cycles as the input. This seems simpler than manually writing out the log statements as we're doing now in the `expected` variable. E.g, you could test 100 cycles or 1000 cycles easily if you just generate the `expected` value as part of the test setup before executing the test.

legacy/reqcounter/reqcounter_test.go
```diff
@@ -124,3 +126,35 @@ func TestRequestCounterIncrement(t *testing.T) {
 	// Ensure the request counter was incremented.
 	require.EqualValues(t, &expected, &requestCounter, "The request counter failed to increment its request field.")
 }
+
+func TestCycle(t *testing.T) {
+	// Create a simple request counter expected to log every 3 seconds.
+	requestCounter := NewRequestCounter(82)
+	requestCounter.Interval = time.Second * 3
```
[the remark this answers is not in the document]
tylerferrara: Sure thing!

legacy/reqcounter/reqcounter_test.go
tylerferrara: @amwat suggested in an earlier PR to explicitly define string literals for expected behavior. It would be nice to generate expected logs, but I think it's more stable & easier to read if we explicitly define them.

tylerferrara: On second thought, this may be advantageous. We should explore this further in #377

listx: Right, so, you could keep this explicit log statement as-is, but for testing multiple test cases of different cycles you could do them programmatically.

listx:
Thanks!

/hold cancel
/lgtm
/approve

PR #380, opened 2021-07-30
[WIP] Replace all logrus.Debug calls with logrus.Info

tylerferrara:
#### What type of PR is this?
/kind cleanup

#### What this PR does / why we need it:
This change removes all logrus.Debug() calls in favor of logrus.Info().

#### Special notes for your reviewer:
Debug statements don't appear to be showing in GCP logs.
This is an attempt to standardize logging statements and show extra information when running `cip audit --verbose`

#### Does this PR introduce a user-facing change?

```release-note
Replace all logrus.Debug calls with logrus.Info
```

PR #381, opened 2021-07-30
Enable verbose log lvl when passing --verbose flag

tylerferrara:
#### What type of PR is this?
/kind bug

This change fixes missing `logrus.Debug()` statements by explicitly setting the logging level. Now, when passing the `--verbose` flag, we should be able to actually see verbose & request counter logs.

Additionally, since the default logging level hides debug statement by default, we no longer need a serviceContext.Debug() wrapper. It's safe to define logrus.Debug, as only --verbose will trigger the log level change.

Supersedes #380 
Partially satisfies #358

Currently, when deploying the auditor with `--verbose`, we were unable to see any Debug logs. This PR should fix that problem.

```release-note
Fix logging level when passing --verbose flag.
```

### DIFF
--- a/legacy/cli/audit.go
+++ b/legacy/cli/audit.go
@@ -49,13 +49,14 @@ func RunAuditCmd(opts *AuditOptions) error {
 		opts.RepoBranch,
 		opts.ManifestPath,
 		opts.UUID,
-		opts.Verbose,
 	)
 	if err != nil {
 		return errors.Wrap(err, "creating auditor context")
 	}
 
 	if opts.Verbose {
+		// Enable verbose logging.
+		logrus.SetLevel(logrus.DebugLevel)
 		// Initialize global counter to track the number of HTTP requests made to GCR.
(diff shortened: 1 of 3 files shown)


legacy/cli/audit.go
```diff
@@ -56,6 +56,8 @@ func RunAuditCmd(opts *AuditOptions) error {
 	}
 
 	if opts.Verbose {
+		// Enable verbose logging.
+		logrus.SetLevel(logrus.DebugLevel)
```
listx, line 60: We should now also probably delete the useless Debug wrapper we have around logrus.Debug (because the if-check is redundant now).

tylerferrara: In that case, we don't even need [VerboseLogging](https://github.com/kubernetes-sigs/k8s-container-image-promoter/blob/master/legacy/audit/types.go#L46) in ServerContext.

listx:
Thanks!

/lgtm
/approve

PR #382, opened 2021-07-30
Add support for persistent request counters

tylerferrara:
#### What type of PR is this?
/kind feature

This change adds the option to enable persistence when creating request counters. When enabled, the number of requests & timestamp will never reset. This acts as an indefinite log of all HTTP requests the auditor has seen since startup.

Part of #358 

NONE

```release-note
Adding support for persistent request counters which do not reset.
```

### DIFF
--- a/legacy/reqcounter/reqcounter.go
+++ b/legacy/reqcounter/reqcounter.go
@@ -26,13 +26,12 @@ import (
 )
 
 // RequestCounter records the number of HTTP requests to GCR.
-// TODO: @tylerferrara in the future, add a field 'persistent bool' to determine if
-// the request counter should ever reset.
 type RequestCounter struct {
-	Mutex    sync.Mutex    // Lock to prevent race-conditions with concurrent processes.
-	Requests uint64        // Number of HTTP requests since recording started.
-	Since    time.Time     // When the current request counter began recording requests.
-	Interval time.Duration // The duration of time between each log.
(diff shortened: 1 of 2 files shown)


legacy/reqcounter/reqcounter.go
```diff
@@ -29,10 +29,11 @@ import (
...
 type RequestCounter struct {
-	Mutex    sync.Mutex    // Lock to prevent race-conditions with concurrent processes.
-	Requests uint64        // Number of HTTP requests since recording started.
-	Since    time.Time     // When the current request counter began recording requests.
-	Interval time.Duration // The duration of time between each log.
+	Mutex      sync.Mutex    // Lock to prevent race-conditions with concurrent processes.
+	Requests   uint64        // Number of HTTP requests since recording started.
+	Since      time.Time     // When the current request counter began recording requests.
+	Interval   time.Duration // The duration of time between each log.
+	Persistent bool          // Determine if the request counter should ever reset.
```
listx, line 36: Nit: I think the comment is easier to read if you just tell the reader what will happen. E.g.,

```
// If true, the Requests field is never set back to 0 when reset() is called.
```

listx: Actually, maybe `Resettable` is a better name? It's more to the point and is almost self-explanatory.

tylerferrara: Ack.

listx:
Thanks!

/lgtm
/approve

Issue #353, opened 2021-07-19
Auditor not querying GCR


tylerferrara:
#### What happened:
The GCP project: `k8s-artifacts-prod` does not appear to show it's GCR API being requested. This is unexpected behavior, as the auditor should be continuously verifying new GCR state changes. When it sees a new image that is not inside k/k8s.io, it consults the GCR to verify this image. However, since there hasn't been ANY queries to GCR in 30+ days, the auditor must not be running correctly.

Related to k/k8s.io issue: [#2364](https://github.com/kubernetes/k8s.io/issues/2364)

#### What you expected to happen:
The API service should show some amount of requests for GCR.

#### How to reproduce it (as minimally and precisely as possible):
N/A 

#### Anything else we need to know?:
![pantheon corp google com_apis_api_containerregistry googleapis com_quotas_project=k8s-artifacts-prod pageState=(%22duration%22_(%22groupValue%22_%22P30D%22,%22customValue%22_null))](https://user-images.githubusercontent.com/20875599/126229173-763d277a-3883-4b3c-8718-a10ce82b062a.png)
![pantheon corp google com_iam-admin_quotas_project=k8s-artifacts-prod pageState=(%22allQuotasTable%22_(%22f%22_%22%255B%257B_22k_22_3A_22Service_22_2C_22t_22_3A10_2C_22v_22_3A_22_5C_22Container%2520Registry%2520API_5C_22_22_2C_22s_22_3At (2)](https://user-images.githubusercontent.com/20875599/126229289-c6fb0321-91a6-4dc7-874e-94f99ef053c3.png)

#### Environment:

- Cloud provider or hardware configuration: GCP
cc: @listx @amwat @justaugustus @kubernetes-sigs/release-engineering

listx:
Thinking about this again, I think it's because the auditor queries the staging GCRs, not prod. See https://github.com/kubernetes-sigs/k8s-container-image-promoter/blob/bed8661305e1e325d0dd22b61e0cea2967b0f0e4/legacy/audit/auditor.go#L261-L266. The `subproject` here means the *staging* repo that belongs to the subproject, not production.

tylerferrara:
[...]

PR #384, opened 2021-07-30
Adding request counter threshold

tylerferrara:
#### What type of PR is this?
/kind feature

This change introduces a threshold for resettable request counters. The request counter will fire a logrus.Warn statement if it observes it's requests exceed that threshold. Non-resettable request counters will explicitly ignore any thresholds, as they indefinitely increment. Now, we can easily query GCP logs for these warning messages to determine if network usage is high.

Part of #353 #358 

NONE

```release-note
Create request counter thresholds which warn when number of requests have been exceeded. 
```

### DIFF
--- a/legacy/reqcounter/reqcounter.go
+++ b/legacy/reqcounter/reqcounter.go
@@ -27,11 +27,13 @@ import (
 
 // RequestCounter records the number of HTTP requests to GCR.
 type RequestCounter struct {
-	Mutex      sync.Mutex    // Lock to prevent race-conditions with concurrent processes.
-	Requests   uint64        // Number of HTTP requests since recording started.
-	Since      time.Time     // When the current request counter began recording requests.
-	Interval   time.Duration // The duration of time between each log.
-	Resettable bool          // Reset after logging.
+	Mutex     sync.Mutex    // Lock to prevent race-conditions with concurrent processes.
+	Requests  uint64        // Number of HTTP requests since recording started.
+	Since     time.Time     // When the current request counter began recording requests.
+	Interval  time.Duration // The duration of time between each log.
+	Threshold uint64        // When to warn of a high request count during a logging cycle. Setting a
+	// non-zero threshold allows the request counter to reset each interval. Therefore, if left uninitialized,
+	// the request counter will be persistent and never warn or reset.
 }
 
 // increment adds 1 to the request counter, signifying another call to GCR.
@@ -47,13 +49,23 @@ func (rc *RequestCounter) Flush() {
 	rc.Mutex.Lock()
 	defer rc.Mutex.Unlock()
 
-	// Log the number of requests within this measurement window.
-	msg := fmt.Sprintf("From %s to %s [%d min] there have been %d requests to GCR.", rc.Since.Format(TimestampFormat), Clock.Now().Format(TimestampFormat), rc.Interval/time.Minute, rc.Requests)
-	Debug(msg)
+	rc.log()
 
-	if rc.Resettable {
+	// Only allow request counters wi
+	if rc.Threshold > 0 {
 		// Reset the request counter.
 		rc.reset()
+	}
+}
+
+// log the number of HTTP requests found. If the number of requests exceeds the
+// threshold, log an additional warning message.
+func (rc *RequestCounter) log() {
+	msg := fmt.Sprintf("From %s to %s [%d min] there have been %d requests to GCR.", rc.Since.Format(TimestampFormat), Clock.Now().Format(TimestampFormat), rc.Interval/time.Minute, rc.Requests)
+	Debug(msg)
+	if rc.Threshold > 0 && rc.Requests > rc.Threshold {
+		msg = fmt.Sprintf("The threshold of %d requests has been surpassed.", rc.Threshold)
+		Warn(msg)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


legacy/reqcounter/reqcounter.go
```diff
@@ -47,16 +48,25 @@ func (rc *RequestCounter) Flush() {
...
+	rc.log()
 
 	if rc.Resettable {
 		// Reset the request counter.
 		rc.reset()
 	}
 }
 
+// log the number of HTTP requests found. If the number of requests exceeds the
+// threshold, an additional warning message will be logged.
```
listx, line 60: Use the present tense. It's simpler.

`an additional warning message will be logged` -> `log an additional warning message`

legacy/reqcounter/reqcounter_test.go
```diff
@@ -125,6 +125,31 @@ func TestFlush(t *testing.T) {
 	// Ensure the request counter did not reset.
 	require.Equal(t, uint64(66), requestCounter.Requests, "Calling Flush() reset the requests of a non-resettable request counter.")
 	require.True(t, defaultTime.Equal(requestCounter.Since), "Calling Flush() reset the timestamp of a non-resettable request counter.")
+
+	// Create a request counter that exceeded a threshold.
+	requestCounter = NewRequestCounter(600)
+	requestCounter.Threshold = 599
+	// Reset debug counter.
+	debugCalls = 0
+	// Mock logrus.Warn
```
listx, line 134: Missing period.

legacy/reqcounter/reqcounter_test.go
```diff
@@ -125,6 +125,31 @@ func TestFlush(t *testing.T) {
...
+	warnCalls := 0
+	rc.Warn = func(args ...interface{}) {
+		warnCalls++
+	}
+	requestCounter.Flush()
+	// Ensure both logrus.Debug and logrus.Warn was called.
+	require.Equal(t, 1, debugCalls, "Flush() failed to trigger a debug statement after exceeding it's threshold.")
```
listx, line 141: Typo `it's` -> `its`. But also, why not just say `the threshold` instead of `its threshold`?

legacy/reqcounter/reqcounter_test.go
```diff
@@ -125,6 +125,31 @@ func TestFlush(t *testing.T) {
...
+	require.Equal(t, 1, warnCalls, "Flush() failed to trigger a warning statement after exceeding it's threshold.")
+
+	// Create a non-resetting request counter that exceeded a threshold.
+	requestCounter = NewRequestCounter(600)
+	requestCounter.Threshold = 599
+	requestCounter.Resettable = false
+	// Reset warn counter.
+	warnCalls = 0
+	requestCounter.Flush()
+	// Ensure non-resetting counters never warn on exceeding a threshold.
```
listx, line 151: Use the word `resettable` instead of `resetting` to make it easier to grep (use consistent language).

legacy/reqcounter/reqcounter.go
```diff
@@ -32,6 +32,7 @@ type RequestCounter struct {
 	Since      time.Time     // When the current request counter began recording requests.
 	Interval   time.Duration // The duration of time between each log.
 	Resettable bool          // Reset after logging.
+	Threshold  uint64        // When to warn of a high request count (only for resettable request counters).
```
listx, line 35: I think you should mention that this threshold is only for a single Interval period.

But also, there's some logical duplication happening here. If Threshold > 0, then obviously we want to log a warning. However, we also control when to log the warning based on whether it is Resettable.

But will it ever make sense to create a RequestCounter that has a Threshold > 0, but where Resettable = false? I think not.

Allowing datatypes to become inconsistent is not good, and should be prevented by either (1) ensuring that the datatype (RequestCounter here) is only able to be created in a guaranteed-correct way, (2) creating a validation check to examine whether the datatype is a good or bad state before using it, or (3) by making it impossible for the datatype to be in a bad state.

I prefer (3) if possible. Here you could remove `Resettable` and just use Threshold to decide whether to warn or not, which would get us to option (3). That is, if Threshold > 0, warn if we exceed it but also reset the counter if we exceed the threshold interval. If Threshold == 0, then never reset the counter and don't bother dealing with the threshold checking logic because it is set to 0.

legacy/reqcounter/reqcounter.go
tylerferrara: Ack.

legacy/reqcounter/reqcounter_test.go
tylerferrara: Ack.

legacy/reqcounter/reqcounter_test.go
tylerferrara: Ack

legacy/reqcounter/reqcounter.go
tylerferrara: Option 3 is probably the best bet here. I think as long as we document this behavior, it avoids any bad state.

legacy/reqcounter/reqcounter.go
```diff
@@ -27,11 +27,13 @@ import (
...
-	Requests   uint64        // Number of HTTP requests since recording started.
-	Since      time.Time     // When the current request counter began recording requests.
-	Interval   time.Duration // The duration of time between each log.
-	Resettable bool          // Reset after logging.
+	Mutex     sync.Mutex    // Lock to prevent race-conditions with concurrent processes.
+	Requests  uint64        // Number of HTTP requests since recording started.
+	Since     time.Time     // When the current request counter began recording requests.
+	Interval  time.Duration // The duration of time between each log.
+	Threshold uint64        // When to warn of a high request count during a logging cycle. Setting a
+	// non-zero threshold allows the request counter to reset each interval. Therefore, if left uninitialized,
```
listx, line 35: Nit: Useless use of `Therefore`

listx (commented):
Thanks!

/lgtm
/approve

legacy/reqcounter/reqcounter.go
tylerferrara: Sorry I didn't see this earlier. Here's that quick fix: #386

PR #388, opened 2021-08-04
Create script to run auditor locally

tylerferrara:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
This change adds a script to run the auditor locally for testing/debugging purposes. The current parameters invoke the auditor to watch the `gcr.io/k8s-artifacts-prod` top-level registry, just like the deployment. Running this script requires a service-account for creating the errorreporting client. This service-account needs no special permissions to the k8s-artifacts-prod project, so any service-account should suffice.

#### Which issue(s) this PR fixes:
Pertains to #353 

#### Special notes for your reviewer:
Running this script is purely a development tool for ensuring valid behavior of the auditor. It merely mimics the deployment version of the auditor which runs inside Cloud Run.

#### Does this PR introduce a user-facing change?
NONE

```release-note
Created script to run auditor locally for testing/debugging.
```
cc: @listx @amwat @justaugustus @kubernetes-sigs/release-engineering

PR #389, opened 2021-08-04
Watch OS signals when providing --verbose flag

tylerferrara:
#### What type of PR is this?
/kind feature

When providing the `--verbose` flag to the auditor, all OS signals are logged when encountered. The two exceptions are SIGKILL and SIGSTOP as the `os/signal` package [cannot catch these signals](https://pkg.go.dev/os/signal#hdr-Types_of_signals).

Capturing all OS signals may gain insight into how and why the auditor keeps restarting #353 

Part of #353 

When observing OS signals, we are expected to\ handle them. This is why we must check for all "termination" signals and force the program to exit. If we did not do this, our program would only exit when given a SIGKILL. Realistically, when deployed in Cloud Run, the container will be stopped after sending a SIGINT, so this would not effect deployment situations. However, when running locally #388, the program would not terminate when given hitting (ctrl-c).

YES
When providing the auditor with the `--verbose`, debug messages will be logged like so:
```
DEBU Watching for OS Signals...                    file="signals/signals.go:50"
INFO Defaulting to port 8080                       file="audit/auditor.go:94"
INFO Listening on port 8080                        file="audit/auditor.go:97"
DEBU Encoutered signal: bus error                  file="signals/signals.go:52"
DEBU Encoutered signal: floating point exception   file="signals/signals.go:52"
DEBU Encoutered signal: urgent I/O condition       file="signals/signals.go:52"
DEBU Encoutered signal: signal 54                  file="signals/signals.go:52"
DEBU Encoutered signal: urgent I/O condition       file="signals/signals.go:52"
DEBU Encoutered signal: child exited               file="signals/signals.go:52"
DEBU Encoutered signal: trace/breakpoint trap      file="signals/signals.go:52"
DEBU Encoutered signal: window changed             file="signals/signals.go:52"
DEBU Encoutered signal: urgent I/O condition       file="signals/signals.go:52"
DEBU Encoutered signal: urgent I/O condition       file="signals/signals.go:52"
```

```release-note
Observe all OS signals directed to the auditor when passing the --verbose flag.
```

### DIFF
--- a/legacy/signals/signals.go
+++ b/legacy/signals/signals.go
@@ -1,0 +1,54 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package signals
+
+import (
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/sirupsen/logrus"
+)
+
+// Watch concurrenlty logs debug statements when encountering interrupt signals from the OS.
+// This function relies on the os/signal package which may not capture SIGKILL and SIGSTOP.
+func Watch() {
+	c := make(chan os.Signal, 1)
+	// Observe all signals, excluding SIGKILL and SIGSTOP.
+	signal.Notify(c)
+	// Continuously watch for signals.
+	go func() {
+		logrus.Debug("Watching for OS Signals...")
+		for sig := range c {
+			logrus.Debug("Encoutered signal: ", sig.String())
+			// If the observed signal is a termination signal, we are
+			// expected to handle this by exiting the program. If we didn't
+			// exit, the program would only stop upon receiving a SIGKILL.
+			if sig == syscall.SIGHUP ||
+				sig == syscall.SIGINT ||
+				sig == syscall.SIGABRT ||
+				sig == syscall.SIGILL ||
+				sig == syscall.SIGQUIT ||
+				sig == syscall.SIGTERM ||
+				sig == syscall.SIGSEGV ||
+				sig == syscall.SIGTSTP {
+				logrus.Debug("Exiting from signal: ", sig.String())
+				os.Exit(0)
+			}
+		}
+	}()
+}
--- a/legacy/cli/audit.go
+++ b/legacy/cli/audit.go
@@ -25,6 +25,7 @@ import (
 
 	"sigs.k8s.io/k8s-container-image-promoter/legacy/audit"
 	"sigs.k8s.io/k8s-container-image-promoter/legacy/reqcounter"
+	"sigs.k8s.io/k8s-container-image-promoter/legacy/signals"
 )
 
 type AuditOptions struct {
@@ -59,6 +60,8 @@ func RunAuditCmd(opts *AuditOptions) error {
 		logrus.SetLevel(logrus.DebugLevel)
 		// Initialize global counter to track the number of HTTP requests made to GCR.
 		reqcounter.Init()
+		// Watch for OS signals.
+		signals.Watch()
 	}
 
 	auditorContext.RunAuditor()


tylerferrara:
We have two options when observing OS signals.

1. Observe all signals
Pro: Capture all behavior directed to the auditor (hopefully determine why the auditor is restarting in cloud run)
Con: We must manually define which signals to terminate on. Otherwise, we ignore the signal.
2. Observe a few specific signals
Pro: Gain insight on the specific signals we care about (termination signals)
Con: We may miss out on signals that show underlying errors causing the auditor to fail.

legacy/signals/signals.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/sirupsen/logrus"
+)
+
+// sigMap holds all signals the Watch() function should terminate
+// when encountering. The key holds the string representation
```
listx, line 28: This first sentence sounds odd but I don't know the best way to reword it. Can you try to reword it?

listx: Looking at the rest of this code, it looks like all of these signals are the ones that should kill (terminate) the auditor (main process). In that case a better name might be `exitSignals`.

legacy/signals/signals.go
```diff
@@ -0,0 +1,61 @@
...
+	signal.Notify(c)
+	// Continuously watch for signals.
+	go func() {
+		logrus.Debug("Watching for OS Signals...")
+		for sig := range c {
+			logrus.Debug("Encoutered signal: ", sig.String())
+			// Ensure the current program stops when encountering
+			// an interrupt. Without this check, these signals would
+			// not cause the program to terminate.
+			if s, exists := sigMap[sig.String()]; exists {
```
listx, line 56: Instead of using `sig.String()` as the key, why not `sig.Signal()` ?

legacy/signals/signals.go
```diff
@@ -0,0 +1,61 @@
...
+				os.Exit(int(s))
```
listx, line 57: It's a bit odd to exit with the integer value of the signal. Better to log it (just include it in the Debug line above at line 52). And then here you can additionally log `"Exiting because of signal %s (%s)", sig.String(), sig.Signal()` or something similar.

legacy/signals/signals.go
tylerferrara: Ack, re-wording

legacy/signals/signals.go
tylerferrara: [sig.Signal() ](https://cs.opensource.google/go/go/+/refs/tags/go1.16.6:src/os/exec.go;l=69-72;bpv=1;bpt=1)does not return anything. For this reason, I use sig.String().

legacy/signals/signals.go
tylerferrara: Concurrently capturing signals requires me to explicitly exit. The above comment explains this behavior. If I didn't do this, and a `SIGINT` was given, the program would not terminate. But I can add an extra logging statement to show the program is about to exit.

However, a [signal](https://pkg.go.dev/syscall#Signal) is of type int, so this casting is actually quite trivial.

listx: The `os.Exit()` is fine, but exiting with the `int(s)` as argument is what is awkward. Better to log it as `int(s)` or some such in the Debug line above it.

legacy/signals/signals.go
```diff
@@ -0,0 +1,61 @@
...
+	"syscall"
+
+	"github.com/sirupsen/logrus"
+)
+
+// exitSignals holds a mapping of os.Signal.String() to syscall for all terminating signals.
+// When the Watch() function encounters a signal within this map, it explicitly forces the
+// program to exit with the corrosponding syscall exit code.
+var exitSignals = map[string]syscall.Signal{
+	"hangup":              syscall.SIGHUP,
```
listx, line 31: Can we have this reversed (use syscall.SIGHUP as the key and "hangup" as the value), or even better, just use an empty interface as the value? It's something of an antipattern to use the String representation of the thing to compare when we have the actual thing itself. It's like comparing 2 integers by their string representation instead of using the int type natively.

legacy/signals/signals.go
listx: Then I guess we can just use `int(s)` instead?

tylerferrara: sig is of type os.Signal which is an interface. I must use [syscall.Signal](https://pkg.go.dev/syscall#Signal)

legacy/signals/signals.go
tylerferrara: I think we are confusing [os.Signal](https://cs.opensource.google/go/go/+/refs/tags/go1.16.6:src/os/exec.go;l=69-72) with [syscall.Signal](https://pkg.go.dev/syscall#Signal). When we observe an incoming OS signal, it's of type: os.Signal (an interface). The way i can identify it is by the unique sig.String() it produces. I map these sig.String() values to syscall.Signals. This way I can gain access to the int exit code.

legacy/signals/signals.go
```diff
@@ -0,0 +1,54 @@
...
+	signal.Notify(c)
+	// Continuously watch for signals.
+	go func() {
+		logrus.Debug("Watching for OS Signals...")
+		for sig := range c {
+			logrus.Debug("Encoutered signal: ", sig.String())
+			// If the observed signal is a termination signal, we are
+			// expected to handle this by exiting the program. If we didn't
+			// exit, the program would only stop upon receiving a SIGKILL.
+			if sig == syscall.SIGHUP ||
```
listx, line 41: Nit: I would have just converted the map you used to have to use these signals as keys directly, and used the "does the key exist in the map" check  instead.

listx:
Thanks!

/lgtm
/approve

PR #391, opened 2021-08-05
Add unit tests to signals package

tylerferrara:
#### What type of PR is this?
/kind cleanup

This PR introduces unit tests to the `signals` package. It also re-factors much of `signals.go` to allow for easier testing.

None

Since the Auditor is designed to run indefinitely, like a server, it does not support graceful termination. Therefore, to stop it, we must send a signal like `SIGINT` (ctrl-c) or `SIGQUIT`. These signals are captured by the `signals` package, but simply call `os.Exit(0)`. In the future, this package may be able to enable graceful termination for the auditor. This would probably mean defining a shutdown function that is called upon exit.

However, let it be known that we cannot capture `SIGKILL` and `SIGSTOP` signals. Meaning, if the Auditor is are given either of these signals, it will be unable to observe them within `signals` and not trigger a graceful termination. As of now, there is no getting around this, since the [os/signal](https://pkg.go.dev/os/signal#hdr-Types_of_signals) package doesn't support this. 

NONE

```release-note
Add unit tests to signals package.
```

### DIFF
--- a/legacy/signals/signals.go
+++ b/legacy/signals/signals.go
@@ -24,31 +24,68 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+var (
+	// ExitSignals are used to determine if an incoming os.Signal should cause termination.
+	ExitSignals map[os.Signal]bool = map[os.Signal]bool{
+		syscall.SIGHUP:  true,
+		syscall.SIGINT:  true,
+		syscall.SIGABRT: true,
+		syscall.SIGILL:  true,
+		syscall.SIGQUIT: true,
+		syscall.SIGTERM: true,
+		syscall.SIGSEGV: true,
+		syscall.SIGTSTP: true,
+	}
+	// ExitChannel is for gracefully terminating the LogSignals() function.
+	ExitChannel chan bool = make(chan bool, 1)
+	// SignalChannel is for listening to OS signals.
+	SignalChannel chan os.Signal = make(chan os.Signal, 1)
+	// Debug is defined globally for mocking logrus.Debug statements.
+	Debug func(args ...interface{}) = logrus.Debug
+)
+
 // Watch concurrenlty logs debug statements when encountering interrupt signals from the OS.
 // This function relies on the os/signal package which may not capture SIGKILL and SIGSTOP.
 func Watch() {
-	c := make(chan os.Signal, 1)
+	Debug("Watching for OS Signals...")
 	// Observe all signals, excluding SIGKILL and SIGSTOP.
-	signal.Notify(c)
-	// Continuously watch for signals.
-	go func() {
-		logrus.Debug("Watching for OS Signals...")
-		for sig := range c {
-			logrus.Debug("Encoutered signal: ", sig.String())
-			// If the observed signal is a termination signal, we are
-			// expected to handle this by exiting the program. If we didn't
-			// exit, the program would only stop upon receiving a SIGKILL.
-			if sig == syscall.SIGHUP ||
-				sig == syscall.SIGINT ||
-				sig == syscall.SIGABRT ||
-				sig == syscall.SIGILL ||
-				sig == syscall.SIGQUIT ||
-				sig == syscall.SIGTERM ||
-				sig == syscall.SIGSEGV ||
-				sig == syscall.SIGTSTP {
-				logrus.Debug("Exiting from signal: ", sig.String())
-				os.Exit(0)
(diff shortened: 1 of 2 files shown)


legacy/signals/signals.go
```diff
@@ -24,31 +24,68 @@ import (
...
+		syscall.SIGABRT: true,
+		syscall.SIGILL:  true,
+		syscall.SIGQUIT: true,
+		syscall.SIGTERM: true,
+		syscall.SIGSEGV: true,
+		syscall.SIGTSTP: true,
+	}
+	// ExitChannel is for gracefully terminating the LogSignals() function.
+	ExitChannel chan bool = make(chan bool, 1)
+	// SignalChannel is for passing observed OS signals.
```
listx, line 41: Nit: I think SignalChannel is for just listening to signals from the OS. "passing" sounds like we (auditor) is doing the passing, which is not true (we always only receive from the end of this channel; we never put (pass) things into it).

legacy/signals/signals.go
```diff
@@ -24,31 +24,68 @@ import (
...
-				sig == syscall.SIGABRT ||
-				sig == syscall.SIGILL ||
-				sig == syscall.SIGQUIT ||
-				sig == syscall.SIGTERM ||
-				sig == syscall.SIGSEGV ||
-				sig == syscall.SIGTSTP {
-				logrus.Debug("Exiting from signal: ", sig.String())
-				os.Exit(0)
-			}
+	signal.Notify(SignalChannel)
```
listx, line 52: Nit: Looking at https://gobyexample.com/signals, apparently you can specify here which signals to listen for. So we don't even need ExitSignals.

Since you've got tests working and stuff, I'll chalk up the above to future refactoring work, but I wanted to capture the thought here.

legacy/signals/signals_test.go
```diff
@@ -0,0 +1,152 @@
...
+		// Ensure the logs are correct.
+		require.EqualValues(t, st.expected, logs, "Unexpected signal observation logs.")
+		if st.terminate {
+			// Ensure Stop() was invoked if the test specifies.
+			require.True(t, terminated(), "LogSignal did not terminate on exit signal.")
+		}
+	}
+}
+
+func TestLogSignals(t *testing.T) {
```
listx, line 114: Nit: Add a comment somewhere in here describing what this test does. I think you're testing whether termination happens in situations where (1) we explicitly call signals.Stop(), or (2) where we pass in a signal that is recognized as one that should exit (line 147). Is this correct?

legacy/signals/signals_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+	// Reset termination bool for new test.
+	terminated = false
+	// Start logging.
+	wg.Add(1)
+	go logSignals()
+	// Pass multiple non-exit signals.
+	signals.SignalChannel <- syscall.SIGTTOU
+	signals.SignalChannel <- syscall.SIGCHLD
+	signals.SignalChannel <- syscall.SIGPIPE
```
listx, line 146: Nit: Unless you're performing some sort of checking of state in between these non-exit signals, these insertions into the channel are questionable. "If you cannot measure it, it is meaningless."

listx (commented):
All nits. Either fix or remove hold to merge. Thanks!

/lgtm
/approve
/hold

legacy/signals/signals.go
tylerferrara: Ack.

legacy/signals/signals.go
tylerferrara: I talked about the trade offs between observing all signals and only some signals in this comment: https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/389#issuecomment-892299750
It seems that we may benifit from observing all signals, in an effort to improve the Auditor's visibility #392

legacy/signals/signals_test.go
tylerferrara: SignalChannel is buffered for 1 signal at a time. Therefore, if the consumer (LogSignals) does not consume them, we will block and the test will fail. This is why I send multiple signals, ensuring we are actually handling them. I'll mention this in a comment.

legacy/signals/signals_test.go
tylerferrara: Added comment.

listx:
Thanks!

/lgtm
/approve

PR #396, opened 2021-08-09
go_with_version.sh: Shellcheck and drop `return` statement

justaugustus:
#### What type of PR is this?

/kind cleanup

(Peeled off of https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/285.)

go_with_version.sh: Shellcheck and drop `return` statement

`return` only works within the context of a function. 
Given we `set -o errexit`, a command failure will cause the script to fail anyway, so this line is extraneous.

/assign @saschagrunert @puerco @cpanato 

```release-note
NONE
```

### DIFF
--- a/go_with_version.sh
+++ b/go_with_version.sh
@@ -28,7 +28,7 @@ printUsage() {
     >&2 echo "Usage: $0 (build | run) path/to/code.go [args...]"
 }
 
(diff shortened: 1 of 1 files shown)


cpanato (approved):
/lgtm

PR #397, opened 2021-08-09
Dockerfile: Use an explicit go version (go1.16.7) for image building

justaugustus:
#### What type of PR is this?

/kind cleanup
/area dependency

- Dockerfile: Use an explicit go version (go1.16.7) for image building
  
  Using a `latest` tag in general, but specifically for Golang-based
  images will eventually lead to unexpected variance in our image builds.
  
  Here we pull an explicit go version and add an entry for it in
  `dependencies.yaml`.
  
  This commit also adds a dependency entry for the go major version in
  `go.mod`, to ensure it stays in sync with image builds.

```release-note
- Dockerfile: Use an explicit go version (go1.16.7) for image building
  
  Using a `latest` tag in general, but specifically for Golang-based
  images will eventually lead to unexpected variance in our image builds.
  
  Here we pull an explicit go version and add an entry for it in
  `dependencies.yaml`.
  
  This commit also adds a dependency entry for the go major version in
  `go.mod`, to ensure it stays in sync with image builds.
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@
 # Determine final build variant [prod | test].
 ARG variant
 
-FROM golang:latest AS base
+FROM golang:1.16.7 AS base
 # Transfer all project files to container.
 WORKDIR /go/src/app
 COPY . .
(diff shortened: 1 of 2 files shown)


justaugustus:

cpanato (approved):
/lgtm

PR #261, opened 2020-10-10
go: Init submodules for lib/ and pkg/

justaugustus:
Go submodule approach to solve https://github.com/kubernetes/release/pull/1619#discussion_r502631879.
Supersedes https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/260.

Now that we're attempting to integrate the manifest merging capabilities into k/release (https://github.com/kubernetes/release/pull/1619), we're running into the dreaded v2 semantic import versioning problem:

```console
require sigs.k8s.io/k8s-container-image-promoter: version "v2.4.0" invalid: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2
```

ref: https://blog.golang.org/v2-go-modules, https://donatstudios.com/Go-v2-Modules, https://peter.bourgon.org/blog/2020/09/14/siv-is-unsound.html

This PR establishes go submodules for `lib/` and `pkg/` so that consumers can import libraries in this repo.

Following this merging, we'd tag minor versions (v0.1.0) of the new submodules:
- tag: `lib/v0.1.0`
- tag: `pkg/v0.1.0`

Signed-off-by: Stephen Augustus <saugustus@vmware.com>

PR #260, opened 2020-10-10
Init v2 module

justaugustus:
Now that we're attempting to integrate the manifest merging capabilities into k/release (https://github.com/kubernetes/release/pull/1619), we're running into the dreaded v2 semantic import versioning problem:

```console
require sigs.k8s.io/k8s-container-image-promoter: version "v2.4.0" invalid: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2
```

ref: https://github.com/kubernetes/release/pull/1619#discussion_r502631879, https://blog.golang.org/v2-go-modules, https://donatstudios.com/Go-v2-Modules, https://peter.bourgon.org/blog/2020/09/14/siv-is-unsound.html

This PR establishes a v2 module for CIP, so that consumers can use the libraries in this repo.

/kind feature
/assign @listx @dims 
cc: @kubernetes-sigs/release-engineering @bytecook

PR #425, opened 2021-09-14
go.mod: Init v3 module

justaugustus:
#### What type of PR is this?

/kind feature api-change

- go.mod: Init v3 module
- v3: Update import paths

---

This is a follow-up attempt to bring the repo into compliance with Semantic Import Versioning for go v2+ modules.

The previous attempts were... less successful, due to bazel:
- https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/260
- https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/261

More context from https://github.com/kubernetes-sigs/k8s-container-image-promoter/commit/657fd4e94ccab3129063c3dc3610797a45e2df10#comments (and https://kubernetes.slack.com/archives/CJH2GBF7Y/p1622457503212900):

> tl;dr-ish:
> by moving the packages back into the CIP repo, it now has content that needs to be imported into other repos (cip --> k/release).
>
> When a v2+ tag is [cut](https://github.com/kubernetes-sigs/k8s-container-image-promoter/releases/tag/v2.0.0), the modules need to be support it, either by:
> - adding a v2 subfolder and moving the packages
> - marking the module as v2 in go.mod
> - release branches with one of the aforementioned options
> 
> ref: https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher
>
> I also experimented with submodules, which went poorly: https://github.com/kubernetes-sigs/k8s-container-image-promoter/pull/261
>
> Now for reasons unclear from the error messages, gazelle was not respecting the declaration of a v3 module and stripping the local imports from the BUILD.bazel files, which causes all the bazel-based tests to fail.
>
> This is likely fixed by newer versions of bazel+gazelle+rules_go, but given we’re trying to remove bazel, it didn’t make sense to explore that again.
>
> So the actual tl;dr — cutting a v1 tag was the least-involved solution.
>
> We can re-explore a v3 module once bazel gets removed.

I used https://github.com/marwan-at-work/mod for this:

```console
GO111MODULE=on go get github.com/marwan-at-work/mod/cmd/mod
~/go/bin/mod upgrade -t 3
```

```release-note
go.mod: Init v3 module
```

### DIFF
--- a/api/files/manifest_test.go
+++ b/api/files/manifest_test.go
@@ -21,7 +21,7 @@ import (
 	"strings"
 	"testing"
 
-	"sigs.k8s.io/k8s-container-image-promoter/api/files"
+	"sigs.k8s.io/k8s-container-image-promoter/v3/api/files"
 )
 
 func TestValidateFilestores(t *testing.T) {
--- a/cmd/cip-mm/main.go
+++ b/cmd/cip-mm/main.go
@@ -23,7 +23,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
-	reg "sigs.k8s.io/k8s-container-image-promoter/legacy/dockerregistry"
+	reg "sigs.k8s.io/k8s-container-image-promoter/v3/legacy/dockerregistry"
 	"sigs.k8s.io/release-utils/log"
 )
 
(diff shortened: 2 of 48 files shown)


cpanato (approved):
/lgtm

/hold for more review or if all good can remove

PR #482, opened 2022-01-14
Add image flag to kpromo pr to allow filtering by images

### LINKED ISSUE #479: Allow explicitly filtering images by name via `kpromo pr --image`

A way to specify one or more images that should be skipped/excluded when running `kpromo pr`. 

For example:

```
kpromo pr --skip-images capd-manager --fork fabriziopandini --project cluster-api --reviewers "@CecileRobertMichon @enxebre @fabriziopandini @vincepri" --tag v1.1.0-beta.1
```

#### Why is this needed:

Some projects such as Cluster API may have tagged images in staging that are only meant for dev/testing purposes and should not be promoted to the prod repo. Having a way to exclude those images would provide an optional way to skip certain images if found, while preserving existing behavior if the flag is not specified.

CecileRobertMichon:
#### What type of PR is this?
/kind feature

Allows specifying one or more images that should be promoted in kpromo pr to allow filtering by image. This relies on the pre-existing functionality of `GrowManifests` and simply exposes the option. It also makes `kpromo pr` closer to feature parity with `cip-mm` which will help deprecate `cip-mm`.

If no image is specified, the default behavior stays the same: all images matching the tag in the project are promoted.

Example PRs opened with this branch:

no image specified: https://github.com/kubernetes/k8s.io/pull/3259
specific images: https://github.com/kubernetes/k8s.io/pull/3262

Fixes #479 

```release-note
Add image flag to kpromo pr to allow filtering by images
```

### DIFF
--- a/cmd/kpromo/cmd/pr/pr.go
+++ b/cmd/kpromo/cmd/pr/pr.go
@@ -65,6 +65,7 @@ type promoteOptions struct {
 	tags            []string
 	reviewers       string
 	interactiveMode bool
+	images          []string
 }
 
 func (o *promoteOptions) Validate() error {
@@ -133,6 +134,14 @@ func init() {
 		"i",
 		false,
 		"interactive mode, asks before every step",
+	)
+
+	PRCmd.PersistentFlags().StringSliceVarP(
+		&promoteOpts.images,
+		"image",
+		"",
+		[]string{""}, // default to a single empty string image filter which will promote all images matching the tag.
+		"image to promote. If not specified, all images matching the tag will be promoted",
 	)
 
 	for _, flagName := range []string{"tag", "fork"} {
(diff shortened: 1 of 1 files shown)


CecileRobertMichon:
> ERROR: gcloud crashed (V2DiagnosticException)

hmm that doesn't seem related to my changes 🤔 

/retest

cmd/kpromo/cmd/pr/pr.go
```diff
@@ -135,6 +136,14 @@ func init() {
 		"interactive mode, asks before every step",
 	)
 
+	PRCmd.PersistentFlags().StringSliceVarP(
+		&promoteOpts.images,
+		"image",
```
fabriziopandini, line 141: I don't have data points about the most frequent behavior is, but I'm wondering if we should have skip-images instead (default all, skip what you don't want to promote vs current implementation default all, explicitly list all you 
want to promote)

CecileRobertMichon: I switched to explicitly listing images to include based on @justaugustus's comment [here](https://github.com/kubernetes-sigs/promo-tools/issues/479#issuecomment-1006216736)

fabriziopandini: thanks for the context! then +1 from me

justaugustus: ref: https://github.com/kubernetes-sigs/promo-tools/issues/479#issuecomment-1006200436

@CecileRobertMichon @fabriziopandini -- I expanded a little on this (inclusion vs exclusion) in the [2022-01-18 RelEng meeting](https://docs.google.com/document/d/16GqCjnEh86w8yADcrUylNoE1y1sqjIMYNC_gdk5WPSQ/edit#bookmark=id.lzc56fdhbbvl)...

Inclusion: It's very likely that you will know exactly what images you're interested in promoting.
Exclusion: You will need to discover which images you need to exclude. While you may know some of them, at any point new images could be created within the staging repository (maybe by error or by another maintainer who stages their images in the repo) which match your exclusion tag.

justaugustus (approved):
Thanks @CecileRobertMichon!

fabriziopandini:
Thanks @CecileRobertMichon! looking forward to using to tool in the next CAPI release!

PR #493, opened 2022-02-04
kpromo pr: Allow sorting output manifest by tags instead of digest

### LINKED ISSUE #473: cip-mm sort by tag instead of checksum
Currently, if you use cip-mm to promote images, it sorts the generated file by the checksum of the promoted container images. Ideally it would allow sorting the generated file by tag instead or have an option to provide that as a argument.

#### Why is this needed:
If you promote images for multiple architectures it makes it very difficult to spot if all of them are there when looking at the file.
See for example: https://github.com/kubernetes/k8s.io/pull/2595/files

CecileRobertMichon:
#### What type of PR is this?
/kind feature

This PRs allows optionally sorting manifests by tag so they are easier to review by human reviewers. The default is the current behavior of sorting by image digest.

Fixes #473 

```release-note
kpromo pr: Allow sorting output manifest by tags instead of digest
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	cloud.google.com/go/grafeas v0.1.0 // indirect
 	cloud.google.com/go/logging v1.4.2
 	cloud.google.com/go/storage v1.21.0
+	github.com/blang/semver v3.5.1+incompatible // indirect
 	github.com/cenkalti/backoff/v4 v4.1.2
 	github.com/google/go-cmp v0.5.7
 	github.com/google/go-containerregistry v0.8.1-0.20220216220642-00c59d91847c
@@ -84,7 +85,6 @@ require (
 	github.com/benbjohnson/clock v1.1.0 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/bgentry/speakeasy v0.1.0 // indirect
-	github.com/blang/semver v3.5.1+incompatible // indirect
 	github.com/cenkalti/backoff/v3 v3.2.2 // indirect
 	github.com/census-instrumentation/opencensus-proto v0.3.0 // indirect
 	github.com/cespare/xxhash/v2 v2.1.2 // indirect
--- a/image/image.go
+++ b/image/image.go
@@ -92,8 +92,10 @@ func (imagesList *ManifestList) Write(filePath string) error {
 // ToYAML serializes an image list into an YAML file.
 // We serialize the data by hand to emulate the way it's done by the image promoter
 func (imagesList *ManifestList) ToYAML() ([]byte, error) {
-	// The image promoter code sorts images by:
-	//	  1. Name 2. Digest SHA (asc)  3. Tag
+	// Images are sorted by:
+	//	  1. Name 2. Tag
+	// If there are multiple tags for a single image digest, the smallest tag is used in the sort.''
+	// Image tags are sorted lexicographically as they are not guaranteed to be Semver compliant. See https://github.com/opencontainers/distribution-spec/issues/154 fpr more details.
 
 	// First, sort by name (sort #1)
 	sort.Slice(*imagesList, func(i, j int) bool {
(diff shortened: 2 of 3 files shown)


fabriziopandini:
Thanks, @CecileRobertMichon, this looks great!

cpanato (approved):
it looks great!
/lgtm
/hold

CecileRobertMichon:
@puerco do I need to change anything in this PR after #494 merged?

CecileRobertMichon:
Used this branch to promote CAPI v1.1.2 https://github.com/kubernetes/k8s.io/pull/3422

CecileRobertMichon:
rebased

@cpanato are we waiting for anyone specific to comment on this? Would love to get this merged soon so we can start using it

cpanato:
> rebased
> 
> @cpanato are we waiting for anyone specific to comment on this? Would love to get this merged soon so we can start using it

was waiting for others to comment, but it is been a long time, let's get this in!
/unhold
/lgtm

CecileRobertMichon:
@justaugustus did you re-add the hold intentionally?

justaugustus:
> @justaugustus did you re-add the hold intentionally?

Yep! A few comments here: https://github.com/kubernetes-sigs/promo-tools/pull/493#pullrequestreview-891530774

CecileRobertMichon:
> Let's just make this the only behavior.

My concern here is that everyone using the tool will see a huge diff the first time then run the new version since all the previous images will get reordered. I was also under the impression from the discussion in the issue (#473) that some projects prefer sorting by image digest since that's more "machine-readable". I think it makes more sense to start with an optional opt in sort so projects can keep using the tool without being affected and then if the demand is there move this to default later on and deprecate the sorted by digest functionality (or not).

> We need a lexicographical sort, as opposed to strictly SemVer.

Thanks for catching that, I assumed all tags would be semver. Will apply these changes when I get a chance.

justaugustus:
> My concern here is that everyone using the tool will see a huge diff the first time then run the new version since all the previous images will get reordered.

That's fine; we can announce the changes.

> I was also under the impression from the discussion in the issue (#473) that some projects prefer sorting by image digest since that's more "machine-readable".

Initially, I thought you wanted to introduce a new structure altogether.
The concern was that co-mingling the old and new structures would cause unexpected unmarshalling errors.

Given it's the same data structure, the promoter isn't going to need to care about what order the content is in.

> I think it makes more sense to start with an optional opt in sort so projects can keep using the tool without being affected and then if the demand is there move this to default later on and deprecate the sorted by digest functionality (or not).

Variance introduces risk.
I'd prefer a single solution/codepath to debug than multiple.
(Part of why I suggested deprecating `cip-mm` in https://github.com/kubernetes-sigs/promo-tools/issues/473#issuecomment-991167871, since `cip-mm` and `kpromo pr`'s behavior is already divergent.)

> > We need a lexicographical sort, as opposed to strictly SemVer.
> 
> Thanks for catching that, I assumed all tags would be semver. Will apply these changes when I get a chance.

SGTM!

CecileRobertMichon:
@justaugustus added a commit to always sort by tags and another to sort lexicographically. PTAL when you get a chance.

@fabriziopandini FYI the tags won't be sorted by semver version unfortunately as @justaugustus pointed out that image tags are not always semver compliant so we need to sort lexicographically. Still an improvement though.

justaugustus (approved):
Awesome! Thanks @CecileRobertMichon!! 🎉 
/hold cancel

CecileRobertMichon:
/hold 

should I squash?

CecileRobertMichon:
/hold cancel 

@justaugustus I squashed

How do we announce this so it doesn't surprise folks?

justaugustus:
> should I squash?

Let's let the bots do the work :)
/label tide/merge-method-squash
/hold cancel

justaugustus:
> How do we announce this so it doesn't surprise folks?

We'll announce for the next release.

PR #494, opened 2022-02-08
The Big Image Promoter Refactor™

puerco:
#### What type of PR is this?
/kind cleanup
/kind design

This PR is a huge refactor of the main image promoter code. The refactor consists of two main parts:

1. All the promotion logic has been moved from the old monolithic `cil.RunPromoteCmd()` function to a new package that captures all the promoter modes of operation.
2. The new `promoter` package itself. This new package exposes 4 main functions that run all the steps to complete each of the promoter's modes of operation. The names are self explanatory, here is the TLDR:
  - 1.  `promoter.CheckManifestLists()`: This function implements a mode that only checks manifest lists and exits.
  - 2. `promoter.Snapshot()` runs the snapshotting feature of the promoter
  - 3. `promoter.SecurityScan()` scans the manifests but instead of copying, performs a vulnerability scan in the new images
  - 4. `promoter.PromoteImages()` performs the image promotions based on the options passed

All the code previously intertwined in the `cli.RunPromoteCmd()` function is now split into smaller. more focused functions.

The external interfaces and all functionality should be exactly the same. The idea of the refactor is to break the code into more manageable pieces that we can iterate in a safer and faster fashion.

After we manage this initial refactor, it will be a lot easier to add image verification/signing and the replication function for the k8s infra replication effort.

This is part of the push to get the container images signed (https://github.com/kubernetes/release/issues/2383) and replicated  (https://github.com/kubernetes/k8s.io/issues/1834).

All changes are split into sequential commits that try to develop the refactor one piece at a time, I tried to comment each with at least a little note.

I've added comprehensive notes to the old `RunPromoteCmd()` function in `legacy/cli/run.go`. These can be found in commits ef8145a8478dd5ddb48b62612d3e9c557cef80fb and ef8145a8478dd5ddb48b62612d3e9c557cef80fb . Those notes detail how the function was to be split (the comments are gone in the final version of the file).

The implementation functions are split into files concerning the mode of operation they apply to:

| File | Contents |
| --- | --- |
| common.go | Functions common to all modes |
| options.go | The promoter options struct and methods |
| promoter.go | The main package of the promoter |
| promotion.go | Methods of the implementation related to image promotion |
| scanner.go | Steps in the implementation to perform security scans |
| snapshot.go | Implementation methods for snapshots |
| util.go | Minor general utility functions and constants |

counterfeiter generation has been enabled and fakes are included in the PR. No integration tests are written yet but are coming soon.

I'm putting a hold on this to try to run some tests locally but I wanted to PR the changes before to get some eyes on it as it is a really large pull request.

/assign @justaugustus @saschagrunert @cpanato 

```release-note
The Big Image Promoter Refactor: The image promoter code (`kpromo cip`) code has been completely refactored to make it cleaner and to to get it ready for image signing and image mirroring to other projects
```


puerco:
/hold 
For manual testing and checks

pkg/promoter/options.go
```diff
@@ -0,0 +1,116 @@
...
+package promoter
+
+import (
+	"errors"
+)
+
+// Options capture the switches available to run the image promoter
+type Options struct {
+	// Threads determines how many promotion threads will run
+	Threads int
```
saschagrunert, line 26: How about making this an unsigned integer?

saschagrunert (commented):
Just a few nits, otherwise LGTM :)

pkg/promoter/options.go
puerco: For now, I want to keep the interactions with the internal code just as it is in case the refactor introduces any bugs.

I've noticed several bugs while testing but I don't want to fix them either just jet as I want this to run exactly as it did before moving things around. WDYT?

pkg/promoter/options.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// Confirm
+	Confirm bool
```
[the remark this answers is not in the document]
puerco: Done I've added a more thorough comment. I think some flags should be reworked, but for now, let's overdocument them all as they are sometimes difficult to understand.

puerco:
Thanks @saschagrunert , I've addressed the nits, ptal!

saschagrunert (approved):
/lgtm

saschagrunert:
> I'm putting a hold on this to try to run some tests locally but I wanted to PR the changes before to get some eyes on it as it is a really large pull request.

Feel free to lift when you're ready

justaugustus (changes requested):
@puerco -- Very exciting to see this work!!

I want to hit the `/hold cancel` ASAP, but I'll be pedantic and make a few nits...

### File names

- common.go, util.go -- is it possible to be more descriptive about these/move them to more appropriate places? It's usually preferred to not have file/package names like these. Are they externally useful or maybe more suitable as `internal/` packages?
- scanner.go -- Is this a generic scanner or specifically security-focused? Follow-up: I don't believe we actually use the security scanning in practice, and I wonder if this is better-suited being offloaded to existing tooling in the ecosystem e.g., https://www.whitesourcesoftware.com/resources/blog/docker-image-security-scanning/
- promoter.go vs promotion.go -- a little confusing if you're not aware of the rest of the system

### Imports

Group imports by:
- stdlib, then:
- non-`*.k8s.io`, then:
- `*.k8s.io` imports

### Top-level packages

For newer packages, I've been preferring keeping them in the top-level of the repo for nicer flow in imports.

Consider:

`sigs.k8s.io/promo-tools/api/files`

VS

`sigs.k8s.io/promo-tools/pkg/promoter/foo`

The former is a little shorter and stutters less.

### Parity w/ file promotion package structure

Let's also consider the file promotion packages, as the image promotion ones should ideally keep a similar structure:

- api/files
- filepromoter
- promobot

They started off as disparate tools, but it would be nice to see an eventual structure like:

- api/{file,image} - representation of the file/image instances we'll pass around
- dothething/{file,image} - functions/methods for file/image promotion (bikeshed the name, but there's some generic form of "moving bits from point A to point B")
- cli/{file,image} - logic for `kpromo` file/image commands
- sign/{file,image} - signing/verification for signing operations

^^ note, this is just the way I've been thinking about it to date, but I figure if we're going to move a bunch of things around, it probably makes sense to minimize the amount of times we'll be doing `git mv`. :)

puerco:
Thanks for your notes @justaugustus! I've made some changes to address your observations as follows:
 
#### File names
`common.go, util.go` → these files are now gone. I've moved the functions to their corresponding files 
`scanner.go` → this file is now `implementation-securityscanner.go` it is still here because it contains the path that implements the security scan. We can remove it for sure, but the purpose of this PR is to leave the CLI untouched, clean the API in the middle and also introduce no changes into the core libraries. No bugs,deprecations or features are handled here.
`promoter.go vs promotion.go` this file is now `implementation-imagepromotion.go`  named evenly with the rest

#### Imports
I think the order was right, I've split the import lines into groups to make the grouping clear

#### Top-level packages &&  Parity w/ file promotion package structure

I've moved the new package code to the top and renamed it `imagepromoter` to match the file equivalent. As no libraries or cli files are introduced or deprecated, those stay the same pending further improvements there.

PR #498, opened 2022-02-14
Verify signatures in candidate images before accepting them from staging 

puerco:
#### What type of PR is this?

/kind feature

This PR adds the signature verification code to the container image promoter.

After parsing and calculating the images to promote, the promoter will now check
if there are signatures attached to the images about to be promoted. If no signatures are found, the promoter
carries the promotion as usual. If, on the other hand signatures are attached to the images, we will now verify
them and block the promotion if they don't validate correctly.

Part of https://github.com/kubernetes/release/issues/2383
Related to https://github.com/kubernetes/release/issues/2227

Currently, no images from staging should contain signatures so this PR is effectively a NOOP for current image promotions.

```release-note
- The image promoter will now check for digital signatures in images considered
  for promotion. If signatures are found, they will be verified. If a manifest
  contains images with invalid signatures, the failed verification will cause
  the promotion to fail. All images without signatures are not be verified and
  are accepted for promotion as usual.
```


puerco:
We need to check if images are signed before running the imageverification checks.

PR is in:   https://github.com/kubernetes-sigs/release-sdk/pull/39

puerco:
This should be good for review!
Please take a look @justaugustus @saschagrunert @cpanato @kubernetes-sigs/release-engineering

promoter/image/promoter.go
```diff
@@ -120,15 +125,31 @@ func (p *Promoter) PromoteImages(opts *options.Options) (err error) {
...
-	)
+	if err := p.impl.PromoteImages(sc, promotionEdges, producerFunc); err != nil {
+		return errors.Wrap(err, "running promotion")
+	}
+
+	if err := p.impl.SignImages(opts, sc, promotionEdges); err != nil {
+		return errors.Wrap(err, "signing images")
+	}
+
+	if err := p.impl.WriteSBOMs(opts, sc, promotionEdges); err != nil {
```
cpanato, line 148: should we write the bom before signing the image?

cpanato: or does not matter?

internal/promoter/image/sign.go
```diff
@@ -0,0 +1,83 @@
...
+	reg "sigs.k8s.io/promo-tools/v3/legacy/dockerregistry"
+	options "sigs.k8s.io/promo-tools/v3/promoter/image/options"
+	"sigs.k8s.io/release-sdk/sign"
+)
+
+// ValidateStagingSignatures checks if edges (images) have a signature
+// applied during its staging run. If they do it verifies them and
+// returns an error if they are not valid.
+func (di *DefaultPromoterImplementation) ValidateStagingSignatures(
+	opts *options.Options, sc *reg.SyncContext, edges map[reg.PromotionEdge]interface{},
```
saschagrunert, line 34: If `sc` and `opts` are unused then we should remove them from the arguments.

internal/promoter/image/sign.go
```diff
@@ -0,0 +1,83 @@
...
+			continue
+		}
+
+		// Check the staged image signatures
+		if _, err := signer.VerifyImage(imageRef); err != nil {
+			return errors.Wrapf(
+				err, "verifying signatures of image %s", imageRef,
+			)
+		}
+		logrus.Infof("Signatures for ref %s verfified", imageRef)
```
saschagrunert, line 64: If one image verification fails, then all other images are affected as well and have no chance for promotion. Do we wanna return a list of `reg.PromotionEdge`s which are successfully verified and still promote them instead of aborting the whole process if one image fails?

promoter/image/promoter.go
puerco: Right now there is no difference, but I've been thinking that the BOM could verify the image before generation, becoming the first consumer of the signature.

internal/promoter/image/sign.go
puerco: Removed :+1:

internal/promoter/image/sign.go
saschagrunert: SGTM

puerco:
@justaugustus can you take a look at this one when you get a chance? :pray:

justaugustus (approved):
Nice!!

PR #501, opened 2022-02-15
Sign container images

puerco:
#### What type of PR is this?

/kind feature

This PR introduces digital signing into the promoter.  The signing flow is enabled by default and will use the default kubernetes service account (krel-trust@k8s-artifacts-prod.iam.gserviceaccount.com). 

Signing is controlled by two new flags in `kpromo cip`:

```
  --sign                          when true, sign promoted images (default true)
  --signer-account string         service account to use as signing identity (default "krel-trust@k8s-artifacts-prod.iam.gserviceaccount.com")
```

Image signing involves four steps which begin after promotion. After promoting, kpromo will look for signatures attached to images in the original registry and copy them to the destination registries. It will generate tokens for the signer account to invoke the keyless signing flow. Once the tokens are available, it will sign one of the images and copy the resulting combined signature to the rest of the mirrors (if any).

 Part of: https://github.com/kubernetes/release/issues/2383

 /assign @justaugustus 

Stephen I think this is ready, please take a look when you can!

```release-note
`kpromo` now signs container images using the sigstore keyless flow. Signing identity can be specified with a new flag `--signer-account`. In addition, image signing can be enabled/disabled using `--sign`.  
```


puerco:
We need to pass the signer an external identity token, if not the cli will ask the user to follow the OIDC flow from the browser:
```
Enter the verification code XXXX-XXXX in your browser at: https://oauth2.sigstore.dev/auth/device?user_code=XXX-XXXX
```

This PR enables that: https://github.com/kubernetes-sigs/release-sdk/pull/42

ameukam:
> We need to pass the signer an external identity token, if not the cli will ask the user to follow the OIDC flow from the browser:
> 
> ```
> Enter the verification code XXXX-XXXX in your browser at: https://oauth2.sigstore.dev/auth/device?user_code=XXX-XXXX
> ```
> 
> This PR enables that: [kubernetes-sigs/release-sdk#42](https://github.com/kubernetes-sigs/release-sdk/pull/42)

Do you plan to use Github as an identity provider ?

puerco:
Hey @ameukam sorry, I missed your message last week. No, since we will run everything in GCP the identity will come from GCP service accounts. The idea is that the service account that runs the promoter will impersonate a service account that will eventually do the signing.

I'm thinking that we should have two accounts: One for the actual signing of the artifacts and one that we can impersonate while we run the tests. I still have to decipher which service accounts are working in the promoter and then we need to set permissions and roles for all of them.

puerco:
I tracked a problem where the staging images were getting wiped out when adding the k8s-org signatures to a problem in cosign. Fixed it here:  https://github.com/sigstore/cosign/pull/1616.

Before that PR merges and we pull the commit into our deps, the promoter will copy the signatures from staging but will delete them when re-signing.

internal/promoter/image/sign.go
```diff
@@ -72,6 +99,140 @@ func (di *DefaultPromoterImplementation) ValidateStagingSignatures(
...
+func (di *DefaultPromoterImplementation) copySignatures(
+	src *reg.PromotionEdge, dsts []reg.PromotionEdge,
+) error {
+	sigTag := digestToSignatureTag(src.Digest)
+	sourceRefStr := fmt.Sprintf(
+		"%s/%s:%s", src.DstRegistry.Name, src.DstImageTag.Name, sigTag,
+	)
+	srcRef, err := name.ParseReference(sourceRefStr)
+	if err != nil {
+		return fmt.Errorf("parsing reference %q: %w", sourceRefStr, err)
```
saschagrunert, line 205: non-blocking nit: either use consistently `errors.Wrap` or wrapping with `%w`.

internal/promoter/image/sign.go
```diff
@@ -82,3 +243,53 @@ func (di *DefaultPromoterImplementation) WriteSBOMs(
...
+
+	// If SignerInitCredentials, initialize the iam client using
+	// the identityu in that file instead of Default Application Credentials
+	if opts.SignerInitCredentials != "" {
+		logrus.Infof("Using credentials from %s", opts.SignerInitCredentials)
+		credOptions = []gopts.ClientOption{
+			gopts.WithCredentialsFile(opts.SignerInitCredentials),
+		}
+	}
+	ctx := context.Background()
```
saschagrunert, line 272: non-blocking nit: consider using a context with timeout here.

saschagrunert (approved):
Great work! I'll not block in my two findings.

PR #504, opened 2022-02-19
WIP Implement Auth with GGCR

puerco:
#### What type of PR is this?
/kind cleanup

Current authentication in the promoter is handled by a token obtained from a gcloud call. This commit adds additional logic to support authentication via keychains in GGCR to allow the promoter to work with other registries outside of GCR.

When defining a service account for a registry, the token authentication will continue as usual. When a service account is not defined, requests to the registries will attempt to use the docker config entries to authenticate.

None

/cc @kubernetes/release-engineering  

```release-note
The image promoter can now authenticate to other registries using the ggcr/authn keychains 
```


puerco:
After discussing for the last one or two weeks, we decided not to continue with this approach.

/close

PR #505, opened 2022-02-22
Implement sc.ReadRegistries in GGCR 

puerco:
#### What type of PR is this?
/kind cleanup

This PR introduces a parallel implementation of the ReadRegistries function written with `google/go-containerregistry`.

The new function exists alongside the old one and the idea is to start replacing each of the calls until we replace all invocation and we can drop the old one. 

The new function is now wired into the snapshot code and tests are passing. This means that snapshots can now be done using other authentication methods read from the local docker config:

```
kpromo cip --snapshot=gcr.io/k8s-staging-ingress-nginx --confirm
```

None

The new function has support for full recursion of GCR registries. Non-recursive reads are not yet implemented.

/cc @justaugustus 

```release-note
There is a new implementation of the registry inventorying function using `google/go-containerregistry`. The snapshot code is now wired to use the new implementation.  
```


legacy/dockerregistry/inventory.go
```diff
@@ -1135,6 +1138,78 @@ func GetTokenKeyDomainRepoPath(registryName RegistryName) (key, domain, repoPath
 	return key, s[:i], s[i+1:]
 }
 
+func (sc *SyncContext) ReadRegistriesGGCR(
+	toRead []RegistryContext, recurse bool,
```
saschagrunert, line 1142: `toRead` seems unused and could be removed.

saschagrunert (approved):
Just a nit, but this can be addressed later on.

PR #527, opened 2022-04-02
kpromo pr: Support for non-SemVer image tags

wespanther:
#### What type of PR is this?

/kind bug

This PR removes the restriction for image tags to adhere to SemVer. Taking debian-base in in k8s-staging-build-image, we cannot match any of the tags, at present:

image

Since the filter requires an explicit match, I don't think a SemvVer validation buys us anything. With this change, we can make promotion PRs for arbitrary tags (like latest, etc.) assuming they exist.

Fixes https://github.com/kubernetes-sigs/promo-tools/issues/520

```release-note
kpromo pr: Support non-SemVer image tags
```

### DIFF
--- a/cmd/kpromo/cmd/pr/pr.go
+++ b/cmd/kpromo/cmd/pr/pr.go
@@ -83,13 +83,6 @@ func (o *promoteOptions) Validate() error {
 		return errors.Wrap(err, "checking user's fork")
 	}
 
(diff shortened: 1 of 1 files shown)


justaugustus (approved):
Awesome, thanks @wespanther!!
/ok-to-test

PR #529, opened 2022-04-04
Add `latest-canary` image tag and manifests for canary promotion jobs

justaugustus:
#### What type of PR is this?

/kind feature

- canary: Add canary manifests from `sigs.k8s.io/bom`
- canary: Use `us.gcr.io/k8s-cip-test-prod/canary/bom` as destination
- Add `latest-canary` image tag for canary promotion jobs

/assign @cpanato @puerco 
cc: @kubernetes/release-engineering 
ref: https://github.com/kubernetes/test-infra/pull/25857, https://github.com/kubernetes/k8s.io/pull/3588, https://github.com/kubernetes-sigs/promo-tools/issues/523

```release-note
NONE
```

### DIFF
--- a/canary/images/k8s-staging-bom/images.yaml
+++ b/canary/images/k8s-staging-bom/images.yaml
@@ -1,0 +1,4 @@
+- name: bom
+  dmap:
+    "sha256:631ff0f765b4b1ff0b3c81d73698f557f2291aebc822eedf4d265a9ec611af42": ["v0.3.0-rc1"]
+
--- a/canary/images/k8s-staging-bom/images.yaml
+++ b/canary/images/k8s-staging-bom/images.yaml
@@ -1,4 +1,3 @@
 - name: bom
   dmap:
     "sha256:631ff0f765b4b1ff0b3c81d73698f557f2291aebc822eedf4d265a9ec611af42": ["v0.3.0-rc1"]
-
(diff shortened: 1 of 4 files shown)


puerco:
/lgtm
(:eye:  Keeping an eye on the linter in https://github.com/kubernetes/test-infra/pull/25857 )

puerco:
@justaugustus the job failed, I was looking at the GCP project and I think we need to do some changes to this strategy:

* First, to use this service account, we would need to use this credential file in order to run as k8s-infra-gcr-promoter@k8s-cip-test-prod.iam.gserviceaccount.com: https://github.com/kubernetes/test-infra/blob/0861cb47b95d14c5c71c6a124b10870574a78787/config/jobs/kubernetes-sigs/promo-tools/promo-tools.yaml#L18-L36 

* Second, note that [no svc account in the cluster is annotated](https://github.com/kubernetes/test-infra/blob/1f10106176e698a36de040c5c86e8e55a4c0a1f4/config/prow/cluster/trusted_serviceaccounts.yaml) to impersonate `k8s-infra-gcr-promoter@k8s-cip-test-prod` so we cannot use workload identity. Even if we manage to somehow run as the account, I would prefer to run the canaries in the same way as the real promotions (using workload identity).
* Finally, this SA does not have access to the signer identity. SO it will fail when trying to impersonate it to sign the canaries.

My suggestion is that we run the canaries with the real promotion service account, using workload identity, and we promote the images to another registry where it has access (not sure which though). WDYT?

PR #609, opened 2022-08-17
Add an authenticated client to gcrane

upodroid:
#### What type of PR is this?
/kind bug

gcrane is copying to target registries without being authenticated and those pushes fail for Artifact Registry.

```release-note
crane copy operations in kpromo now use `gcrane.Keychain` to authenticate.  
```

https://kubernetes.slack.com/archives/CJH2GBF7Y/p1660658716152579

### DIFF
--- a/internal/legacy/dockerregistry/inventory.go
+++ b/internal/legacy/dockerregistry/inventory.go
@@ -1750,7 +1750,10 @@ func (sc *SyncContext) Promote(
 						)
 					}
 
-					if err := crane.Copy(srcVertex, dstVertex); err != nil {
+					opts := []crane.Option{
+						crane.WithAuthFromKeychain(gcrane.Keychain),
+					}
+					if err := crane.Copy(srcVertex, dstVertex, opts...); err != nil {
 						logrus.Error(err)
 						errors = append(
 							errors,
--- a/internal/promoter/image/sign.go
+++ b/internal/promoter/image/sign.go
@@ -26,6 +26,7 @@ import (
 
 	credentials "cloud.google.com/go/iam/credentials/apiv1"
 	"github.com/google/go-containerregistry/pkg/crane"
+	"github.com/google/go-containerregistry/pkg/gcrane"
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/nozzle/throttler"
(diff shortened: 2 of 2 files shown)


puerco:
It's seem the change breaks the authentication using json tokens :\

upodroid:
We are loading the creds in a nonstandard fashion.

```
      - name: CIP_E2E_KEY_FILE
        value: /etc/k8s-cip-test-prod-service-account/service-account.json
```

needs to be 

```
      - name: GOOGLE_APPLICATION_CREDENTIALS
        value: /etc/k8s-cip-test-prod-service-account/service-account.json
```

My change configures an ADC to be used which is a bit tricky on Prow.

https://pkg.go.dev/golang.org/x/oauth2/google#FindDefaultCredentialsWithParams

```
It looks for credentials in the following places, preferring the first location found:

1. A JSON file whose path is specified by the GOOGLE_APPLICATION_CREDENTIALS environment variable. For workload identity federation, refer to https://cloud.google.com/iam/docs/how-to#using-workload-identity-federation on how to generate the JSON configuration file for on-prem/non-Google cloud platforms.
2. A JSON file in a location known to the gcloud command-line tool. On Windows, this is %APPDATA%/gcloud/application_default_credentials.json. On other systems, $HOME/.config/gcloud/application_default_credentials.json.
3. On Google App Engine standard first generation runtimes (<= Go 1.9) it uses the appengine.AccessToken function.
4. On Google Compute Engine, Google App Engine standard second generation runtimes (>= Go 1.11), and Google App Engine flexible environment, it fetches credentials from the metadata server.
``` 

It is using 4 which is a google service account that should have no privileges.

cpanato (approved):
/lgtm

/hold
/assign @puerco

puerco:
All right, let's give this a go, thanks for solving it on prow too, thanks @upodroid ! :)

/hold cancel

Since this is the same change as #632 which is also passing now with the ADC set.
/cc  @imjasonh

upodroid:
While we are here, let's do the same thing at https://github.com/kubernetes-sigs/release-sdk/pull/105

After that, we can fully remove `gcloud` from the `image-promoter`

PR #646, opened 2022-10-20
Use new ImagesSigned() API from release-sdk

saschagrunert:
#### What type of PR is this?

/kind cleanup

Using the new API introduced in https://github.com/kubernetes-sigs/release-sdk/pull/123

Refers to https://github.com/kubernetes-sigs/promo-tools/issues/637

```release-note
None
```


internal/promoter/image/sign.go
```diff
@@ -53,69 +52,51 @@ const (
...
+		}
 
-			if !isSigned {
-				t.Done(nil)
-				return
+		if isSigned {
+			edge, ok := refsToEdges[ref]
+			if !ok {
+				logrus.Errorf("Reference %s is not in edge map", ref)
+				return true
```
xmudrii, line 93: We log an error, but we return true here. Is this as expected? In other cases where we log an error, we return false.

xmudrii (commented):
Just one question, I assume it's okay, but I want to double-check to be sure.

internal/promoter/image/sign.go
saschagrunert: Yes, it is intended to continue the loop in that case. Usually this can only happen if the release sdk implementation is wrong.

go.mod
```diff
@@ -31,7 +31,7 @@ require (
 	gopkg.in/src-d/go-git.v4 v4.13.1
 	gopkg.in/yaml.v2 v2.4.0
 	k8s.io/utils v0.0.0-20220728103510-ee6ede2d64ed
-	sigs.k8s.io/release-sdk v0.9.3
+	sigs.k8s.io/release-sdk v0.9.4-0.20221020144501-0f5793938dbc
```
cpanato, line 34: we should do a sdk release asap

cpanato (approved):
i think that looks good

/lgtm

go.mod
saschagrunert: Let’s wait for https://github.com/kubernetes-sigs/release-sdk/issues/124

PR #648, opened 2022-10-20
set useragent

### LINKED ISSUE #647: image promoter should set a useragent

Ideally, the image promoter would have a unique useragent so we can identify what requests it's sending to the registry (in particular, to debug why you're getting 429s).

#### Why is this needed:

See https://github.com/kubernetes-sigs/promo-tools/issues/637#issuecomment-1286180731

cpanato:
#### What type of PR is this?

/kind feature

- set useragent

Fixes #647 

/assign @saschagrunert @puerco @jeremyrickard @Verolop 
cc @BenTheElder @jonjohnsonjr 

```release-note
set useragent
```

### DIFF
--- a/internal/legacy/dockerregistry/inventory.go
+++ b/internal/legacy/dockerregistry/inventory.go
@@ -839,11 +839,11 @@ func (sc *SyncContext) recordFoundTags(
 // We use the term "registry" to mean the "root repository" in this program, but
 // to be technically correct, for gcr.io/google-containers/foo/bar/baz:
 //
-//  - gcr.io is the registry
-//  - gcr.io/google-containers is the toplevel repository (or "root" repo)
-//  - gcr.io/google-containers/foo is a child repository
-//  - gcr.io/google-containers/foo/bar is a child repository
-//  - gcr.io/google-containers/foo/bar/baz is a child repository
+//   - gcr.io is the registry
+//   - gcr.io/google-containers is the toplevel repository (or "root" repo)
+//   - gcr.io/google-containers/foo is a child repository
+//   - gcr.io/google-containers/foo/bar is a child repository
+//   - gcr.io/google-containers/foo/bar/baz is a child repository
 //
 // It may or may not be the case that the child repository is empty. E.g., if
 // only one image gcr.io/google-containers/foo/bar/baz:1.0 exists in the entire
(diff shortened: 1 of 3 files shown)


internal/legacy/dockerregistry/inventory.go
```diff
@@ -839,11 +839,11 @@ func (sc *SyncContext) recordFoundTags(
...
-//  - gcr.io is the registry
-//  - gcr.io/google-containers is the toplevel repository (or "root" repo)
-//  - gcr.io/google-containers/foo is a child repository
-//  - gcr.io/google-containers/foo/bar is a child repository
-//  - gcr.io/google-containers/foo/bar/baz is a child repository
+//   - gcr.io is the registry
+//   - gcr.io/google-containers is the toplevel repository (or "root" repo)
+//   - gcr.io/google-containers/foo is a child repository
+//   - gcr.io/google-containers/foo/bar is a child repository
+//   - gcr.io/google-containers/foo/bar/baz is a child repository
```
cpanato, line 846: this came from go fmt 🤷

internal/legacy/dockerregistry/inventory.go
```diff
@@ -1752,6 +1752,7 @@ func (sc *SyncContext) Promote(
 
 					opts := []crane.Option{
 						crane.WithAuthFromKeychain(gcrane.Keychain),
+						crane.WithUserAgent("k8s-image-promoter"),
```
puerco, line 1755: Can we const this away in types/image?

```suggestion
						crane.WithUserAgent(image.UserAgent),
```

puerco (commented):
Thanks Carlos, one small nit:

internal/legacy/dockerregistry/inventory.go
puerco: And can we name it `kpromo`  :pray:

cpanato: done, thanks for the feedback

BenTheElder (commented):
should we add `-A` to the curl call?

jonjohnsonjr:
> should we add `-A` to the curl call?

Similarly, it would be good to plumb something for the useragent down into the signature code.

saschagrunert (approved):

cpanato:
> should we add `-A` to the curl call?

which curl call? 🤔

BenTheElder:
https://github.com/kubernetes-sigs/promo-tools/blob/5e139353c19b2fe0b0a3c53e4f31a74262a0a2b4/internal/legacy/dockerregistry/registry/registry.go#L235

on my TODO to file a bug, better answer is to switch to net/http here too but ... too many things going on

PR #661, opened 2022-11-02
Take digests into account for prow diff

saschagrunert:
#### What type of PR is this?

/kind feature

We now also take the possible digests to promote into account rather than the whole image file. This should provide an additional performance boost especially for large manifests.

Refers to https://github.com/kubernetes-sigs/promo-tools/issues/637 and https://github.com/kubernetes/k8s.io/issues/4374

None

```release-note
Filter digests if `--use-prow-manifest-diff` is used.
```

### DIFF
--- a/image/manifest/manifest.go
+++ b/image/manifest/manifest.go
@@ -19,7 +19,7 @@ package manifest
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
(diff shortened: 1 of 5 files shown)


saschagrunert:

cpanato (approved):
/lgtm

PR #691, opened 2022-12-07
Add support for batch jobs

xmudrii:
#### What type of PR is this?

/kind feature

Prow can merge PRs in batches. In that case, only one set of presubmits is run and that set of presubmits tests all the PRs in a batch. This can be a very effective mechanism to speed up presubmits and reduce the time needed to merge PRs.

However, when presumbits are run in a batch, they're called `batch` instead of `presubmit`. This means that `prow.k8s.io/type` is `batch` instead of `presubmit`.

This difference is causing `kpromo` to fail with the following error:

```
time="12:43:05.387" level=info msg="PromoteImages start" diff=0s
time="12:43:05.387" level=warning msg="Not setting a service account" diff=0s
time="12:43:05.920" level=info msg="Parsing manifests" diff=533ms
time="12:43:05.921" level=info msg="Using prow diff" diff=0s
time="12:43:05.921" level=fatal msg="run `cip run`: promote images: parsing manifests: parsing thin manifest directory: get prow diff files: unknown job type: batch" diff=0s
```

This PR adds support for batch jobs to kpromo. This PR treats those jobs the same as presubmits. It should work because batch jobs also have the pull base ref (`PULL_BASE_SHA`) which we need in presubmits.

None

This is not tested because I'm not really sure how can we test this.

```release-note
Add support for Batch ProwJobs
```

### DIFF
--- a/internal/legacy/dockerregistry/schema/manifest.go
+++ b/internal/legacy/dockerregistry/schema/manifest.go
@@ -333,6 +333,7 @@ func diffProwFiles(dir string) (digests []string, err error) {
 		jobTypeEnv        = "JOB_TYPE"
 		jobTypePostsubmit = "postsubmit"
 		jobTypePresubmit  = "presubmit"
+		jobTypeBatch      = "batch"
 	)
 
 	jobType := os.Getenv(jobTypeEnv)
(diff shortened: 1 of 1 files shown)


cpanato (approved):
/lgtm

saschagrunert:
Makes sense, LGTM

PR #704, opened 2023-01-17
Support syncing to S3

justinsb:
Initial support for kpromo to sync to S3 buckets.

```release-note
Added initial support for kpromo file synchronization to S3 buckets.
```


justaugustus (commented):
@justinsb -- AFK for the moment, but can you check if it would instead make sense to implement portions of this within https://github.com/kubernetes-sigs/release-sdk/blob/main/object/store.go?

justinsb:
@justaugustus  I'm not sure that API is a great match here, as it looks like in order to copy files across providers we would need providers to know about each other.   (We have an pretty broad vfs library [here](https://github.com/kubernetes/kops/tree/master/util/pkg/vfs) if you want generic functionality , I think the abstractions there are a bit more evolved.)

But for this PR, the real nuance is in carefully setting the metadata etc to make the common case fast (i.e. the case when nothing has to be copied).  That's why we have to use a non-parallel object upload, so that the ETag is the MD5 hash.  I think for integrity purposes we also want to upload additional hashes as metadata, but we're really very optimized for our particular use-case here.

I'm happy to evolve the API in release-sdk, but I'd probably steer it towards something with objects paths as first class concepts; WDYT about just sharing the vfs layer with kOps?

justaugustus:
> @justaugustus I'm not sure that API is a great match here, as it looks like in order to copy files across providers we would need providers to know about each other. (We have an pretty broad vfs library [here](https://github.com/kubernetes/kops/tree/master/util/pkg/vfs) if you want generic functionality , I think the abstractions there are a bit more evolved.)

TIL about https://github.com/kubernetes/kops/tree/master/util/pkg/vfs!

> I'm happy to evolve the API in release-sdk, but I'd probably steer it towards something with objects paths as first class concepts; WDYT about just sharing the vfs layer with kOps?

@justinsb -- Happy for you to move forward using kOps vfs here, if it's going to dedupe some of the work.

Are changes to `kops/util/pkg` generally minimal across Kubernetes major versions? (My concern being tying the release engineering tooling to the k/k versioning complexities.)
A way to minimize this concern could be to donate `vfs` (and/or some larger subset of the `util`) package to sigs.k8s.io/release-sdk. Would that be something y'all would be open to?

justinsb:
Definitely open to putting vfs somewhere shared :-)  It's already shared between etcdadm and kOps, so it is generally useful.  It's pretty stable although it's recently seen a flood of activity because we're adding context.Context as part of the great logging/tracing rewrite.

I was also thinking though that the API we have here is probably a reasonable minimal API, and we have higher level functionality (like folder-to-folder syncing) in kpromo.  So we could also just move the filestores.

I think two good options are to put kOps VFS in a standalone repo somewhere (because it's not really only release tooling, either) or to sync kpromo's file abstractions with release-sdk (they are more logically coupled).

I would ask though if we can merge this PR in the interim though, it's all part of the S3 mirroring that reduces the CNCF expenditure on egress bandwidth - we finally have AWS buckets :-)

justaugustus:
> I would ask though if we can merge this PR in the interim though, it's all part of the S3 mirroring that reduces the CNCF expenditure on egress bandwidth - we finally have AWS buckets :-)

Yep, let's just import what makes sense here and minimize the pain!

justinsb:
Thanks @justaugustus ... I was thinking that this and release are actually pretty specific functionality, for example I think we want to upload in a potentially suboptimal way to preserve metadata, we might want to force the metadata always to be created etc, so it belongs in a library that promo-tools and the rest of the release tooling share,it's not a generic VFS layer.

So I propose we merge this here, and then I'll submit some refactorings both to release-sdk and here so that we can have a shared library in release-sdk and use it here (i.e. I'll try upstream the code here and refactor release-sdk to avoid calling gsutil where it isn't applicable).  I haven't poked around where all the places release-sdk is used yet though, not entirely sure what I'm signing up for yet :-)

promoter/file/filestore.go
```diff
@@ -62,6 +61,25 @@ func openFilestore(
...
+		return openGCSFilestore(ctx, filestore, useServiceAccount, confirm)
+	}
+	if strings.HasPrefix(filestore.Base, "s3://") {
+		return openS3Filestore(filestore)
+	}
+
+	return nil, fmt.Errorf(
+		"unrecognized scheme %q (supported schemes: s3, gs)",
+		filestore.Base,
+	)
```
justaugustus, line 75: Let's define something like:

```go
var supportedURISchemes := []string{
  "s3",
  "gs",
}
```

And then:

```go
	return nil, fmt.Errorf(
		"unrecognized scheme %q (supported schemes: %s)",
		filestore.Base,
                strings.Join(supportedURISchemes, ", "),
	)
```

promoter/file/filestore.go
```diff
@@ -73,11 +91,7 @@ func openFilestore(
 	}
 
 	if u.Scheme != "gs" {
-		return nil, fmt.Errorf(
-			"unrecognized scheme %q (supported schemes: %s)",
-			filestore.Base,
-			object.GcsPrefix,
-		)
+		return nil, fmt.Errorf("unrecognized scheme %q, expected gs://", filestore.Base)
```
justaugustus, line 94: Same comment about `object.GcsPrefix` as above.

justaugustus (approved):
@justinsb -- Left some nits. If you want to fix them here, go for it.

If you want to do these as a follow-up, feel free to release the hold.
/hold

promoter/file/filestore.go
justinsb: Done - put these into the api, as I think they are _our_ values, not the library's values (i.e. if the library changed the value, we would not want to change our values!)

promoter/file/filestore.go
justinsb: Created a register of providers, so that everything stays in sync here.  I think that's the underlying concern, and I agree with it :+1:

justaugustus (approved):
Thanks @justinsb!

PR #722, opened 2023-02-08
S3 promotion: support anonymous authentication

justinsb:
If we don't have AWS credentials (or don't want to use them), we need
to actively specify that rather than attempt credential discovery
which then fails with an error.

Reuse the `--use-service-acccount=false` flag for this meaning.

```release-note

When --use-service-account=false (which is the default), we will issue anonymous requests to AWS and ignore any AWS credentials.  This should be sufficient for dry-run, and ensures that dry-run is safe even if manifests are hostile.
```

### DIFF
--- a/promoter/file/s3.go
+++ b/promoter/file/s3.go
@@ -30,8 +30,10 @@ import (
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 	"github.com/sirupsen/logrus"
 
 	api "sigs.k8s.io/promo-tools/v3/api/files"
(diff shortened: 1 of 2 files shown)


justinsb:
/retest

I'm wondering if I ran concurrent e2e jobs, as the failure looks very unrelated

promoter/file/s3.go
```diff
@@ -111,28 +117,16 @@ func (p *s3Provider) findRegionForBucket(ctx context.Context, bucket string) (st
 		lookupRegion = "us-east-2"
 	}
 
-	awsConfig := aws.NewConfig()
-	awsConfig = awsConfig.WithRegion(lookupRegion)
-
-	awsSession, err := session.NewSession(awsConfig)
+	// This is an unauthenticated request (it just does a HEAD on the bucket),
+	// so we don't need to be careful about credentials.
+	awsSession, err := session.NewSession()
```
xmudrii, line 122: Nit: since this is an unauthenticated request, should we enforce using the anonymous credentials (something like we have above)? If I’m not mistaken, this function is going to use credentials if they are available and I’m wondering if we should avoid that.

justinsb: Great point - it's not a risk worth taking :-)  Fixed!

saschagrunert (approved):

PR #642, opened 2022-10-19
Remove the `credHelpers` and use the golang native option.

upodroid:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:
Don't use the `docker-credential-gcloud` credHelper

#### Which issue(s) this PR fixes:

Part of #637 

#### Special notes for your reviewer:

I'm thinking of creating an beta version of pull-k8sio-cip with this image. In theory, it should be faster.

#### Does this PR introduce a user-facing change?
Yes

```release-note
stop using `docker-credential-gcloud` in the image
```

/cc @BenTheElder @puerco

PR #740, opened 2023-02-22
Add remaining AR Regions

upodroid:
For now, I intend on adding 2 new regions in March which are us-west3 and us-west4. 

That infra change was approved in https://github.com/kubernetes/k8s.io/pull/4419

I'm adding the other regions for now so I don't have to comeback and cut a new kpromo.

Also, I still need to work on removing the credhelpers #642 

For the registry freeze, we need `gcr.io`, `us-west3-docker.pkg.dev` and `us-west3-docker.pkg.dev` configured.

/cc @puerco @ameukam 

```release-note
added all the Artifact Registry locations available in Feb 2023 to the docker credHelpers for future use.
```


ameukam:
We skip `europe-west3` on purpose ?

upodroid:
No, pushing a commit to fix it

saschagrunert (approved):

PR #760, opened 2023-03-06
Replace k8s.gcr.io references with registry.k8s.io

xmudrii:
#### What type of PR is this?

/kind cleanup
/kind api-change

This PR does three things as part of https://github.com/kubernetes/k8s.io/issues/4738:

- Add registry.k8s.io to `internal/legacy/dockerregistry`
  - I'm not sure if that's needed or how's that code used at all, but we have k8s.gcr.io there and looking by how that function is used, it makes sense _to me_
  - I'm happy to revert this commit if needed because I'm not sure if it makes any sense
  - I also added some tests cases based on `registry.k8s.io`
- Use registry.k8s.io in dependencies.yaml
  - This should be just a cosmetic change
- Point ProdRegistry to registry.k8s.io and add ProdRegistryLegacy that points to k8s.gcr.io
  - This is the most important change. We use this const in krel when validating images: https://github.com/kubernetes/release/blob/eea315df4938e5687ae571eee1019c229eb3575d/pkg/anago/release.go#L419
  - As we'll not push to k8s.gcr.io any longer, we need to change it to registry.k8s.io otherwise we'll fail to validate if images are pushed
  - I thought the best way to accomplish this is to change the underlying const
  - To cover the cases where we need the old value, I added ProdRegistryLegacy. I also did a search with cs.k8s.io to ensure that no other project uses ProdRegistry variable

There are some other `k8s.gcr.io` references, but those make sense and are still needed, so I left them in place.

xref https://github.com/kubernetes/k8s.io/issues/4738

```release-note
ProdRegistry const is changed to `registry.k8s.io`. A new const called ProdRegistryLegacy is introduced and it points to `k8s.gcr.io`
```

### DIFF
--- a/cmd/kpromo/cmd/pr/pr.go
+++ b/cmd/kpromo/cmd/pr/pr.go
@@ -203,7 +203,7 @@ func runPromote(opts *promoteOptions) error {
 
 	// Path to the promoter image list
 	imagesListPath := filepath.Join(
-		image.ProdRegistry,
+		image.ProdRegistryLegacy,
 		"images",
 		filepath.Base(image.StagingRepoPrefix)+opts.project,
 		"images.yaml",
@@ -224,7 +224,7 @@ func runPromote(opts *promoteOptions) error {
 
 		opt := manifest.GrowOptions{}
 		if err := opt.Populate(
-			filepath.Join(repo.Dir(), image.ProdRegistry),
+			filepath.Join(repo.Dir(), image.ProdRegistryLegacy),
 			image.StagingRepoPrefix+opts.project, opts.images, opts.digests, opts.tags); err != nil {
 			return fmt.Errorf("populating image promoter options for tag %s with image filter %s: %w", opts.tags, opts.images, err)
 		}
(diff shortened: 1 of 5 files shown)


cmd/kpromo/cmd/pr/pr.go
```diff
@@ -203,7 +203,7 @@ func runPromote(opts *promoteOptions) error {
 
 	// Path to the promoter image list
 	imagesListPath := filepath.Join(
-		image.ProdRegistry,
+		image.ProdRegistryLegacy,
```
cpanato, line 206: should we start using the new one and stop using the old?

xmudrii: No, this requires changing the layout of the k/k8s.io repo and that's something that's way beyond scope of this PR.

xmudrii:
Slack discussion: https://kubernetes.slack.com/archives/CJH2GBF7Y/p1678120596620139

upodroid:
Lets discuss https://github.com/kubernetes/release/issues/2947 before merging this one.

Breaking k/k8s.io repo is fine as we need to freeze the image registries anyway and move the manifests to the new folder.

xmudrii:
@upodroid will take care of replacing `k8s.gcr.io` references as part of https://github.com/kubernetes/release/issues/2947, so I'll close this PR.
/close

PR #771, opened 2023-03-16
Rate limit crane.Copy() operations

puerco:
#### What type of PR is this?

/kind feature

This PR adds a new `ratelimiter` packaged based on [geranos' roundtripper](https://github.com/kubernetes/registry.k8s.io/blob/main/cmd/geranos/ratelimitroundtrip.go).

Based on it, we now rate limit the crane operations in the signature package to ensure
they are under the [AR quotas set up by k8s infra](https://github.com/kubernetes/registry.k8s.io/issues/153#issuecomment-1460913153).

Update: I've added the rate limit to the image promotion copy operations too.

(Thanks @BenTheElder ! )

Part of https://github.com/kubernetes/release/issues/2962

/cc @BenTheElder @xmudrii @palnabarun @ameukam 

More context in [this slack thread](https://kubernetes.slack.com/archives/CJH2GBF7Y/p1678895816204009?thread_ts=1678861049.737169&cid=CJH2GBF7Y)

```release-note
Concurrent copy operations are now rate limited when promoting and replicating and mirroring signatures using a new `ratelimiter` package based on the [geranos rate limiter](https://github.com/kubernetes/registry.k8s.io/blob/main/cmd/geranos/ratelimitroundtrip.go) (Thanks @BenTheElder )
```

### DIFF
--- a/internal/promoter/image/sign.go
+++ b/internal/promoter/image/sign.go
@@ -97,7 +97,7 @@ func (di *DefaultPromoterImplementation) CopySignatures(
 ) error {
 	// Cycle all signedEdges to copy them
 	logrus.Infof("Precopying %d previous signatures", len(signedEdges))
-	rtripper := ratelimit.NewRoundTripper(ratelimit.MaxEvents)
+
 	t := throttler.New(opts.MaxSignatureCopies, len(signedEdges))
 	for e := range signedEdges {
 		go func(edge reg.PromotionEdge) {
@@ -124,7 +124,7 @@ func (di *DefaultPromoterImplementation) CopySignatures(
 			opts := []crane.Option{
 				crane.WithAuthFromKeychain(gcrane.Keychain),
 				crane.WithUserAgent(image.UserAgent),
-				crane.WithTransport(rtripper),
+				crane.WithTransport(ratelimit.Limiter),
 			}
 			if err := crane.Copy(srcRef.String(), dstRef.String(), opts...); err != nil {
 				t.Done(fmt.Errorf(
@@ -287,15 +287,13 @@ func (di *DefaultPromoterImplementation) replicateSignatures(
 		}{ref, dsts[i].DstRegistry.Token})
 	}
 
-	rtripper := ratelimit.NewRoundTripper(ratelimit.MaxEvents)
-
 	// Copy the signatures to the missing registries
 	for _, dstRef := range dstRefs {
 		logrus.WithField("src", srcRef.String()).Infof("replication > %s", dstRef.reference.String())
 		opts := []crane.Option{
 			crane.WithAuthFromKeychain(gcrane.Keychain),
 			crane.WithUserAgent(image.UserAgent),
-			crane.WithTransport(rtripper),
+			crane.WithTransport(ratelimit.Limiter),
 		}
 		if err := crane.Copy(srcRef.String(), dstRef.reference.String(), opts...); err != nil {
 			return fmt.Errorf(
--- a/internal/promoter/image/sign.go
+++ b/internal/promoter/image/sign.go
@@ -45,12 +45,6 @@ const (
 	signatureTagSuffix = ".sig"
 
 	TestSigningAccount = "k8s-infra-promoter-test-signer@k8s-cip-test-prod.iam.gserviceaccount.com"
-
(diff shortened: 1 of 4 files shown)


puerco:
:thinking: 
/test pull-cip-unit-tests

internal/promoter/image/sign.go
```diff
@@ -96,6 +103,7 @@ func (di *DefaultPromoterImplementation) CopySignatures(
 ) error {
 	// Cycle all signedEdges to copy them
 	logrus.Infof("Precopying %d previous signatures", len(signedEdges))
+	rtripper := ratelimit.NewRoundTripper(rateLimitEvents)
```
BenTheElder, line 106: This object is thread safe, maybe a global until we can more easily plumb through a single copy?

promoter/image/ratelimit/roundtripper.go
```diff
@@ -0,0 +1,53 @@
...
+	return &RoundTripper{
+		rateLimiter:  rate.NewLimiter(limit, burst),
+		roundTripper: http.DefaultTransport,
+	}
+}
+
+func (rt *RoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
+	// only rate limit read type calls, not writes
+	if r.Method == http.MethodGet || r.Method == http.MethodHead ||
+		r.Method == http.MethodPut || r.Method == http.MethodPost {
```
BenTheElder, line 46: Put / Post should be write calls?

puerco: I removed the comment but I was wondering if we should sniff at all :thinking:

internal/promoter/image/sign.go
puerco: I was just moving it to a global :) And also added the rate limiter to the promotion crane calls.

promoter/image/ratelimit/roundtripper.go
BenTheElder: Don't we only want to limit read calls? Put / Post should count as Write and have a different rate limit (none atm).

internal/promoter/image/sign.go
```diff
@@ -46,6 +47,12 @@ const (
 	TestSigningAccount = "k8s-infra-promoter-test-signer@k8s-cip-test-prod.iam.gserviceaccount.com"
 )
 
+var rtripper *ratelimit.RoundTripper
```
BenTheElder, line 50: Maybe move both to round trip package so there's a single process-scoped global?

most ideal would be instancing once and passing to everywhere but that may require breaking changes etc...

BenTheElder: Unless the process authenticates with multiple accounts or something

promoter/image/ratelimit/roundtripper.go
puerco: Ah gotcha I thought they had the same limit, reverted to the original.

internal/promoter/image/sign.go
puerco: OK I've moved them to a `ratelimiter.Limiter` global :eyes:

promoter/image/ratelimit/roundtripper.go
BenTheElder: aha. docs: https://cloud.google.com/artifact-registry/quotas

> 60000 requests per minute in each region or multi-region.
> 18000 write or delete requests per minute in each region or multi-region.

these are the shared baselines that can be raised by request (and we have raised some of them), then we have the per-user/per-ip read quota configured but not the per-user write quota (since kpromo should be the _only_ thing performing writes/deletes anyhow)

BenTheElder: https://github.com/kubernetes/registry.k8s.io/issues/153#issuecomment-1452753165
> Requests per project per user per minute per user
> Write requests per project per user per minute per user

https://github.com/kubernetes/registry.k8s.io/issues/153#issuecomment-1460913153
> Requests per project per user per minute per user  set to 5000.

BenTheElder (commented):
/lgtm

TODO: see if we can pass the crane transport (ratelimit.RondTripper) around instead of global, but hopefully the global enables us to quickly integrate this everywhere in crane calls.

we went through that transition in geranos as well. quicker to get in place with the global, refactor later.

PR #792, opened 2023-03-23
Replace k8s.gcr.io references with registry.k8s.io

pkprzekwas:
#### What type of PR is this?

/kind cleanup

As `k8s.gcr.io` registry is now deprecated we need to update registry references to `registry.k8s.io`.

Fixes https://github.com/kubernetes/k8s.io/issues/4738

```release-note
NONE
```


docs/image-promotion.md
```diff
@@ -280,4 +280,4 @@ check to occur.
 [k/k8s.io]: https://git.k8s.io/k8s.io
```
pkprzekwas, line 280: I'm not sure why https://git.k8s.io/k8s.io/k8s.gcr.io/Vanity-Domain-Flip.md#prow-integration was not duplicated in  https://git.k8s.io/k8s.io/registry.k8s.io .

BenTheElder: Because we're already promoting to both backends simultaneously. We'll stop promoting to k8s.gcr.io backends soon but that's been on pause with the redirect rolling out. More discussion in k8s infra and the associated KEP

BenTheElder: The vanity domain flip is also a historical document for a different process involving GCR => GCR. This time it's GCR => Kubernetes DIY Alias in front of AR,S3

docs/image-promotion.md
xmudrii, line 1: TODO as a follow-up: check if we need to update this document because of k8s.gcr.io freeze.

internal/legacy/dockerregistry/inventory.go
```diff
@@ -692,7 +692,7 @@ func ParseContainerParts(s string) (
 		}
 		return strings.Join(parts[0:2], "/"), strings.Join(parts[2:], "/"), nil
 	default:
-		if parts[0] != "k8s.gcr.io" && parts[0] != "staging-k8s.gcr.io" {
+		if parts[0] != "registry.k8s.io" && parts[0] != "staging-k8s.gcr.io" {
```
xmudrii, line 695: I'm wondering if we should keep k8s.gcr.io here, since it's kinda a valid registry. Or maybe it makes sense to drop it because of the registry freeze.

@puerco @upodroid What do you think about this?

internal/legacy/dockerregistry/schema/manifest_test.go
```diff
@@ -59,11 +59,11 @@ func TestParseThinManifestsFromDirPostsubmit(t *testing.T) {
 			}
 		}
 
-		expectedDigestCount := 12344
+		expectedDigestCount := 14603
 		if onlyProwDiff {
-			expectedDigestCount = 1
+			expectedDigestCount = 14603
```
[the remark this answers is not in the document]
xmudrii: This is because the commit that you picked doesn't promote any image. The way `onlyProwDiff` option works is that it's going to promote/reconcile only images that are promoted in the given commit. Since the given commit doesn't promote any image, it's going to reconcile all images.

I recommend picking some commit that promotes some image. I tried with `e2fd478e1b498d0e87b2ac7e78a1dd53e686551c` and it worked for me.

internal/legacy/dockerregistry/schema/manifest_test.go
```diff
@@ -37,19 +37,19 @@ func TestParseThinManifestsFromDirPostsubmit(t *testing.T) {
 	const (
 		repo   = "https://github.com/kubernetes/k8s.io"
 		git    = "git"
-		commit = "86b8f390aac2e6c244868143ea03c8326c9064a0"
+		commit = "599de960a797bf03a6a90730bd073ea8ef24c46f"
```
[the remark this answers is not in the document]
xmudrii: https://github.com/kubernetes-sigs/promo-tools/pull/792#discussion_r1161656490

xmudrii:
/assign @puerco @upodroid 
/hold
I think we should keep this on hold until 1.27.0 and April patch releases

internal/legacy/dockerregistry/inventory.go
pkprzekwas: Does the registry freeze involve `<1.27` patch releases?

pkprzekwas:
> /assign @puerco @upodroid cc @kubernetes-sigs/release-engineering /hold I think we should keep this on hold until 1.27.0 and April patch releases

I think we should be good to unhold this issue now.

internal/legacy/dockerregistry/inventory.go
BenTheElder: needs more context on what this method is used for

PR #803, opened 2023-03-31
Use throttler for HTTP calls

jonjohnsonjr:
These weren't subject to throttling, so may contribute to quota issues.

#### What type of PR is this?

/kind cleanup

Use the throttler for some HTTP requests in the promoter that were missed.

None

```release-note
NONE
```

/cc @puerco


puerco:
Thanks!!
/lgtm
/approve

PR #868, opened 2023-05-15
Use recursive signing for multi-arch images

saschagrunert:
#### What type of PR is this?

/kind feature

Update the release-sdk and set the option to `true`.

None

We would require a new kpromo release for this.

```release-note
Use recursive signing for multi-arch images.
```

### DIFF
--- a/internal/promoter/image/sign.go
+++ b/internal/promoter/image/sign.go
@@ -114,6 +114,9 @@ func (di *DefaultPromoterImplementation) SignImages(
 	}
 	signOpts.IdentityToken = token
 
+	// We want to sign all entities for multi-arch images
+	signOpts.Recursive = true
+
 	// Creating a new Signer after setting the identity token is MANDATORY
 	// because that's the only way to propagate the identity token to the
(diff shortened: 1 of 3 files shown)


cpanato:
saw some times, but not 100% sure :/

cpanato:
@saschagrunert maybe can we open a test PR to check if the test will pass and maybe we try to fix there if is a flake? i can open the PR and we check, wdyt?

saschagrunert:
> @saschagrunert maybe can we open a test PR to check if the test will pass and maybe we try to fix there if is a flake? i can open the PR and we check, wdyt?

Sounds good to me!

cpanato:
> > @saschagrunert maybe can we open a test PR to check if the test will pass and maybe we try to fix there if is a flake? i can open the PR and we check, wdyt?
> 
> Sounds good to me!

it passes in the first try https://github.com/kubernetes-sigs/promo-tools/pull/869 🤔

saschagrunert:
@cpanato PTAL, looks like the recursive signing is adding layers as well.

cpanato (approved):
thanks
/lgtm

PR #1262, opened 2024-03-31
Upgrade go to 1.22

cpanato:
#### What type of PR is this?

/kind cleanup
/kind feature

- Upgrade go to 1.22

/assign @puerco @saschagrunert @xmudrii 

/hold for test-infra job update

None

```release-note
Upgrade go to 1.22
```

### DIFF
--- a/internal/legacy/cli/audit.go
+++ b/internal/legacy/cli/audit.go
@@ -90,7 +90,7 @@ func (o *AuditOptions) set() {
 
 	// TODO: Should we allow this to be configurable via the command line?
 	o.UUID = os.Getenv("CIP_AUDIT_TESTCASE_UUID")
-	if len(o.UUID) > 0 {
+	if o.UUID != "" {
 		logrus.Infof("Starting auditor in Test Mode (%s)", o.UUID)
 	} else {
 		o.UUID = guuid.NewString()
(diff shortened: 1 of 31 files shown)


internal/legacy/cli/audit.go
```diff
@@ -90,7 +90,7 @@ func (o *AuditOptions) set() {
 
 	// TODO: Should we allow this to be configurable via the command line?
 	o.UUID = os.Getenv("CIP_AUDIT_TESTCASE_UUID")
-	if len(o.UUID) > 0 {
+	if o.UUID != "" {
```
msfidelis, line 93: According to RFC 4211,A UUID has 36 characters, including four hyphens that are used as separators. 

The structure is typically represented as 8-4-4-4-12, making a total of 32 hexadecimal characters and 4 hyphens.

Wouldn't it be interesting to reinforce this validation here?

cpanato: maybe, i am not 100% that is a valid UUID :)
also, in this case, we just add a log line, and if that is empty, we generate a new one, so we don't need to add an extra validation in this case.
but thanks

xmudrii (approved):
/lgtm
/approve

saschagrunert (approved):

PR #1531, opened 2025-05-09
WIP: fix race condition in promotion

justinsb:
I believe there is a race condition here, and we are closing the channel
during promotion before we finish filling it.  This is because the WaitGroup
is incremented _after_ we add to the channel.

Refactor to make the logic (hopefully) clearer.


internal/legacy/dockerregistry/inventory.go
```diff
@@ -1479,18 +1532,17 @@ func (sc *SyncContext) ValidateEdge(edge *PromotionEdge) error {
...
-			logrus.Info("Nothing to promote.")
-			return
-		}
+) []stream.ExternalRequest {
 
+	var requests []stream.ExternalRequest
+
+	if len(toPromote) == 0 {
+		logrus.Info("Nothing to promote.")
+	} else {
```
justinsb, line 1545: This is not the most idiomatic, but preserves indentation for easier review.

PR #1533, opened 2025-05-09
Bump golangci-lint to 1.64.8

justinsb:
Fixes some golangci-lint false-positives


cpanato:
done here with new version https://github.com/kubernetes-sigs/promo-tools/pull/1540
/close

PR #1535, opened 2025-05-09
chore: refactor containers to use generics

justinsb:
golangci-lint was not happy with the type-safety previously.

### DIFF
--- a/internal/legacy/dockerregistry/registry/set.go
+++ b/internal/legacy/dockerregistry/registry/set.go
@@ -24,59 +24,13 @@ import (
 // Various set manipulation operations. Some set operations are missing,
 // because, we don't use them.
 
-// ToSet converts a RegInvImage to a Set.
-func (a RegInvImage) ToSet() container.Set {
-	b := make(container.Set)
-	for k, v := range a {
-		b[k] = v
-	}
-
-	return b
-}
-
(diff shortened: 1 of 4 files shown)


cpanato:
@justinsb can you rebase?

cpanato (approved):
seems good, but i also would like to have another pair of 👁️ on this

/lgtm

/hold
/assign @saschagrunert @puerco

justinsb:
Sorry for delay - I rebased (and also fixed up the test to continue to use sets)

internal/legacy/dockerregistry/registry/set.go
```diff
@@ -24,59 +24,13 @@ import (
 // Various set manipulation operations. Some set operations are missing,
 // because, we don't use them.
 
-// ToSet converts a RegInvImage to a Set.
-func (a RegInvImage) ToSet() container.Set {
```
justinsb, line 28: These are unused and (IMO) are not well defined - observe how ToSet "flattens" a map into a set, but the values have meaning...

justaugustus: Makes sense; thanks Justin!

justaugustus (approved):
Thanks @justinsb!
/hold cancel

PR #1536, opened 2025-05-09
lint: mark incorrect golangci flag

justinsb:
The test is checking that the yaml output is stable,
golangci suggestion is wrong.


cpanato:
thanks @justinsb 
updated here with the new version of it https://github.com/kubernetes-sigs/promo-tools/pull/1540

we can close this one

/close

justinsb:
Thanks for bumping golangci.  However, I think in this case we want to check that the _text_ output is identical, not that the _yaml_ output is identical.  We're checking that the PR workflow is possible.

/reopen

justinsb:
Sigh .... the robots are getting feisty.

Reopened as #1542

PR #1537, opened 2025-05-09
Update golang-ci and fix issues

justinsb:
golangci seems to be blocking fixes from merging.

I updated golang ci and fixed the latest flagged issues.

I've also sent most of these (all but the last) separately to make for easier reviewing,
but I think we have a chicken/egg situation on our hands here.

- **chore: avoid unchecked cast to keep linter happy**
- **lint: mark incorrect golangci flag**
- **chore: refactor containers to use generics**
- **Bump golangci-lint to 1.64.8**
- **golangci: mark cast as ignored**

```release-note

Refactors to keep golangci happy
```


justinsb:
This is a roll-up, and we bumped golangci to the next major in a separate PR.  I'll recheck the individual PRs separately.

/close

PR #1539, opened 2025-05-11
Change "normal" logging to Debugf

justinsb:
The logs from a typical promotion run are now huge.

Reduce the "normal case"s to Debug, so we can see the more
important message at Info.


justinsb:
/test pull-cip-image-cip

Unclear what went wrong here, there's no build-log, and no errors that I can see

saschagrunert (approved):

PR #1542, opened 2025-05-13
lint: mark incorrect golangci flag

justinsb:
The test is checking that the yaml output is stable, golangci suggestion is wrong.

```release-note

NONE
```


justinsb:
Reopened this one from #1536 ... we want to check that the text output is identical, not that the yaml output is identical. We're checking that the PR workflow is possible.

cpanato (approved):

Issue #1714, opened 2026-02-24
Move signature replication to a dedicated periodic Prow job


saschagrunert:
## Summary

Decouple signature replication from the promotion pipeline by moving it to a dedicated periodic Prow job. This eliminates rate limit contention between cosign (which makes untracked HTTP requests during signing) and signature copy operations.

## Background

PR #1713 split signing and replication into separate pipeline phases. However, both still run within the same promotion job and share rate limit budget. Since cosign's HTTP requests bypass our rate limiter entirely, the effective rate limits during signing are unreliable. A dedicated replication job would fully decouple the two workloads.

## Proposal

- Add a `kpromo cip replicate-signatures` subcommand
- It reads the existing promotion manifests to discover which images should exist in which registries
- For each image, it checks whether the signature tag exists in the primary registry and copies it to any mirror registries where it is missing
- The operation is fully idempotent; re-running when everything is already replicated is a series of fast existence checks
- Set up a Prow periodic job running every 30 minutes

## Progress

- [x] Add `kpromo cip replicate-signatures` subcommand (#1715)
- [x] Release v4.3.0 with the new subcommand (#1722)
- [x] Merge [`ci-k8sio-image-signature-replication`](https://prow.k8s.io/?job=ci-k8sio-image-signature-replication) periodic Prow job (kubernetes/test-infra#36516)
- [x] Optimize signature replication with batch tag listing (#1749)
- [x] Soak and verify [`ci-k8sio-image-signature-replication`](https://prow.k8s.io/?job=ci-k8sio-image-signature-replication) job reliability ([sig-release-releng-informing](https://testgrid.k8s.io/sig-release-releng-informing#ci-k8sio-image-signature-replication), [sig-k8s-infra-k8sio](https://testgrid.k8s.io/sig-k8s-infra-k8sio#ci-k8sio-image-signature-replication))
- [x] Remove inline `replicate` phase from `PromoteImages()` (#1750, pipeline goes from 8 to 7 phases: setup → plan → provenance → validate → promote → sign → attest)

Until the final rollout, both the inline phase and the periodic job run in parallel, which is safe since replication is idempotent. This overlap period validates the periodic job before the inline phase is removed from production.

## Trade-offs

**Benefits:**
- Signing and replication have completely independent rate budgets
- Replication failures don't block or fail the promotion job
- Self-healing: missed or failed replications are caught on the next run
- Replication concurrency can be tuned independently

**Costs:**
[...]

PR #1726, opened 2026-02-25
Handle missing source signatures in replicateSignatures

saschagrunert:
#### What type of PR is this?

/kind bug

Skip images whose source `.sig` tag does not exist in the primary registry instead of failing the entire replication job with `MANIFEST_UNKNOWN`. This matches the 404 handling already present in `copyAttachedObjects`.

The `ci-k8sio-image-signature-replication` periodic job fails when it encounters images that were promoted before signing was enabled or whose signatures were garbage collected.

Part of #1714

None

```release-note
Fix signature replication failing on images without signatures
```


xmudrii (approved):
/lgtm
/approve

PR #1727, opened 2026-02-25
Improve standalone signature replication performance

saschagrunert:
#### What type of PR is this?

/kind bug

Two performance improvements for the standalone `replicate-signatures` mode:

1. **Full rate budget**: Give the signing transport the full 50 req/sec instead of 15 (30% of the 70/30 promotion/signing split). Standalone replication has no promotion or signing workload competing for budget.

2. **Source HEAD check**: Before iterating 20+ mirror registries, check if the source `.sig` tag exists on the primary registry. If it doesn't, skip the entire group with a single HEAD instead of 20+ HEAD requests. This is the dominant cost in steady state where most images are already replicated.

With ~46k image groups and 21 regions, steady state was doing ~920k HEAD requests at 15 req/sec (~17 hours). With these changes: ~46k source HEAD checks at 50 req/sec (~15 minutes) for the common case where signatures are already fully replicated.

Part of #1714

None

```release-note
Improve standalone signature replication throughput by using the full rate budget and skipping unsigned images early
```

### DIFF
--- a/internal/promoter/image/sign.go
+++ b/internal/promoter/image/sign.go
@@ -334,11 +334,26 @@ func (di *DefaultPromoterImplementation) replicateSignatures(
 	sourceRefStr := fmt.Sprintf(
 		"%s/%s:%s", src.DstRegistry.Name, src.DstImageTag.Name, sigTag,
 	)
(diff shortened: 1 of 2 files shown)


saschagrunert:
/hold

for now

saschagrunert:
/unhold

PR #1741, opened 2026-02-26
Remove rate limit budget allocator (cleanup)

saschagrunert:
#### What type of PR is this?

/kind cleanup

All pipeline phases run sequentially, so splitting the rate limit budget between promotion and signing is unnecessary overhead. This removes the `BudgetAllocator` and replaces the two separate `RoundTripper` instances with a single shared one that gets the full budget for all phases.

None

```release-note
Give the full rate limit budget to all pipeline phases instead of splitting between promotion and signing
```


xmudrii (approved):
/lgtm
/approve

PR #1748, opened 2026-02-27
Add progress logging and bound concurrency in replicate phase

saschagrunert:
#### What type of PR is this?

/kind bug

The signature replication phase could run silently for ~10 minutes because mirror HEAD checks logged at Debug level and had unbounded goroutine concurrency against the shared rate limiter. This adds per-group progress logging and limits inner mirror concurrency to 10.

None

```release-note
Add progress logging and bound goroutine concurrency in signature replication phase
```


saschagrunert:

PR #1758, opened 2026-03-02
Drop deprecated credential and service account support

### LINKED ISSUE #1693: Drop the use of credentials files
In the latests gcp sdk functions, the `WithCredentialsFile` option is deprecated. 

The derprectation is now causing errors on our lint jobs, I've silenced them for now, but the real fix is to drop the use of credential files entirely:

- Now that the promoter is supposed to get a revamp, let's plan for a stable identity in the jobs so that we can properly impersonate the signer service account from the promoter workload identity.

- We also need to revamp the e2e tests to drop the use of the CIP_E2E_KEY_FILE credential file. 

I think those are the places where json creds are used. If there are more, feel free to add them to this issue.

saschagrunert:
#### What type of PR is this?

/kind cleanup

Removes deprecated credential file support and unused service account fields. The promoter now exclusively uses Application Default Credentials (ADC).

Removed:
- `CredentialsFile` field from `GCPIdentityTokenProvider`
- `GCPServiceActivator` and `ServiceActivator` interface
- `--key-files` CLI flag and `KeyFiles`/`SignerInitCredentials` options
- `CIP_E2E_KEY_FILE` environment variable handling
- `ActivateServiceAccounts` pipeline phase
- `--use-service-account` CLI flag and `UseServiceAcct` option (unused)
- `--json-log-summary` CLI flag and `JSONLogSummary` option (unused)
- `--snapshot-service-account` CLI flag and `SnapshotSvcAcct` option (unused by crane provider)
- `ServiceAccount` from `RegistryConfig` (never consumed by crane provider)

Fixes #1693

The `WithCredentialsFile` call had a `//nolint:staticcheck` suppression because the GCP SDK deprecated it. The code already fell back to ADC when no credentials file was provided, so this change just removes the dead code path.

`ServiceAccount` is kept on `registry.Context` for YAML manifest compatibility but removed from the internal `RegistryConfig` since the crane provider never uses it.

```release-note
Remove deprecated --key-files, --use-service-account, --json-log-summary, and --snapshot-service-account flags from kpromo; use Application Default Credentials instead
```


saschagrunert:

ameukam:
/lgtm
❤️

PR #1761, opened 2026-03-02
Use two-phase tag listing for signature replication

saschagrunert:
#### What type of PR is this?

/kind cleanup

Uses a two-phase tag listing strategy in `computeCopiesFromInventory` to reduce Artifact Registry API calls during signature replication:

1. List tags for source repositories only
2. Filter to groups whose source has a signature
3. List tags for mirror repositories of signed groups only

This skips listing mirror repos for unsigned images (~57% of groups), cutting the total API calls roughly in half.

None

```release-note
NONE
```

### DIFF
--- a/internal/promoter/image/sign.go
+++ b/internal/promoter/image/sign.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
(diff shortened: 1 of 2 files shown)


saschagrunert:

PR #1903, opened 2026-08-06
Implement attestation signing and fix bug generating them

puerco:
#### What type of PR is this?

/kind bug
/kind feature

This PR has two purposes:

First, it fixes a bug that was causing the promotion record attestations from being generated. Currently, the images are only signed but no attestation is attached after an image is promoted, see for example:
https://secint.dev.carabiner.dev/?kind=image&coords=registry.k8s.io%2Fkube-proxy%3Av1.37.0-beta.0

Secondly, it implements the missing attestation signing. As we have never published attestations from the promoter, we leapfrog cosign v2 and now push them directly using the OCI referrers API (the same way that cosign v3 does it now by default).

It adds a new integration test to verify that the signed attestations are verifiable with cosign.

None

/cc @kubernetes-sigs/release-engineering 

```release-note
- Fixed a bug breaking the promotion record generation
- The promoter now signs the promotion record attestation before pushing it to the registry
```


puerco:
OK, tests are green. Note that I've implemented this still using the cosign v2 oci APIs.   Once this is stable and we have verifiable attestations and tests I will port it to cosign v3.

internal/promoter/image/attestation_signer.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	return buf.Bytes(), nil
+}
+
+// ensureAttestationSigner initializes the attestation signer if it is not
+// already set. The signing token comes from the promoter's identity token
+// provider for --signer-account which is the same identity that signs the
+// promoted images. This controls that attestations are only signed with the
+// token and never falling back unintentionally to ambien credentials, etc.
```
saschagrunert: nit: `ambien` → `ambient`

internal/promoter/image/sign.go
```diff
@@ -330,20 +329,33 @@ func (di *DefaultPromoterImplementation) GetIdentityToken(
...
+// WriteProvenanceAttestations generates promotion record attestations for
+// promoted images, signs them into sigstore bundles, and attaches them
+// to the destination image as an OCI 1.1 artifact using the referrers API.
 func (di *DefaultPromoterImplementation) WriteProvenanceAttestations(
 	ctx context.Context,
 	opts *options.Options,
 	edges map[promotion.Edge]any,
 	generator provenance.Generator,
 ) error {
+	// Do no write the attestation if signing is disabled
```
saschagrunert: nit: `Do no write` → `Do not write`

Also: this gates attestation generation on `--sign=true`. Previously attestations were written regardless of the signing flag (though the bug meant they weren't written at all). Is this coupling intentional? It makes sense since signing requires an identity token, but worth calling out as a deliberate behavior change.

internal/promoter/image/sign.go
```diff
@@ -415,64 +427,76 @@ func (di *DefaultPromoterImplementation) pushAttestation(
...
-// hasPredicateType checks if the .att image for the given digest already
-// contains a layer with the specified predicateType annotation.
-func hasPredicateType(digest name.Digest, predicateType string, opts ...ociremote.Option) bool {
-	se := ociremote.SignedUnknown(digest, opts...)
-
-	atts, err := se.Attestations()
+// hasBundleForPredicate checks if the given digest already has an
+// attestation bundle referrer with the specified predicate type.
+//
+// When dealing with descriptors without annotations, we fetching the
```
saschagrunert: nit: `we fetching the` → `we fetch the`

internal/promoter/image/sign.go
```diff
@@ -415,64 +427,76 @@ func (di *DefaultPromoterImplementation) pushAttestation(
...
+			return true
+		}
+
+		if len(desc.Annotations) > 0 {
+			continue
+		}
 
-	for _, sig := range sigs {
-		ann, err := sig.Annotations()
+		// No annotations in the descriptor fetch thethe manifest and check it
```
saschagrunert: nit: `fetch thethe manifest` → `fetch the manifest`

internal/promoter/image/sign.go
```diff
@@ -415,64 +427,76 @@ func (di *DefaultPromoterImplementation) pushAttestation(
...
+		raw, err := remote.Get(
+			digest.Context().Digest(desc.Digest.String()), di.remoteOptions()...,
+		)
 		if err != nil {
 			continue
 		}
 
-		if ann["predicateType"] == predicateType {
+		// Parsse the manifest to find the predicate type annotation
```
saschagrunert: nit: `Parsse` → `Parse`

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/config v1.32.34
 	github.com/aws/aws-sdk-go-v2/feature/s3/manager v1.22.38
 	github.com/aws/aws-sdk-go-v2/service/s3 v1.106.3
+	github.com/carabiner-dev/signer v0.5.4-0.20260804221829-a7025b56e114
```
saschagrunert: `carabiner-dev/signer` is pinned to a pseudo-version. Any timeline for a tagged release? Would be good to move off the commit pin before this ships.

internal/promoter/image/sign.go
puerco: Yes precisely, my thinking was that an unsigned attestation does more harm than good (plus it's not usable from cosign) so we might as well not generate it.

go.mod
puerco: ok, its now pinned to a new release. I thought more was going to be needed for this change, but not really. (and bonus: it picked up a sigstore-go bump )

saschagrunert (approved):
LGTM, nice fix and clean migration to the OCI referrers flow. Left a few nits inline, nothing blocking.

internal/promoter/image/attestation_signer.go
```diff
@@ -0,0 +1,84 @@
...
+// statementSigner signs an in-toto statement and returns a sigstore bundle
+// that can be attached to an image.
+type statementSigner interface {
+	SignStatement(statement []byte) ([]byte, error)
+}
+
+// carabinerSigner implements statementSigner using the carabiner-dev
+// signer with sigstore keyless signing. The underlying signer is safe
+// for concurrent use, so signing parallelizes up to MaxSignatureOps.
+// The Fulcio cert is fetched once and reused by every signature  until
```
saschagrunert, line 38: Nit: double space before "until".

internal/promoter/image/attestation_signer.go
```diff
@@ -0,0 +1,84 @@
...
+limitations under the License.
+*/
+
+package imagepromoter
+
+import (
+	"bytes"
+	"fmt"
+
+	signer "github.com/carabiner-dev/signer"
```
saschagrunert, line 23: Nit: the alias `signer` collides with the struct field name on line 41 (`signer *signer.Signer`), reads a bit like a tongue-twister. Maybe pick a distinct alias.

internal/promoter/image/sign.go
```diff
@@ -415,64 +427,76 @@ func (di *DefaultPromoterImplementation) pushAttestation(
...
+	// Cycle all the manifest descriptors
+	for i := range idx.Manifests {
+		desc := &idx.Manifests[i]
+
+		// Best case scenario: we find the cosign annotation
+		if desc.Annotations[bundlePredicateTypeAnnotation] == predicateType {
+			return true
+		}
+
+		if len(desc.Annotations) > 0 {
```
saschagrunert, line 478: Optional: this assumes that if the registry propagated any descriptor annotations it also propagated the predicate type one (which the OCI 1.1 spec supports, annotations come from the manifest wholesale). A short comment here explaining that assumption would help future readers.

PR #1916, opened 2026-08-21
Bump golangci-lint to v2.13.1

puerco:
#### What type of PR is this?

/kind cleanup

This PR bumps golangci-lint to v2.13 and fixes all nits surfaced by the bump.

None

/cc @cpanato @saschagrunert 

Bumped the installer in the job, config file and all nits.

```release-note
NONE
```

### DIFF
--- a/cmd/kpromo/cmd/pr/pr.go
+++ b/cmd/kpromo/cmd/pr/pr.go
@@ -388,8 +388,8 @@ func mustRun(opts *promoteOptions, question string) bool {
 	if err != nil {
 		logrus.Error(err)
 
-		var userInputErr helpers.UserInputError
-		if errors.As(err, &userInputErr) {
+		//nolint:errcheck // only the type match matters, err is already logged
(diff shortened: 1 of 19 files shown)


cmd/kpromo/cmd/pr/pr.go
```diff
@@ -388,8 +388,8 @@ func mustRun(opts *promoteOptions, question string) bool {
 	if err != nil {
 		logrus.Error(err)
 
-		var userInputErr helpers.UserInputError
-		if errors.As(err, &userInputErr) {
+		//nolint:errcheck // only the type match matters, err is already logged
```
saschagrunert, line 391: Nit: "err is already logged" reads like it refers to the `AsType` return value, but it actually means the `logrus.Error(err)` two lines above. Something like `// discarding matched error value; we only need the type check` would be clearer.

saschagrunert (approved):

PR #1964, opened 2026-09-16
Use longest prefix match for inventory registry lookup

### LINKED ISSUE #1948: Nested destination registries break inventory lookup, causing re-signing and skipped tag move checks
#### What happened:

`splitByKnownRegistries` (`promoter/image/registry/crane.go:241-258`) returns the first base registry that is a path prefix, and the base registries come from map iteration (`GetBaseRegistries`, `promoter/image/promotion/edge.go:288-301`). `k8s-staging-kubernetes` and `k8s-staging-etcd` promote to `<region>-docker.pkg.dev/k8s-artifacts-prod/images`, a prefix of every other subproject destination. If both are part of a run, images like `kube-state-metrics/kube-state-metrics` are keyed under the wrong registry in roughly half of the regions, so already promoted digests look missing:

- they are copied and signed again
- tag move detection (`edge.go:459-485`) is skipped for them
- signatures land in non-canonical regions

Introduced in 416e45d (v4.4.0).

Inventory lookup uses the longest matching prefix (or exact repositories), and already promoted edges are never promotion candidates.

The signature manifest `sha256-42cfe3723a5f058171c627537fb57a3ea0f26e4380fa18555a95cb1a1b4cfc5b.sig` of `kube-state-metrics:v2.20.0` got one new signature per daily `ci-k8sio-image-promo` run between 2026-08-18 and 2026-09-01 (`--manifest-diff-since=14 days` always included kubernetes digests): 8 layers in us-central1, 4 in asia-east1, 2 in asia-northeast1 and 1 in asia-northeast2. `TestSplitByKnownRegistries` has no nested case.

#### Anything else we need to know?:

Keyless signatures never deduplicate in cosign, so signing should also skip digests already signed by the promoter identity.

saschagrunert:
#### What type of PR is this?

/kind bug

`splitByKnownRegistries` returned the first base registry that is a path prefix of the image path, and the base registries come from map iteration. `k8s-staging-kubernetes` and `k8s-staging-etcd` promote to `<region>-docker.pkg.dev/k8s-artifacts-prod/images`, a prefix of every other subproject destination, so when both are part of a run, images are keyed under the wrong registry in a random subset of regions. Already promoted digests then look missing, which copies and signs them again and skips tag move detection.

Pick the longest matching registry instead. Keyless signatures never deduplicate, so this produced one extra signature layer per daily run: `kube-state-metrics:v2.20.0` collected 8 layers in us-central1, 4 in asia-east1, 2 in asia-northeast1 and 1 in asia-northeast2 between 2026-08-18 and 2026-09-01.

Introduced in 416e45d (v4.4.0), so this is worth a patch release.

Fixes #1948

The regression test is a separate function so it does not conflict with #1916, which rewrites `TestSplitByKnownRegistries`. It fails on the current code and covers both list orders, because the registry order is not deterministic.

Skipping digests that the promoter identity already signed, mentioned at the end of the issue, is deliberately out of scope here. It is defense in depth against a second path to duplicate signatures, not part of the reported behavior, and deserves its own change.

```release-note
Fixed inventory lookup for destination registries that are path prefixes of each other. Images promoted to nested registries are no longer copied and signed again on every run, and tag move detection is no longer skipped for them.
```


saschagrunert:

puerco (approved):

Issue #1952, opened 2026-09-15
Per-project provenance policy in promoter manifests


saschagrunert:
#### What would you like to be added:

Projects declare expectations in `promoter-manifest.yaml`:

- trusted signer identities and OIDC issuers for staging signatures and attestations
- builder IDs, source repositories and build types
- required predicate types
- mode: `off` (default), `warn` or `require`

The staging build identity is separate from the promoter signing identity. Dry runs show the effective policy per edge.

#### Why is this needed:

Verification uses global identity flags (production: `krel-staging|krel-trust`). A staging signature from a project identity fails the validate phase today, and attestations are verify-if-present, so projects can neither sign with their own identity nor opt in to stronger guarantees. `promoter-manifest.yaml` is owned by k8s.io reviewers, which makes it the right place for policy.

Depends on #1951, first adopter kubernetes-sigs/security-profiles-operator#3442

Issue #1951, opened 2026-09-15
Discover all attestations of staging images and artifacts


saschagrunert:
#### What would you like to be added:

Collect every attestation and signature bundle of a staging digest from:

- OCI referrers of the index (including Docker manifest lists) and each platform manifest
- legacy cosign `.att` tags
- statements whose subjects are layer digests (files packaged as OCI artifacts)

Dry runs print a per digest report (predicate type, signer identity, location), so the credential-less presubmit shows what was found. Discovery errors must be distinguishable from "no attestations".

#### Why is this needed:

A VSA has to list every attestation it considered, and builders attach them in different places. Cloud Build native provenance is out of scope: it is stored in Artifact Analysis, which is disabled in staging projects and not readable by presubmits. carabiner-dev/collector already has `oci` and `coci` drivers and should be evaluated first (it skips identity checks, which stay in kpromo). Bundle signature detection overlaps with release-sdk (#1947).

Depends on #1943, pilot data from kubernetes-sigs/security-profiles-operator#3442

PR #1965, opened 2026-09-16
Fix provenance verification of staging attestations

### LINKED ISSUE #1943: Provenance verification blocks promotion of any staging image with an attestation
#### What happened:

The provenance phase (`promoter/image/provenance/cosign.go`) blocks promotion as soon as a staging image has a cosign `.att` tag:

- `PredicateType` is never set, so cosign returns `missing predicate type` for matching attestations (cosign v2.6.3 `verify_attestation.go:335`, v3.1.3 `verify_attestation.go:234`). Production v4.5.1 is affected.
- Attestations signed by other identities (for example a project build service account) fail with `ErrNoMatchingAttestations`.
- On main with the production flags, cosign v3 fails even earlier with `KeyAndIdentityParseError`, because `--certificate-identity` keeps its default while `--certificate-identity-regexp` is set.

This is latent because staging images carry no `.att` tags today.

Attestations from trusted identities verify, attestations from other identities are ignored with a warning, and only cryptographic or subject mismatches of trusted identities fail promotion.

Promote (dry run) a staging image with a cosign attestation using the production flags `--certificate-identity-regexp=(krel-staging@…)|(krel-trust@…) --certificate-oidc-issuer=https://accounts.google.com`.

#### Anything else we need to know?:

- `Result.Builder` and `Result.SourceRepo` are never set.
- Verification runs per edge (region and tag) without the rate limited transport.
- `--certificate-identity` and `--certificate-identity-regexp` should be mutually exclusive.

Keep this fix minimal. Referrer discovery is #1951, per-project identities are #1952.

saschagrunert:
#### What type of PR is this?

/kind bug

The provenance phase blocked promotion of any staging image carrying a cosign `.att` tag:

- `PredicateType` was never set, so cosign rejected every attestation with `missing predicate type`
- `CertIdentity` keeps its default while `--certificate-identity-regexp` is set in production, which cosign rejects as mutually exclusive
- attestations signed by another identity failed the phase

This is latent today because staging images carry no `.att` tags, and it blocks the Security Profiles Operator pilot.

Set a predicate type (SLSA provenance v1 by default), pass only the exact identity or its regular expression, and treat an attestation that does not match the configured identity or predicate type as a warning. Verification now also shares the promoter rate limited transport and runs once per source digest instead of once per destination region and tag.

Fixes #1943

The part that needs review is how "signed by someone else" is told apart from "does not verify". Cosign returns the same `ErrNoMatchingAttestations` for both, so a failed first pass triggers a second pass accepting any identity. If that verifies, the attestation is valid but foreign and promotion continues; if nothing verifies, the attestation is malformed or tampered with and promotion is blocked. A third party attestation signed with a key rather than keyless would still block. Per-project identities in #1952 close that gap.

`Result.Builder` and `Result.SourceRepo` are still unset. Populating them means parsing the verified statement, which is what #1951 builds.

```release-note
Fixed provenance verification blocking the promotion of staging images that carry a cosign attestation. Attestations from identities or predicate types other than the configured ones are now logged and ignored, and only attestations that fail to verify at all block a promotion.
```


saschagrunert:

puerco (approved):
Thanks!

We should later discuss how to handle verifying for other identities, to allow the rest of the projects to define their own. Also: I'm planning to replace our own homegrown verification for the new slsa verifier once we cut the first stable release (probably this week or so)

saschagrunert:
> Thanks!
> 
> We should later discuss how to handle verifying for other identities, to allow the rest of the projects to define their own. Also: I'm planning to replace our own homegrown verification for the new slsa verifier once we cut the first stable release (probably this week or so)

Yes, a release + rollout + see how it goes sounds great after a bunch of fixes. :+1:

Issue #1946, opened 2026-09-15
Rework sigcheck for canonical signatures and promotion attestations


saschagrunert:
#### What would you like to be added:

`kpromo sigcheck` checks and repairs signatures and promotion attestations on the canonical registry only:

- repairs sign or attest there with the `registry.k8s.io` subject
- identity checks honor `--certificate-identity-regexp`
- `.att` tags are not checked as images
- the command fails if problems remain

Then update `periodic-release-verify-image-signatures` in kubernetes/test-infra: new version, and for repairs `--confirm`, the promoter service account and the `k8sio-image-promo` job queue.

#### Why is this needed:

sigcheck still assumes signatures on all mirrors:

- Mirror hosts come from a hard coded `k8s-staging-kubernetes` manifest, so every image promoted since #1829 is reported as partially signed.
- `FixPartialSignatures` copies `.sig` tags to mirrors, `FixMissingSignatures` signs `Missing[0]` (a random mirror) without `SignContainerIdentity`.
- The identity regexp is ignored and attestations are not checked.
- The periodic job runs without `--confirm`, and `CheckSignatures` returns success even if problems are found.

If signing or attesting fails after the copy, reruns skip those edges because the images already exist in production, so nothing repairs them today.

#1944 is done (v4.6.0), so this can start.

Issue #1945, opened 2026-09-15
Serve OCI referrers from the canonical registry through registry.k8s.io


saschagrunert:
#### What would you like to be added:

Track the infrastructure changes to make signatures and attestations on the canonical registry reachable:

- kubernetes/k8s.io: allow `^/v2/.+/referrers/.+$` in `infra/gcp/terraform/modules/oci-proxy/cloud-armor.tf` (the comment "AR doesn't support referrers API" is outdated). Keep `k8s-artifacts-prod` (created by `infra/gcp/bash/lib_ar.sh`) free of cleanup policies for untagged manifests.
- kubernetes/registry.k8s.io: route `/v2/<name>/referrers/<digest>` to `SIGNATURE_UPSTREAM_ENDPOINT`, and fall back to it for manifests by digest and blobs that only exist there. The fallback needs a cached existence check off the hot path, because every pull fetches platform manifests by digest.
- e2e coverage from a GCP client for `cosign verify` and referrers.

#### Why is this needed:

Referrers requests to registry.k8s.io return 404 at Cloud Armor, while `us-central1-docker.pkg.dev/v2/k8s-artifacts-prod/images/<name>/referrers/<digest>` returns 200. archeio sends every non-blob request to the regional backend of the serving Cloud Run instance, only `sha256-<hex>.(sig|att)` tags go to the canonical upstream. Signature manifests by digest therefore still return 404 (kubernetes/registry.k8s.io#324 is reproducible with the `kube-apiserver:v1.37.0` signature manifest), and signature blobs are missing for GCP clients (#1949). Availability and egress concerns from kubernetes/registry.k8s.io#321 need to be addressed.

Done when `cosign verify` and `cosign verify-attestation` work through registry.k8s.io from any region without affecting image pulls.

PR #1966, opened 2026-09-16
Write promotion attestations only to the canonical registry

### LINKED ISSUE #1944: Write promotion attestations only to the canonical registry

Push one signed promotion attestation per identity and digest to `us-central1-docker.pkg.dev`:

- Target the canonical destination even if only other regions are promotion candidates (see #1948).
- Use the `registry.k8s.io` reference (`targetIdentity`) as subject name. cosign v3 bundles carry no `docker-reference`, so the subject is the only place for the production name.
- Serialize the exists check and push per digest, so multiple tags of one digest create a single referrer.
- Decide how tagless edges (index children) are covered, they get no attestation today.

#### Why is this needed:

`WriteProvenanceAttestations` (#1903, unreleased, production runs v4.5.1) signs and pushes per edge: 22 or more different bundles and transparency log entries per digest, with regional subject names. `hasBundleForPredicate` checks and pushes without locking, so tags of the same digest race. Signatures prefer the canonical registry since #1829, but only among candidate edges.

Serving through registry.k8s.io: #1945

saschagrunert:
#### What type of PR is this?

/kind feature

`WriteProvenanceAttestations` signed and pushed per promotion edge, which produces one bundle and transparency log entry per region and tag of a digest, 22 or more for a fully promoted image, each carrying a regional subject name. The exists check and the push also ran unlocked, so tags of the same digest raced each other and attached one referrer per tag.

Group the edges by target identity and digest, as signing already does, and write a single attestation per group to `us-central1-docker.pkg.dev`. The destination is rewritten to the canonical registry, so a digest gets its attestation there even when only other regions are promotion candidates. Cosign v3 bundles carry no `docker-reference`, which makes the subject the only place for the production name, so the record now holds the `registry.k8s.io` reference instead of the regional one.

#1903 is unreleased and production runs v4.5.1, so this should land before the next release.

Fixes #1944

On the open question in the issue: tagless edges are now attested as well. An attestation attaches to a digest, and those digests are promoted like any other. Signing still skips them, which belongs to #1946.

The canonical rewrite only applies inside `k8s-artifacts-prod/images` on an Artifact Registry host, so project and test registries are untouched.

Serving these attestations through registry.k8s.io is #1945.

```release-note
Promotion attestations are now written once per digest to the canonical registry us-central1-docker.pkg.dev, with the registry.k8s.io reference as subject name, instead of once per destination region and tag. Digests promoted without a tag are covered as well.
```


saschagrunert:

puerco (approved):
This is great, thanks!!

PR #1974, opened 2026-09-17
Record promotion context in the promotion predicate

saschagrunert:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

The promotion record only held the references, digest, a shared timestamp and the builder ID. The manifest, commit, Prow job and promoter commit were available at runtime but dropped, which makes audits harder.

The `https://k8s.io/promo-tools/promotion/v1` predicate now also records, additively:

- the promoted tags
- source and destination as in-toto resource descriptors
- the manifest file listing the digest (relative path, `gitCommit`, and the repository as `git+https` URI without credentials)
- the promoter version and commit
- the Prow job name, type, build ID and ProwJob ID

`make update-proto` regenerates the protobuf code with a pinned buf, `make verify-proto` checks it as part of `make verify`. The predicate is documented in `docs/promotion-predicate.md`.

#### Which issue(s) this PR fixes:

Fixes #1953

#### Special notes for your reviewer:

- in-toto's ResourceDescriptor proto is not available to buf, so a minimal message with the same field names mirrors it.
- `WriteProvenanceAttestations` now receives the parsed manifests to map edges to their manifest.
- The Prow job URL is not recorded, the environment has no reliable Deck host or log bucket.
- Linking the source VSA or provenance is left for later.
- `https://k8s.io/promo-tools/promotion/v1` still needs a redirect to the new spec in kubernetes/k8s.io.

On hold until v4.6.0 (#1970) is released, so this lands in the release after it.

/hold

#### Does this PR introduce a user-facing change?

```release-note
Promotion record attestations now include the promoted tags, source and destination descriptors, the manifest file and commit, the promoter version and commit, and the Prow job. The predicate is documented in docs/promotion-predicate.md.
```

Issue #1957, opened 2026-09-15
Validate OCI artifact types before promotion


saschagrunert:
#### What would you like to be added:

- Planning rejects unsupported shapes before copying, for example the deprecated `application/vnd.oci.artifact.manifest.v1+json` or index children that are neither images nor manifests.
- Vulnerability scans report non-image artifacts as not applicable.

Declaring accepted `artifactType` values per image or registry moved to #1959, since it depends on the conventions defined there.

#### Why is this needed:

Helm charts and SPO seccomp profiles already promote through `kpromo cip`, but the manifest schema knows nothing about artifact types. Recursive signing signs the index and then fails with `unknown mime type` on unsupported children after the copy, which stops the pipeline with partially signed objects. Grafeas scans return a clean result for any unscanned reference, including artifacts.

go-containerregistry copies an index child with an unknown media type like `application/vnd.oci.artifact.manifest.v1+json` through the blob endpoint. On registries that also serve manifests as blobs the copy succeeds and the signing walk then fails, elsewhere the promote phase fails. `TestRecursiveSigningWalkArtifacts` in #1976 documents the current behavior and has to change with this issue.

PR #1976, opened 2026-09-17
Test promotion of OCI 1.1 artifacts

### LINKED ISSUE #1958: Test OCI 1.1 artifact promotion

Integration tests (in-memory registry) for copy, signing and attestation of:

- empty config (`application/vnd.oci.empty.v1+json`) with a custom `artifactType`
- Helm chart media types
- manifests with a `subject`
- an index of per-platform artifacts
- a single layer blob of 300 MB or more

#### Why is this needed:

Tests only use random images and indexes, and `CraneProvider.ReadRegistries` is untested (the in-memory registry lacks Google's tags list extension). Artifact promotion works in production, but it is not covered by tests.

Related: #1957

saschagrunert:
#### What type of PR is this?

/kind cleanup

Helm charts and seccomp profiles already promote through `kpromo cip`, but the tests only used random images. This adds integration tests with in-memory registries for:

- an OCI 1.1 artifact with an empty config and a custom `artifactType`
- Helm chart media types
- a manifest with a `subject`, which stays discoverable as referrer after promotion
- an index of per-platform artifacts
- a single layer blob of 300 MiB (32 MiB with the race detector, so the unit test job stays within its memory limit; skipped with `-short`)

They cover copying, promotion attestations and carrying staging signatures. `CraneProvider.ReadRegistries` is now tested as well, against a registry that serves the Google tags list extension.

Signing itself needs an identity token and Rekor, so the tests check the walk recursive signing uses instead. One test documents that an index child with the deprecated `application/vnd.oci.artifact.manifest.v1+json` media type is copied, but fails that walk with `unknown mime type`, which is what #1957 addresses.

Fixes #1958

go-containerregistry copies index children of unknown media types as blobs. The test pushes such a child as a blob as well, like registries that store manifests as blobs serve it. The in-memory registry keeps them apart, so the promote phase would fail there already.

The race detector is detected from the build info instead of a `//go:build race` file, because the boilerplate check only accepts `// +build` lines before the license header.

#1974 adds a parameter to `WriteProvenanceAttestations`, which `TestWriteProvenanceAttestationsArtifacts` calls, so whichever of the two merges second needs a one line update.

```release-note
NONE
```


saschagrunert:

cpanato (approved):

PR #1977, opened 2026-09-17
feat: add --issue flag to kpromo pr for issue cross-references

cpanato:
#### What type of PR is this?

/kind feature

Add --issue flag to kpromo pr

Allows linking a GitHub issue to an image promotion PR. When set, the
flag is validated as a full http(s) URL and rendered as an `xref: <link>`
line near the top of the PR body, plus echoed in the reproduction
command block.

/assign @saschagrunert @Verolop @puerco 

None

```release-note
feat: add --issue flag to kpromo pr for issue cross-references
```


cmd/kpromo/cmd/pr/pr.go
```diff
@@ -422,6 +439,10 @@ func generatePRBody(opts *promoteOptions) string {
 		args += " --reviewers \"" + opts.reviewers + "\""
 	}
 
+	if opts.issue != "" {
+		args += " --issue " + opts.issue
```
saschagrunert: Could we quote the value like `--reviewers` above? A URL with `&`, for example with query parameters, gets cut off when the reproduction command is pasted into bash.

```suggestion
		args += " --issue \"" + opts.issue + "\""
```

cmd/kpromo/cmd/pr/pr.go
```diff
@@ -79,6 +81,13 @@ func (o *promoteOptions) Validate() error {
 		return errors.New("cannot start promotion --tag is required")
 	}
 
+	if o.issue != "" {
+		issueURL, err := url.Parse(o.issue)
+		if err != nil || (issueURL.Scheme != "http" && issueURL.Scheme != "https") || issueURL.Host == "" {
```
saschagrunert: This accepts any http(s) host, while the flag help says GitHub issue. Is that intended, for example for other trackers? If not, checking `issueURL.Host == "github.com"` would match the help text. Either way, a small table test for `Validate()` would be nice: no `--issue`, a valid https URL, a URL without scheme, `ftp://…`, and `https://` without host.

cmd/kpromo/cmd/pr/pr.go
```diff
@@ -440,6 +461,11 @@ func generatePRBody(opts *promoteOptions) string {
 	args += imageString.String()
 
 	prBody := fmt.Sprintf("Image promotion for %s %s\n", opts.project, strings.Join(opts.tags, " / "))
+
+	if opts.issue != "" {
+		prBody += "\nxref: " + opts.issue + "\n\n"
```
saschagrunert, line 470: Nit: the body layout only changes with `--issue`. A test for `generatePRBody()` with and without `--issue` would pin down the body and the reproduction command, since `cmd/kpromo/cmd/pr` has no tests yet.

saschagrunert (commented):
Thanks, looks good overall. Two things before merging: quoting the URL in the reproduction command, and a few tests for `Validate()` and `generatePRBody()`, since `cmd/kpromo/cmd/pr` has none. Optional: the release note could say "`kpromo pr` accepts `--issue` to reference a GitHub issue in the promotion PR body." instead of repeating the title with the `feat:` prefix.

cmd/kpromo/cmd/pr/pr.go
cpanato: done

saschagrunert (approved):
Thanks for addressing the feedback!

/lgtm

PR #1979, opened 2026-09-17
Accept relative and trailing slash thin manifest dirs

saschagrunert:
#### What type of PR is this?

/kind bug

`kpromo cip --thin-manifest-dir` failed with `unexpected manifest path` when the directory was given as `.`, `./` or with a trailing slash, because the manifest path was shortened by trimming the argument as a string prefix. The path is now made relative to the directory with `filepath.Rel`, so these forms work like a plain absolute path.

None

Found while running a local dry run against a copy of the security-profiles-operator thin manifest.

```release-note
`kpromo cip --thin-manifest-dir` accepts relative directories and a trailing slash.
```


saschagrunert:

puerco (approved):
Thanks!

